PR #1, opened 2020-09-15
Apply GitHub migrations

marcellp:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,191 @@
...
+    - id: cache-seek
+      name: Check for npm cache hit
+      uses: c-hive/gha-npm-cache@v1
+
+    - id: install
+      name: Install dependencies
+      run: |-
+        npm ci
+
+    # - id: test
```
ivababukova: we need to fix this

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,191 @@
...
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: ./Dockerfile
+          platforms: linux/amd64
+          tags: ${{ format('{0}/{1}:{2}', steps.login-ecr.outputs.registry, steps.ref.outputs.repo-name, steps.ref.outputs.tag) }}
+          push: false
+
+      - id: push
+        name: Push docker image to ECR
```
ivababukova, line 117: we should constrain this step to be applied only when we will be deploying. Otherwise, it will only be creating waste in ECR. We need to add in here:
```
if:
      github.ref == 'refs/heads/master'
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,191 @@
...
+        run: |-
+          echo Pushing image $IMAGE_NAME to ECR.
+          docker push $IMAGE_NAME
+        env:
+          IMAGE_NAME: ${{ format('{0}/{1}:{2}', steps.login-ecr.outputs.registry, steps.ref.outputs.repo-name, steps.ref.outputs.tag) }}
```
ivababukova, line 122: what is an example value for `steps.ref.outputs.tag`?

marcellp: It is documented in the CI on lines 72-74.

    # the final tag is something like:
    # refs-heads-master-a4f8bc313dae
    # this is what we push to ECR

.github/workflows/ci.yaml
marcellp: Fixed.

.github/workflows/ci.yaml
marcellp: Fixed.

ivababukova (approved):

PR #3, opened 2020-09-15
bug fixed

StefanBabukov:


src/pages/experiments/[experimentId]/data-exploration/components/header/index.jsx
```diff
@@ -100,6 +100,7 @@ const Header = (props) => {
           <Space>
             <FeedbackButton />
             <Dropdown
+              trigger={['click']}
               key='search-menu-dropdown'
               overlay={searchMenu}
               visible={addMenuVisible}
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/header/index.jsx:106]: I think the problem is the visible stuff here. Did you try to remove it entirely? Because I think that removing it entirely might actually be the fix we need.

ivababukova (approved):

PR #4, opened 2020-09-16
Fix cell info coordinates

ivababukova:


package.json
```diff
@@ -33,40 +33,46 @@
     "redux-thunk"
   ],
   "dependencies": {
-    "express": "^4.16.3",
```
ivababukova, line 36: no idea how and why these things got changed ...

src/pages/experiments/[experimentId]/data-exploration/components/CellInfo.jsx
```diff
@@ -19,16 +19,14 @@ const CellInfo = (props) => {
...
-
     }
   }, [cellInfo]);
 
   const renderCellInfo = () => (
     <Card
       size='small'
       style={{
-        zIndex: 6,
+        zIndex: 0,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/CellInfo.jsx]: this change is not needed

package.json
ivababukova: just checked all of these, they and their versions are all the same. Only vitessce is changed from `"^0.1.8"` to `"^0.2.1"` and this change was needed in order to solve the cell coordinates bug

src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx
```diff
@@ -212,7 +209,7 @@ const Embedding = (props) => {
         cells={convertCellsData(data)}
         mapping='PCA'
         selectedCellIds={selectedCellIds}
-        cellColors={(selectedCell) ? { ...cellColors, [selectedCell]: [0, 0, 0] } : cellColors}
+        cellColors={(selectedCell) ? new Map(Object.entries({ ...cellColors, [selectedCell]: [0, 0, 0] })) : new Map(Object.entries(cellColors))}
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx:212]: Is this related to the bug where the coloring was glitching out?

src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx
```diff
@@ -109,13 +107,13 @@ const Embedding = (props) => {
   }, [focusedExpression]);
 
   const updateCellCoordinates = (newView) => {
-    if (selectedCell && newView.viewport.project) {
-      const [x, y] = newView.viewport.project([data[selectedCell][0], data[selectedCell][1]]);
+    if (selectedCell && newView.project) {
+      const [x, y] = newView.project(selectedCell);
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx:111]: am I understanding this correctly that somehow there is a project method in the newView that is aware of the concept of a cell ID?

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx:111]: somehow yes. No idea how it works, but when I put the cellId there, it just worked.

src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/embedding/Embedding.jsx:212]: I have no idea why Vitessce introduced it, but now it it required to pass `Map` instead of plain `object`.

marcellp (approved):

PR #7, opened 2020-09-29
Redux actions files

StefanBabukov:


src/redux/reducers/cellSets/cellInfoReducer.js
```diff
@@ -0,0 +1,15 @@
+import { UPDATE_CELL_INFO } from '../../actionTypes';
+
+const cellInfoReducer = (state = {}, action) => {
+  switch (action.type) {
+    case UPDATE_CELL_INFO:
+      return {
+        ...state,
+        ...action.data,
```
ivababukova: how is this even working? It should be `action.payload`, not `action.data`, because `data` has been renamed to `payload` in `updateCellInfo` action

ivababukova (approved):

PR #18, opened 2020-10-20
Total variable in differential expression when in volcano is no longer undefined

StefanBabukov:


src/redux/actions/loadDifferentialExpression.js
```diff
@@ -50,8 +49,9 @@ const loadDifferentialExpression = (
       experimentId, REQUEST_TIMEOUT, body, pagination,
     );
     const data = JSON.parse(response.results[0].body);
-    const { rows, total } = data;
-
+    let { total } = data;
```
ivababukova, line 52:
```
let {total} = data;
const {rows} = data;
if (!data.total) {
    total = rows.length;
}
```

ivababukova (approved):

PR #30, opened 2020-10-29
Increase cache ttl to 12 hours and fix some issues with tests

ivababukova:


src/utils/cacheRequest.js
```diff
@@ -18,57 +16,6 @@ const objectToSortedString = (object) => {
   return sortedString;
 };
 
-const cacheFetch = async (endpoint, options = {}, ttl = 900) => {
```
ivababukova, line 21: this code is not used anyway so I thought that it is best to just delete it

marcellp (commented):
What is the difference between `_set` and `set`?

ivababukova:
No difference. I renamed it, because it is used in external files and since it is used aoutside of the cache class, I think it makes sense to be named without an `_`

marcellp (approved):

PR #34, opened 2020-11-05
Fix typo in CI deployment

marcellp:


.github/workflows/ci.yaml
```diff
@@ -170,19 +170,14 @@ jobs:
...
+            CHART_REF="STAGING_CHART_REF"
           fi
           
-          if [ "${{ github.ref }}" = "refs/heads/master"]; then
-            CHART_REF="$GITHUB_SHA"
-          else
-            CHART_REF="$GITHUB_REF"
-          fi
-
-          echo "::set-output name=sandbox-id::$SANDBOX_ID"
```
marcellp, line 181: The idea is the same as with the sandbox ID. If we are deploying to a staging candidate file, we want the chart reference to be an easy value for the script to overwrite, so the user can configure the deployment to automatically update the chart or keep it at the latest version always.

.github/workflows/ci.yaml
```diff
@@ -170,19 +170,14 @@ jobs:
...
-
           yq w -d1 .flux.yaml metadata.name "$DEPLOYMENT_NAME" \
           | yq w -d1 - metadata.namespace "$DEPLOYMENT_NAME-$SANDBOX_ID" \
           | yq w -d0 - metadata.name "$DEPLOYMENT_NAME-$SANDBOX_ID" \
+          | yq w -d0 - metadata.tags.sandboxId "$SANDBOX_ID" \
+          | yq w -d1 - metadata.tags.sandboxId "$SANDBOX_ID" \
```
ivababukova, line 181: why twice?

marcellp: They're different documents. `d0` is the first one in `.flux.ci` (the namespace), and `d1` is the second (the HelmRelease).

ivababukova (approved):

PR #39, opened 2020-11-19
More controls added to plots and tables

StefanBabukov:


src/pages/experiments/[experimentId]/data-processing/filter-cells/components/CellSizeDistribution/CellSizeDistribution.jsx
```diff
@@ -107,11 +109,11 @@ class CellSizeDistribution extends React.Component {
     const coloringExpressionPlot2 = `(datum.u < ${minUnknown}) ? 'low' : (datum.u >${minHigh}) ? 'high' : 'unknown'`;
 
     if (config.legendEnabled) {
-      config.legendOrientation = config.plotToDraw ? 'top-left' : 'top-right';
+      // config.legendPosition = config.plotToDraw ? 'top-left' : 'top-right';
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-processing/filter-cells/components/CellSizeDistribution/CellSizeDistribution.jsx]: Is this line needed? If not, we should get rid of it.

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -257,27 +300,32 @@ const VolcanoPlot = () => {
...
-                  onChange={(value) => {
-                    useRef(_.throttle(() => updatePlotWithChanges({ textThresholdValue: value }), 20));
+                  onChange={value => {
+                    useRef(
+                      _.throttle(
+                        () =>
+                          updatePlotWithChanges({ textThresholdValue: value }),
+                        20,
+                      ),
+                    );
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx]: This is not correct. `useRef` should be declared in the top of the component, not in the `return` block. Refs get created once when the component renders for the first time and persist across renders. If you use `useRef` in the return block, the ref gets created on each render, which makes it useless.

Do this as you did with other components where you used `onUpdateThrottled`.

marcellp (changes requested):

src/pages/experiments/[experimentId]/data-processing/filter-cells/components/CellSizeDistribution/CellSizeDistribution.jsx
```diff
@@ -1,11 +1,18 @@
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/filter-cells/components/CellSizeDistribution/CellSizeDistribution.jsx:2]: why are these needed?

Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/data-processing/filter-cells/components/CellSizeDistribution/CellSizeDistribution.jsx:2]: so it doesnt give errors on the `````` elements, because the linter requires them to have events which i dont use

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -156,26 +192,35 @@ const VolcanoPlot = () => {
...
-    return (<Skeleton />);
+    return <Skeleton />;
   }
 
   const renderPlot = () => {
     if (error) {
       return (
         <PlatformError
-          description='Could not load differential expression data.'
+          description="Could not load differential expression data."
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:219]: single quotes is fine, why changing it?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -185,69 +230,70 @@ const VolcanoPlot = () => {
...
-      <Header plotUuid={plotUuid} experimentId={experimentId} finalRoute={route} />
+      <Header
+        plotUuid={plotUuid}
+        experimentId={experimentId}
+        finalRoute={route}
+      />
       <Row gutter={16}>
         <Col span={16}>
-          <Space direction='vertical' style={{ width: '100%' }}>
+          <Space direction="vertical" style={{ width: '100%' }}>
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:247]: fix your editor to stop changing single quotes to double quotes

ivababukova (approved):

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:219]: Something is wrong with your linter @StefanBabukov Can you save over all your files again so you get a consistent quote standard?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx]: This is still not resolved as far as I can see.

marcellp (changes requested):
See unresolved thread.

marcellp (approved):

PR #43, opened 2020-11-26
Make UI able to handle categorical metadata as cell sets

marcellp:


src/redux/actions/cellSets/createCellSet.js
```diff
@@ -17,7 +17,7 @@ const createCellSet = (experimentId, name, color, cellIds) => (dispatch, getStat
     key: uuidv4(),
     name,
     color,
-    cellIds: Array.from(cellIds),
+    cellIds,
```
marcellp, line 20: cells are now sets, so this is not a problem

src/redux/reducers/cellSets/cellSetsCreate.js
```diff
@@ -17,7 +17,7 @@ const cellSetsCreate = (state, action) => {
   });
 
   newState.properties[key] = {
-    key, name, color, cellIds,
+    key, name, color, cellIds: new Set(cellIds),
```
marcellp, line 20: the reducer needs to create a set from cells now

ivababukova (approved):

PR #46, opened 2020-12-03
Add union and intersection functionality to the UI

marcellp:


marcellp:
Tests still outstanding.

src/pages/experiments/[experimentId]/data-exploration/components/cell-sets-tool/CellSetOperation.jsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+import {
+  Button,
+} from 'antd';
+
+import ClusterPopover from '../embedding/ClusterPopover';
+
+const CellSetOperations = (props) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/cell-sets-tool/CellSetOperation.jsx:10]: nit-pick: call the file "CellSetOperations"

Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/cell-sets-tool/CellSetOperation.jsx]: call this class `CellSetOperation` instead

src/pages/experiments/[experimentId]/data-exploration/components/embedding/ClusterPopover.jsx
```diff
@@ -23,20 +25,34 @@ const ClusterPopover = (props) => {
...
   return (
-    <div style={{ position: 'absolute', left: popoverPosition.current.x + 20, top: popoverPosition.current.y + 20 }}>
-      <Popover title='Add cell set' content={content} visible />
+    <div style={style}>
+      <Popover title={message} content={content} {...restOfProps}>
+        {children}
+      </Popover>
     </div>
   );
+  /* eslint-enable react/jsx-props-no-spreading */
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/embedding/ClusterPopover.jsx:41]: put the eslint disable at the start of the file instead

ivababukova (commented):
just a few small things

src/pages/experiments/[experimentId]/data-exploration/components/embedding/ClusterPopover.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/embedding/ClusterPopover.jsx:41]: I wouldn't, because it only applies to the prop spreading stuff inside the return statement. My rationale is it's best practice to only disable linting for rules like this for the segment that actually needs to bypass it.

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/embedding/ClusterPopover.jsx:41]: you are right actually, I agree

ivababukova (approved):

PR #47, opened 2020-12-04
Add heatmap menu on the gene table for selecting genes

ivababukova:
To view the functionality, it is deployed to staging and can be accessed here: https://ui-heatmap.scp-staging.biomage.net/


src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/SelectionActions.jsx
```diff
@@ -78,27 +80,38 @@ const SelectionIndicator = (props) => {
...
           </Text>
-          <Button type='link' size='small' onClick={clearAll}>Clear</Button>
+          <Button type='link' size='small' onClick={clearAll}>Clear selected</Button>
           {renderCopyClipboard()}
+          <Button
+            type='link'
+            size='small'
+            onClick={() => { setListed(!listed); onListSelected(!listed); }}
+          >
+            {listed ? 'Hide selected' : 'List selected'}
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/SelectionActions.jsx]: just say clear, copy, list

src/redux/reducers/genes/genesExpressionError.js
```diff
@@ -1,10 +1,21 @@
...
       ...state.expression,
+      views: {
+        ...state.properties.views,
+        [componentUuid]: {
+          ...initialViewState,
+          ...state.expression.views[componentUuid],
+          fetching: false,
+          error,
+        },
+      },
```
ivababukova, line 19: dig into this

src/redux/reducers/genes/initialState.js
```diff
@@ -1,42 +1,3 @@
-/**
- * The store has the following format:
- * {
- *    genes: {
```
ivababukova, line 4: reconsider this

src/redux/reducers/genes/initialState.js
```diff
@@ -59,6 +20,7 @@ const initialState = {
     loading: [],
     error: false,
     data: {},
+    views: {},
```
ivababukova, line 23: remove this

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -20,59 +21,43 @@ const HeatmapPlot = (props) => {
...
-      <center style={{ marginTop: height / 2 }}>
-        <Spin size='large' />
-        <HeatmapCrossHairs />
-      </center>
-    );
-  }
+    if (!selectedGenes || selectedGenes.length === 0) {
+      console.log('skipping because the selected genes dont exist or are none');
+      return;
+    }
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: don't do anything with data if selectedGenes is none or empty -- in this case `createVegaData` would break, so we'd get nothing back. let's leave it at the previous state.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -20,59 +21,43 @@ const HeatmapPlot = (props) => {
...
-            dispatch(loadGeneExpression(experimentId, selectedGenes, componentType));
-          }
-        }}
-      />
-    );
-  }
+    if (_.intersection(selectedGenes, loadingGenes).length > 0) {
+      setVegaData(null);
+      return;
+    }
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: if currently loading, set the data to null and return.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/VegaHeatmap.jsx
```diff
@@ -41,14 +41,16 @@ const VegaHeatmap = (props) => {
...
+  const vegaSpec = {
+    ...spec,
+    height: getAdjustedHeight(),
+    width: getAdjustedWidth(),
+    axes: getAxes(),
+    data: spec.data.map((datum) => ({
+      ...datum,
+      values: data[datum.name],
+    })),
+  };
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/VegaHeatmap.jsx:53]: this was the main problem, we modified the prop directly, which fucked with the rendering somehow.

Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/VegaHeatmap.jsx:53]: This is a problem because:

* `data` as a prop was never passed on to `Vega`. That is, a modification to `data` will not cause re-render of `<Vega/>` within `<VegaHeatmap/>`, only `<VegaHeatmap/>` itself.
* modification of `spec` as a prop will not cause a re-render in and of itself either, because props are meant to be immutable. modifying them within a component is really bad.

So, the solution:

* create a copy of the `spec`, name it `vegaSpec`, and construct it from the `spec` and `data` props
* use that `vegaSpec` to render `<Vega/>`. if any of the things in the spec get modified, we will get a new heatmap

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -154,11 +98,46 @@ const HeatmapPlot = (props) => {
...
+  if (!selectedGenes || selectedGenes.length === 0) {
+    return (
+      <center>
+        <Empty
+          description={(
+            <span>
+              Please add gene(s) to the heatmap from the Gene list tool
+            </span>
+          )}
+        />
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Using "please" in instructions like this is not recommended in UX, it makes the interaction a bit stuffy. It is also not obvious why you should do it in the first place, see [here](https://ux.stackexchange.com/questions/10853/should-an-interface-ever-say-please) re "please" in UX.

Typically a good empty state has the following:

* a header describing *what* it is you're meant to see here [e.g. gene expression data], and making it clear that there is an action you can perform to do it
* give you hints on how to do it in a friendly, conversational tone

See also

> Good, modern empty state designs I’ve observed are typically two parts instruction, one part delight. If a UI is empty, it’s the UX designer’s job to help users understand what will go in the space when it’s populated. If it’s the user’s job to populate it, the messaging should include instructions about how. (For example, an empty “Classes” table in an edtech product might include copy informing teachers how to create their first class.)

Potentially better: 

```
                <Empty description={(
                  <>
                    <div><Text type='primary'>No expression data to show</Text></div>
                    <div><Text type='secondary'>You can add genes to display here from the gene list tool.</Text></div>
                  </>
                )}
                />
```

marcellp (approved):
Just one misc change, other than that LGTM.

PR #48, opened 2020-12-04
Fix bug in embedding plots & tables and clean up the code a bit.

marcellp:


src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
```diff
@@ -92,6 +121,7 @@ const EmbeddingCategoricalPlot = () => {
     newCellSets = newCellSets.map(({ key }) => ({
       cellSetId: key,
       ...cellSets.properties[key],
+      cellIds: Array.from(cellSets.properties[key].cellIds),
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx:124]: This was the actual bug.

src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
```diff
@@ -116,33 +154,22 @@ const EmbeddingCategoricalPlot = () => {
...
-        <Vega spec={vegaSpec} renderer='canvas' />
+        <Vega spec={plotSpec} renderer='canvas' />
       </center>
     );
   };
 
-  const vegaSpec = generateSpec(config);
-  // due to a bug in vega with React with using data with source coming from other data,
-  // we have to inject the data in the Vega spec.
-  generateData(vegaSpec);
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx:136]: Doing stuff like this is kind of bad, because it will not allow us to load the link directly from the browser, which I found out when I was looking for the bug.

src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
```diff
@@ -57,26 +57,55 @@ const EmbeddingCategoricalPlot = () => {
...
   useEffect(() => {
-    if (isBrowser) {
-      if (!data) {
-        dispatch(loadEmbedding(experimentId, embeddingType));
-      }
-      dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
-      dispatch(loadCellSets(experimentId));
+    if (!experimentId || !isBrowser) {
+      return;
+    }
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx:66]: If we don't check for whether experimentId exists, this will start loading embeddings/plot configs/cell sets with experimentId undefined, which is a bug.

ivababukova (approved):

PR #65, opened 2021-01-11
Implement the UI changes required for multiple-sample support

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-425

https://ui-5xdznj5e78cutequhytt9k6qiy.scp-staging.biomage.net/

N/A

N/A

# Changes
### Code changes
Differential expression is now aware of multiple samples. Implemented Vicky's wireframes as described in the document for multi-sample support. Fixed unit tests and cleaned up the diff expr code a bit.

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers
- [x] (UX changes) Approved by @vickymorrison (this is her username, tag her if you need approval)

### Just before merging:

### Optional
image

### DIFF
--- a/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
+++ b/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
@@ -95,8 +95,6 @@ const DiffExprCompute = (props) => {
 
   // Validate form when the groups selected changes.
   useEffect(() => {
-    console.log(selectedGroups);
-
     validateForm();
   }, [selectedGroups]);
 
--- a/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
+++ b/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
@@ -4,7 +4,7 @@ import {
 } from 'react-redux';
 
 import {
-  Button, Form, Select, Typography, Radio,
+  Button, Form, Select, Typography, Radio, Empty,
 } from 'antd';
 
 import PropTypes from 'prop-types';
@@ -32,6 +32,10 @@ const DiffExprCompute = (props) => {
   const [selectedGroups, setSelectedGroups] = useState(cellSets);
   const [type, setType] = useState(diffExprType || ComparisonType.between);
 
+  // `between sampoles/groups` makes no sense if there is no metadata. First,
+  // assume there is and correct this if it's necessary.
+  const [hasMetadata, setHasMetadata] = useState(false);
+
   /**
    * Loads cell set on initial render if it does not already exist in the store.
    */
@@ -39,14 +43,18 @@ const DiffExprCompute = (props) => {
     dispatch(loadCellSets(experimentId));
   }, []);
 
-  /**
-   * Re-renders the list of selections when the hierarchy or the properties change.
-   *
-   * If the cell set previously selected is deleted, the selection is reset to the default.
-   */
   useEffect(() => {
     if (hierarchy.length === 0) return;
 
+    // Make sure we are not rendering metadata-related options if there is no metadata in the data set.
+    const numMetadata = Object.values(properties).filter((o) => o.type === 'metadataCategorical').length;
+
+    if (!numMetadata) {
+      setType(ComparisonType.within);
+    }
+    setHasMetadata(numMetadata > 0);
+
+    // Update the hierarchies if a previously selected set were to be deleted.
     setSelectedGroups(_.mapValues(selectedGroups, (cellSetKey) => {
       if (!cellSetKey) {
         return null;
@@ -98,7 +106,6 @@ const DiffExprCompute = (props) => {
    * @param {string} option The option string (`cellSet` or `compareWith`).
    */
   const onSelectCluster = (cellSet, option) => {
-
     setSelectedGroups({
       ...selectedGroups,
       [option]:
@@ -113,7 +120,7 @@ const DiffExprCompute = (props) => {
     title, option, filterType,
   }) => {
     // Dependiung on the cell set type specified, set the default name
-    const placeholder = `Select a ${filterType === 'metadataCategorical' ? 'sample/group' : 'cell set'}`;
+    const placeholder = filterType === 'metadataCategorical' ? 'sample/group' : 'cell set';
 
     // Get all the stuff we are going to show.
     const tree = composeTree(hierarchy, properties, filterType);
@@ -154,7 +161,7 @@ const DiffExprCompute = (props) => {
     return (
       <Form.Item label={title}>
         <Select
-          placeholder={placeholder}
+          placeholder={`Select a ${placeholder}...`}
           style={{ width: 200 }}
           onChange={(cellSet) => onSelectCluster(cellSet, option)}
           value={selectedGroups[option]}
@@ -202,7 +209,7 @@ const DiffExprCompute = (props) => {
           basis: null,
         });
       }} value={type}>
-        <Radio style={radioStyle} value={ComparisonType.between}>
+        <Radio style={radioStyle} value={ComparisonType.between} disabled={!hasMetadata}>
           Compare a selected cell set between samples/groups
         </Radio>
         <Radio style={radioStyle} value={ComparisonType.within}>
--- a/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
+++ b/src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
@@ -87,6 +87,8 @@ const DiffExprCompute = (props) => {
 
   // Validate form when the groups selected changes.
   useEffect(() => {
+    console.log(selectedGroups);
+
     validateForm();
   }, [selectedGroups]);
 
@@ -178,7 +180,7 @@ const DiffExprCompute = (props) => {
             ))
           }
         </Select>
-      </Form.Item>
+      </Form.Item >
     );
   };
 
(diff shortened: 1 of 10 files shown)


vickymorrison:
Screenshot 2021-01-11 at 18 57 52
When 'compare cell sets within a sample/group' is selected, the second menu should not have 'All other cells' at the top. (Users would, instead, select 'Rest of Louvain clusters' as this would be the same thing. See screenshot above.

ivababukova:
also, the first menu option always says `Compare cell set:` no matter which radiobutton option is being selected

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -46,92 +47,144 @@ const DiffExprCompute = (props) => {
...
 
+  // Validate form when the groups selected changes.
   useEffect(() => {
     validateForm();
-  }, [selectedCellSets]);
+  }, [selectedGroups]);
+
+  // Erase selection when a different type of comparison is selected.
+  useEffect(() => {
+
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: does this empty `useEffect` actually erase state like this?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -46,92 +47,144 @@ const DiffExprCompute = (props) => {
...
    */
   const onSelectCluster = (cellSet, option) => {
-    setSelectedCellSets({
-      ...selectedCellSets,
-      [option]: cellSet,
+    console.log({
+      ...selectedGroups,
+      [option]:
+        cellSet,
+    });
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: remove console log

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -46,92 +47,144 @@ const DiffExprCompute = (props) => {
...
+          return <Option key={uniqueKey} disabled={shouldDisable(uniqueKey)}>
+            {name}
           </Option>
-        );
-      });
+        });
+      }
     };
+
+    console.log(selectedGroups);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: remove this

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -176,10 +289,12 @@ const DiffExprCompute = (props) => {
 };
 
 DiffExprCompute.defaultProps = {
+  diffExprType: null,
 };
 
 DiffExprCompute.propTypes = {
   experimentId: PropTypes.string.isRequired,
+  diffExprType: PropTypes.string,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx:293]: this should also be required, otherwise wouldn't the code in this component fail?

ivababukova:
@marcellp it is quite hard to test the full functionality here without any metadata and this is occurring in most of the multiple-sample support features. We could run staging with Vicky's multiple-sample dataset, how easy would it be to change staging to point to it instead?

marcellp:
> When 'compare cell sets within a sample/group' is selected, the second menu should not have 'All other cells' at the top. (Users would, instead, select 'Rest of Louvain clusters' as this would be the same thing. See screenshot above.

@vickymorrison -- this works fine for Louvain clusters, but breaks for scratchpad -- or for any other grouping of sets whose union is not the entire experiment.

Say I lasso out a cell set and create a cluster for it in scratchpad. Without this, there would be no way to compare that cell set to the rest of the cells in the experiment.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: no, that was leftover code, fixed.

marcellp:
> @marcellp it is quite hard to test the full functionality here without any metadata and this is occurring in most of the multiple-sample support features. We could run staging with Vicky's multiple-sample dataset, how easy would it be to change staging to point to it instead?

@ivababukova -- multisample is now available as an experiment in all staging environments.

https://ui-5xdznj5e78cutequhytt9k6qiy.scp-staging.biomage.net/experiments/multisample/data-exploration

ivababukova:
Screenshot 2021-01-12 at 14 50 46
There is a small usability issue here -- the menu options fields look like they are disabled when the user hasn't specified anything. I think it is do with the font colour. Can we fix this so that they look active?

vickymorrison:
> > When 'compare cell sets within a sample/group' is selected, the second menu should not have 'All other cells' at the top. (Users would, instead, select 'Rest of Louvain clusters' as this would be the same thing. See screenshot above.
> 
> @vickymorrison -- this works fine for Louvain clusters, but breaks for scratchpad -- or for any other grouping of sets whose union is not the entire experiment.
> 
> Say I lasso out a cell set and create a cluster for it in scratchpad. Without this, there would be no way to compare that cell set to the rest of the cells in the experiment.

Then there are two options for how to fix this: 1. 'All cells' remains at the top as you have it, and 'Rest of Louvain clusters' is removed. Or 2. Move 'All cells' to be under the Scratchpad heading.
Otherwise, the user could get confused because 'All cells' and 'Rest of Louvain clusters' is the same thing in this instance.
I don't feel strongly about which of the above options is best - happy for you to choose based on engineering technicalities.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -46,92 +47,133 @@ const DiffExprCompute = (props) => {
...
    */
   const onSelectCluster = (cellSet, option) => {
-    setSelectedCellSets({
-      ...selectedCellSets,
-      [option]: cellSet,
+
+    setSelectedGroups({
+      ...selectedGroups,
+      [option]:
+        cellSet,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx:110]: specifying `option` here as a key list creates the need for having more complicated logic in all places where we will be handling these selected groups. Is there any reason why we don't instead structure this like this:

```
...
selectOption: [option],
cellset, 
...
```
instead of 
```
[option]:
   cellset,
```

marcellp:
> Then there are two options for how to fix this: 1. 'All cells' remains at the top as you have it, and 'Rest of Louvain clusters' is removed. Or 2. Move 'All cells' to be under the Scratchpad heading.
> Otherwise, the user could get confused because 'All cells' and 'Rest of Louvain clusters' is the same thing in this instance.
> I don't feel strongly about which of the above options is best - happy for you to choose based on engineering technicalities.

@vickymorrison This is also not great, because then the converse happens. If someone selects something in scratchpad, then `Rest of scratchpad` and `All other cells` are *not* the same.

Similarly, say you have a particular metadata that you are interested in comparing, but it does not involve your entire experiment. Say, you have something like a data set which has a pooled WT and separate people with a disease. Then it is reasonably expected that `age` as a variable will not be recorded for the pooled WT sample, but only for the clinical cases. Then, `Rest of age` would correspond to rest of all cells that have `age` recorded, and `All other cells` would be the rest of all cells, irrespective of the value of the `age` field.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx:110]: `option` refers to `cellSet`, `compareWith`, and `basis`. The created data structure is essentially:

```
{
    cellSet: 'louvain-0',
    compareWith: 'background',
    basis: 'all'
}
```

which gets passed straight into `DifferentialExpression` as a work body. I don't see how the proposal would work with this data structure. Maybe you misunderstood what the `option` parameter is, or do you think refactoring is still the best idea? Would be happy to discuss.

vickymorrison:
You are right Marcell. The way you have it is the best way to encompass all use cases.

vickymorrison (approved):

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx:110]: we discussed over zoom and it is good actually, I misunderstood it

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx:293]: No, see [here](https://github.com/biomage-ltd/ui/pull/65/files/457298551f9708358095cb1c9cf881ddf75a6473#diff-b0755e09efec37c88e6905392bcc561eea89f5633c608c60ad8aac226da7a70bR33).

marcellp:
@ivababukova -- usability issues fixed. Color is now the primary color for the heatmaps, and the option with cross-sample DE is disabled and inaccessible when experiment has no metadata.

ivababukova (approved):

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -4,153 +4,269 @@ import {
...
       setIsFormValid(false);
       return;
     }
+
     setIsFormValid(true);
   };
 
+  // Validate form when the groups selected changes.
   useEffect(() => {
+    console.log(selectedGroups);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: remove this console log

PR #66, opened 2021-01-12
Frequency fixes

StefanBabukov:
@vickymorrison 

This implements the fix in the occasion, when there is no metadata clusters in the hierarchy. 
When this happens, the diagram plotted is only including the louvain clusters (with metadata clusters, intersection is done with the selected cluster and all of the louvain).
[Staging link](https://ui-xormzk31uipw5yoyakuo819ctc.scp-staging.biomage.net/experiments/5e959f9c9f4b120771249001/plots-and-tables/frequency)
image

**Note** - when no metadata clusters are available I made the menu disabled (see screenshot). Tell me in terms of UI if that is fine or should I put something else (tooltip saying no metadata clusters are in the sample  for example).


vickymorrison:
If there is only a single sample, can the greyed out menu say 'Sample' rather than 'condition'. A tooltip if the user hovers over this box would also be useful, as you suggested. E.g. "The x-axis cannot be changed as this dataset has only a single sample." 

Louvain should have a capital 'L'.

Also, the colouring of this plot should reflect the chosen colours in Data Exploration. I assume this is the case but I cannot tell from the screenshot.

StefanBabukov:
> If there is only a single sample, can the greyed out menu say 'Sample' rather than 'condition'. A tooltip if the user hovers over this box would also be useful, as you suggested. E.g. "The x-axis cannot be changed as this dataset has only a single sample."
> 
> Louvain should have a capital 'L'.
> 
> Also, the colouring of this plot should reflect the chosen colours in Data Exploration. I assume this is the case but I cannot tell from the screenshot.

Good that you mentioned it, I'll make sure the colours match.

src/pages/experiments/[experimentId]/plots-and-tables/frequency/components/SelectCellSets.jsx
```diff
@@ -3,31 +3,48 @@ import PropTypes from 'prop-types';
...
   Select,
+  Tooltip,
 } from 'antd';
 
 const SelectCellSets = (props) => {
   const { onUpdate, config, cellSets } = props;
-
   const changeClusters = (val) => {
-    onUpdate({ chosenClusters: val.key });
+    const newValue = val.key.charAt(0).toLowerCase() + val.key.slice(1);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/frequency/components/SelectCellSets.jsx]: Isn't this `String.toLowerCase()`?

vickymorrison:
When the plot shows counts, why does Scratchpad add up to 5000 but Louvain adds up to around 2400? This doesn't make sense. This is the cell count so should be the same if all the cells have been allocated custom clusters in Scratchpad which seems to be the case here i.e. it should be the total number of cells shown in the embedding in Data Exploration.

A small thing: when the plot is set to proportion the y-axis should show 0 to 100 (currently on Scratchpad the y-axis goes to 110 which it shouldn't).

StefanBabukov:
> When the plot shows counts, why does Scratchpad add up to 5000 but Louvain adds up to around 2400? This doesn't make sense. This is the cell count so should be the same if all the cells have been allocated custom clusters in Scratchpad which seems to be the case here i.e. it should be the total number of cells shown in the embedding in Data Exploration.
> 
> A small thing: when the plot is set to proportion the y-axis should show 0 to 100 (currently on Scratchpad the y-axis goes to 110 which it shouldn't).



> When the plot shows counts, why does Scratchpad add up to 5000 but Louvain adds up to around 2400? This doesn't make sense. This is the cell count so should be the same if all the cells have been allocated custom clusters in Scratchpad which seems to be the case here i.e. it should be the total number of cells shown in the embedding in Data Exploration.
> 
> A small thing: when the plot is set to proportion the y-axis should show 0 to 100 (currently on Scratchpad the y-axis goes to 110 which it shouldn't).

**resolved**

src/pages/experiments/[experimentId]/plots-and-tables/frequency/components/SelectCellSets.jsx
```diff
@@ -3,31 +3,48 @@ import PropTypes from 'prop-types';
...
     ));
-    return filteredOptions;
+    if (!filteredOptions.length) {
+      return [];
+    }
+    // making the options with capital letters as per requirement
+    const upperCaseOptions = [];
+    filteredOptions.forEach((option) => {
+      upperCaseOptions.push({
+        value: option.value.charAt(0).toUpperCase() + option.value.slice(1),
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/frequency/components/SelectCellSets.jsx]: I see this operation repeated several times in this file (lines : 38, 59, 77) : `option.value.charAt(0).toUpperCase() + option.value.slice(1)`. Can you refactor it into a function?

src/redux/reducers/componentConfig/initialState.js
```diff
@@ -139,6 +139,7 @@ const frequencyInitialConfig = {
   frequencyType: 'proportional',
   metadata: 'condition',
   legendEnabled: true,
+  metadataDisabled: false,
```
aerlaut: Can you tell me what is this property for?

aerlaut (commented):
Looks good. Just need to refactor the three functions on `index.js`

src/redux/reducers/componentConfig/initialState.js
StefanBabukov: Not needed, just forgot to remove it.

aerlaut (approved):

PR #67, opened 2021-01-13
Feature/make UI able to request z score

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-429

https://ui-qex8r3r5g4g5f59u3ikexxb9br.scp-staging.biomage.net/

N / A

- When `expressionType` is changed, `loadGeneExpression` is called with `forceReloadAll = true`, to force update for all genes, otherwise new data will not be loaded
- Added `loadGeneExpression` debounce criteria to support concurrent updates of multiple panels: if there is still genes `loading`, check if the requested call is due to `expressionType`
- Right now, cells in the heatmap is not ordered (?), so when new data is loaded, the heatmap seems to scramble/rearrange.

# Changes
### Code changes

- API for `loadGeneExpression` is changed, added new optional parameter `expressionType = 'raw'` before `forceReloadAll`.
- `expressionType` now used to calculate cached work hash
- Cosmetic changes: Spinner in embedding is centered, like in heatmap

# Definition of DONE

### Testing
- [X] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


ivababukova:
@aerlaut regarding the comment in the description stating `Right now, cells in the heatmap is not ordered (?), so when new data is loaded, the heatmap seems to scramble/rearrange.` - the cells are ordered according to clusters. The refreshing of the gene expression data should not change the ordering of the cells.

aerlaut:
The clusters are still ordered, however the arrangement of cell ID changes every time the data is reloaded. In these pictures, I hover over the left-most cell on gene S100A9. I change the expressionType before taking each picture (raw -> z-score, z-score -> raw, raw -> z-score)

image

image

image

ivababukova:
@vickymorrison is this a big issue and should we worry about it?

ivababukova:
I actually just noticed that this gene rearranging is happening also if you displaying genes on the heatmap, after a click on an eye icon. So it is not a bug introduced with this change, if Vicky thinks that it is something we have to fix, I would suggest that we don't fix it as part of this PR, but either as part of this ticket: https://biomage.atlassian.net/browse/BIOMAGE-406 if the cause of the issue is the same, or as part of a new ticket.

vickymorrison:
> @vickymorrison is this a big issue and should we worry about it?

My understanding of the issue (correct me if I'm wrong) is that the user can control only one level of ordering so within in this case Louvain cluster 0 the cells are randomly ordered and are redistributed every time the heatmap re-renders.

I don't think this is a massive problem for users.

ivababukova:
understood, thanks for clarification Vicky! In that case Agi, let's not worry about this particular issue with cell ids getting rearranged.

src/redux/actions/genes/updateGeneExpressionType.js
```diff
@@ -0,0 +1,33 @@
...
+  // Get components currently using the gene data, then update the components
+  const updates = Object.keys(views).map((componentUuid) => dispatch(
+    loadGeneExpression(
+      experimentId,
+      views[componentUuid].data,
+      componentUuid,
+      expressionType,
+      true,
+    ),
+  ));
```
ivababukova, line 18: There is a better way to get all components that are currently using the gene expression data.

All expression data that is being visualised anywhere in the platform is stored under `genes/expression/data` in redux. Here is an example screenshot, where we have gene `FOS` displayed on the embedding and genes `S100A9`, `CD74` and `LYZ` displayed on the heatmap and the contents of the gene store:
Screenshot 2021-01-13 at 12 40 06

Therefore, in order to get all genes who need their expression updated, all you need to do is to get all gene names that are under `genes/expression/data`

src/redux/reducers/genes/initialState.js
```diff
@@ -21,6 +21,7 @@ const initialState = {
     error: false,
     data: {},
     views: {},
+    expressionType: 'raw',
```
ivababukova, line 24: shouldn't `expressionType` be under `expression` in the `genes` store? At the end of the day, it is a thing specific to the expression of the genes, the `properties` part of the `genes` store does not care about it.

src/__test__/utils/cacheRequest.test.js
```diff
@@ -4,42 +4,46 @@ import { /* cacheFetch, */ fetchCachedWork } from '../../utils/cacheRequest';
...
   },
   E: { hello: 'world' },
 };
 
 const fakeCacheContents = {
-  fd3161a878f67ebf54018720cffd6a66: 'A', // pragma: allowlist secret
-  '10250a11679234110a1c260d6fd81d3c': 'B', // pragma: allowlist secret
-  f5c957411a28de68f35e1f5c8a29da7e: 'C', // pragma: allowlist secret
-  '33faa711a94a2028b5bae1778126aec0': 'E', // pragma: allowlist secret
+  '5b995fe9ed8f00cff155a59afa6c523c': 'A', // pragma: allowlist secret
```
ivababukova, line 43: I wonder how and why this comment `// pragma: allowlist secret` got there

src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/ComponentActions.jsx
```diff
@@ -31,7 +32,7 @@ const ComponentActions = (props) => {
       newGenes = displayedGenes.filter((gene) => !selectedGenes.includes(gene));
     }
 
-    dispatch(loadGeneExpression(experimentId, newGenes, componentType));
+    dispatch(loadGeneExpression(experimentId, newGenes, componentType, expressionType));
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/ComponentActions.jsx:35]: instead of having to make any changes to `ComponentActions` and make it "know" and "care" about `expressionType`, what you can do is to read the value of `expressionType` directly from `loadGeneExpression` action, similarly to how it is done in action `loadPaginatedGeneProperties`, line 16 as an example.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -299,7 +300,7 @@ const HeatmapPlot = (props) => {
       <PlatformError
         description={error}
         onClick={() => {
-          dispatch(loadGeneExpression(experimentId, selectedGenes, COMPONENT_TYPE));
+          dispatch(loadGeneExpression(experimentId, selectedGenes, COMPONENT_TYPE, expressionType));
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx:303]: similarly to my comment for `ComponentActions`, you shouldn't need those changes here.

src/redux/actions/genes/updateGeneExpressionType.js
aerlaut: `loadGeneExpression` requires `genes` and `componentUuid` as a part of its parameters, that's why I used `genes/expression/views` to get `componentUuid` and the genes under each component and update them accordingly. I suppose we can use `genes/expression/data` directly and pass it into `loadGeneExpression`, and then use a preset `componentUuid` (e.g. 'InteractiveHeatmap'), but that would cause problems when a gene toggled for a panel (e.g. embedding) but not for other panel (e.g. heatmap) will show up in the wrong panel.

Else, I can try to update the genes directly without going through `loadGeneExpression`, e.g. refactor the data fetching part into a separate function, and have it under `loadGeneExpression` and `updateGeneExpressionType`.

src/redux/reducers/genes/initialState.js
aerlaut: It is actually under `expression` (line 19), it's cut by the diff (have to expand)

src/__test__/utils/cacheRequest.test.js
aerlaut: It's part of package [`detect-secrets`](https://github.com/Yelp/detect-secrets#inline-allowlisting) to detect secrets in code-bases. Maybe Marcell is using it in the pipeline to sift for sensitive strings?

src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/ComponentActions.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/generic-gene-table/ComponentActions.jsx:35]: Oh ya, that would make it better API-wise for the function too. So better change it?

src/redux/reducers/genes/initialState.js
aerlaut: It's actually under `expression`. There property is there on line 19, but the diff cut it so have to expand to see it,

src/__test__/utils/cacheRequest.test.js
marcellp: Correct. `detect-secrets` looks for patterns for sensitive credentials (such as AWS, GitHub PATs, crypto keys), and it has an additional function where it looks for any high-entropy string, which I thought was valuable in case we started using something else in the future.

Obviously, hashes fall under this category, so that's why we explicitly ignore them.

src/redux/actions/genes/updateGeneExpressionType.js
marcellp: I agree with Agi, I just have one concern here regarding `expressionType` being in the root of `expressionType`. Let's discuss on a call.

aerlaut:
I've decided to implement the modification in a new branch, so I'm closing this pull request

PR #74, opened 2021-01-19
Feature/persist diffexpr chosen options

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-427

https://ui-78ns9ga3rkuyfdhcx7ngc5bq4y.scp-staging.biomage.net/

The staging link doesn't really show the ability to save between the different options, because the first radio button is disabled due to lack of metadata. If possible, please try to pull the PR and check locally.

N/A

# Changes
### Code changes

- Added 2 new actions and reducers for handling context switches and changes to groups
- Added new key `comparison` under `differentialExpression` which stores `type` of the expression currently active and also `group` keyed by the `type` to store selected options.
- Added new key `comparisonType` under `differentialExpression > properties` to store the comparison type that was used to generate the differential expression. 
- Synced data between differential expression and volcano plot - changing one changes the other.

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
@@ -17,7 +17,7 @@ const { Item } = Form;
 
 const mockStore = configureMockStore([thunk]);
 
-const store = mockStore({
+const initialState = {
   cellSets: {
     properties: {
       'cluster-a': {
@@ -93,7 +93,26 @@ const store = mockStore({
       },
     ],
   },
-});
+  differentialExpression: {
+    properties: {
+      data: [],
+      cellSets: {},
+      loading: false,
+      error: false,
+      total: 0,
+    },
+    comparison: {
+      type: 'between',
+      group: {
+        between: {
+          cellSet: null,
+          compareWith: null,
+          basis: null,
+        },
+      },
+    },
+  },
+};
 
 describe('DiffExprCompute', () => {
   beforeAll(async () => {
@@ -118,6 +137,8 @@ describe('DiffExprCompute', () => {
 
   configure({ adapter: new Adapter() });
   it('renders correctly with no comparison method', () => {
+    const store = mockStore(initialState);
+
     const component = mount(
       <Provider store={store}>
         <DiffExprCompute
@@ -149,6 +170,25 @@ describe('DiffExprCompute', () => {
   });
 
   it('clicking on the second comparison option changes items in form', () => {
+    const store = mockStore({
+      ...initialState,
+      differentialExpression: {
+        ...initialState.differentialExpression,
+        comparison: {
+          ...initialState.differentialExpression.comparison,
+          type: 'within',
+          group: {
+            ...initialState.differentialExpression.comparison,
+            within: {
+              cellSet: null,
+              compareWith: null,
+              basis: null,
+            },
+          },
+        },
+      },
+    });
+
     const component = mount(
       <Provider store={store}>
         <DiffExprCompute
@@ -159,21 +199,13 @@ describe('DiffExprCompute', () => {
       </Provider>,
     );
 
-    let form = component.find(Form);
+    const form = component.find(Form);
 
     // There should be one form.
     expect(form.length).toEqual(1);
 
     // It should have a radio button group at the top
     expect(form.find(Radio.Group).length).toEqual(1);
-
-    // Clicking on the second one should re-render the tool.
-    const { onChange } = form.find(Radio.Group).getElement().props;
-
-    onChange({ target: { value: 'within' } });
-
-    component.update();
-    form = component.find(Form);
 
     // The form should still have four items.
     expect(form.find(Item).length).toEqual(4);
@@ -190,6 +222,8 @@ describe('DiffExprCompute', () => {
   });
 
   it('the `versus` option renders correctly when a set is already selected', () => {
+    const store = mockStore(initialState);
+
     const component = mount(
       <Provider store={store}>
         <DiffExprCompute
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
@@ -9,6 +9,9 @@ import {
   Form, Radio, Button, Select,
 } from 'antd';
 import DiffExprCompute from '../../../../../../../pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute';
+import {
+  DIFF_EXPR_COMPARISON_GROUP_SET, DIFF_EXPR_COMPARISON_TYPE_SET,
+} from '../../../../../../../redux/actionTypes/differentialExpression';
 
 jest.mock('localforage');
 jest.mock('../../../../../../../utils/environment', () => false);
@@ -170,7 +173,9 @@ describe('DiffExprCompute', () => {
   });
 
   it('clicking on the second comparison option changes items in form', () => {
-    const store = mockStore({
+    // MockStore can not update state, therefore the store here reflects changes after click
+    // and click is checked by action
+    const store = mockStore(() => ({
       ...initialState,
       differentialExpression: {
         ...initialState.differentialExpression,
@@ -178,7 +183,6 @@ describe('DiffExprCompute', () => {
           ...initialState.differentialExpression.comparison,
           type: 'within',
           group: {
-            ...initialState.differentialExpression.comparison,
             within: {
               cellSet: null,
               compareWith: null,
@@ -187,7 +191,7 @@ describe('DiffExprCompute', () => {
           },
         },
       },
-    });
+    }));
 
     const component = mount(
       <Provider store={store}>
@@ -206,6 +210,14 @@ describe('DiffExprCompute', () => {
 
     // It should have a radio button group at the top
     expect(form.find(Radio.Group).length).toEqual(1);
+
+    // Clicking on the second one should re-render the tool.
+    const { onChange } = form.find(Radio.Group).getElement().props;
+
+    onChange({ target: { value: 'within' } });
+
+    // Check if the correct actions are fired
+    expect(store.getActions()[0].type).toEqual(DIFF_EXPR_COMPARISON_TYPE_SET);
 
     // The form should still have four items.
     expect(form.find(Item).length).toEqual(4);
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
@@ -10,7 +10,7 @@ import {
 } from 'antd';
 import DiffExprCompute from '../../../../../../../pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute';
 import {
-  DIFF_EXPR_COMPARISON_GROUP_SET, DIFF_EXPR_COMPARISON_TYPE_SET,
+  DIFF_EXPR_COMPARISON_TYPE_SET,
 } from '../../../../../../../redux/actionTypes/differentialExpression';
 
 jest.mock('localforage');
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
@@ -82,7 +82,6 @@ describe('DiffExprManager', () => {
   });
 
   it('on click of compute with changed parameters, DiffExprManager calls the results view', () => {
-    console.log(filledStore.getState());
     const component = mount(
       <Provider store={filledStore}>
         <DiffExprManager experimentId='1234' view='compute' width={100} height={200} />
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
@@ -16,7 +16,7 @@ jest.mock('../../../../../../../utils/environment', () => false);
 
 const mockStore = configureMockStore([thunk]);
 
-const emptyStore = mockStore({
+const emptyState = {
   differentialExpression: { ...initialState },
   cellSets: {
     hierarchy: [],
@@ -25,7 +25,30 @@ const emptyStore = mockStore({
   genes: {
     focused: undefined,
   },
-});
+};
+
+const filledState = {
+  ...emptyState,
+  differentialExpression: {
+    ...emptyState.differentialExpression,
+    comparison: {
+      ...emptyState.differentialExpression.comparison,
+      type: 'between',
+      group: {
+        ...emptyState.differentialExpression.comparison.group,
+        between: {
+          ...emptyState.differentialExpression.comparison.group.between,
+          cellSet: 'condition/condition-treated',
+          compareWith: 'condition/condition-control',
+          basis: 'louvain/cluster-a;',
+        },
+      },
+    },
+  },
+};
+
+const emptyStore = mockStore(emptyState);
+const filledStore = mockStore(filledState);
 
 describe('DiffExprManager', () => {
   beforeAll(async () => {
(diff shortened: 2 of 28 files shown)


src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
```diff
@@ -93,7 +93,26 @@ const store = mockStore({
...
+    },
+    comparison: {
+      type: 'between',
+      group: {
+        between: {
+          cellSet: null,
+          compareWith: null,
+          basis: null,
+        },
+      },
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx:115]: is there any reason why we are saving the `cellSet`, `compareWith` and `basis` under `group` and then under `between` instead of just doing this for example:

```
type: 'between',
cellSet: null,
compareWith: null,
basis: null,
```

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
```diff
@@ -159,22 +199,14 @@ describe('DiffExprCompute', () => {
...
     expect(form.find(Radio.Group).length).toEqual(1);
 
-    // Clicking on the second one should re-render the tool.
-    const { onChange } = form.find(Radio.Group).getElement().props;
-
-    onChange({ target: { value: 'within' } });
-
-    component.update();
-    form = component.find(Form);
-
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx]: if this piece of code is being deleted, how do we trigger a change in the component in order to test if a change is done?

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
```diff
@@ -149,6 +170,25 @@ describe('DiffExprCompute', () => {
...
+              basis: null,
+            },
+          },
+        },
+      },
+    });
+
     const component = mount(
       <Provider store={store}>
         <DiffExprCompute
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx:198]: do we still need to pass in `DiffExprCompute` `cellSets` as props? Since they are in the redux store, is there any reason why we can't read them directly from the `differentialExpression` store directly inside the `DiffExprCompute` component?

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
```diff
@@ -59,37 +82,36 @@ describe('DiffExprManager', () => {
   });
 
   it('on click of compute with changed parameters, DiffExprManager calls the results view', () => {
+    console.log(filledStore.getState());
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx]: don't forget to remove this

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/__snapshots__/DiffExprManager-regression-load-on-second-render.test.jsx.snap
```diff
@@ -13,9 +13,9 @@ exports[`DiffExprManager regression test -- diff exp would not reload after \`go
 Object {
   "payload": Object {
     "cellSets": Object {
-      "basis": "all",
-      "cellSet": "cluster-2",
-      "compareWith": "rest",
+      "basis": "cluster-a;",
+      "cellSet": "condition-treated",
+      "compareWith": "condition-control",
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/__snapshots__/DiffExprManager-regression-load-on-second-render.test.jsx.snap]: shouldn't this actually be formatted as `group/cluster-key` as we spoke about? Or is this the snapshot for the DifferentialExpression task that is being sent to the worker?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -21,18 +22,19 @@ const ComparisonType = Object.freeze({ between: 'between', within: 'within' });
...
   const [isFormValid, setIsFormValid] = useState(false);
-  const [selectedGroups, setSelectedGroups] = useState(cellSets);
-  const [type, setType] = useState(diffExprType || ComparisonType.between);
+  const comparisonGroup = useSelector((state) => state.differentialExpression.comparison.group);
+  const selectedComparison = useSelector((state) => state.differentialExpression.comparison.type);
 
-  // `between sampoles/groups` makes no sense if there is no metadata. First,
+
+  // `between samples/groups` makes no sense if there is no metadata. First,
   // assume there is and correct this if it's necessary.
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: reading from the code, it seems like we first assume there ISN'T metadata. If that is indeed the case, can you fix this comment to be accurate?

src/redux/actionTypes/differentialExpression.js
```diff
@@ -16,6 +16,19 @@ const DIFF_EXPR_LOADED = `${DIFF_EXPR}/loaded`;
 
 const DIFF_EXPR_ERROR = `${DIFF_EXPR}/error`;
 
+/**
+ * Sets the state for diffential expression settings
+ *
+ */
+const DIFF_EXPR_COMPARISON_TYPE_SET = `${DIFF_EXPR}/comparisonType`;
+const DIFF_EXPR_COMPARISON_GROUP_SET = `${DIFF_EXPR}/comparisonGroup`;
```
ivababukova, line 24: Any particular reason why we can't merge them and have one action about this?

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx:115]: So that both states of the views (`between` and `within`) can be saved, and so user can go back and forth and see their selected options still there

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx:198]: Nope, deleted in new commit

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx]: Hmm... I might have deleted it to check some stuff. I'll check it again,

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: I don't understand this one either, I was just fixing a typo :)) But I'll try to dig what that means

src/redux/actionTypes/differentialExpression.js
aerlaut: `comparisonType` fires when user changes the active tab. That does not mean there are settings changed, which will be fire `comparisonGroup`. So I think these 2 actions have to be separate.

ivababukova: By "the active tab" do you mean when the user changes the tab under `Tools` from `Gene List` to `Differential Expression`?

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/__snapshots__/DiffExprManager-regression-load-on-second-render.test.jsx.snap
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/__snapshots__/DiffExprManager-regression-load-on-second-render.test.jsx.snap]: updated in commit 5bbaf2a

src/redux/actionTypes/differentialExpression.js
aerlaut: No, I mean switching between the two radio buttons to switch the type of differential expression

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: it is about the default value for `hasMetadata` boolean which in here is set to `false`:

```
const [hasMetadata, setHasMetadata] = useState(false);
```

Judging by the code, by default we assume that there is no metadata, but the comment above says the opposite, so it should be corrected.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -50,42 +52,44 @@ const DiffExprCompute = (props) => {
...
-    // Update the hierarchies if a previously selected set were to be deleted.
-    setSelectedGroups(_.mapValues(selectedGroups, (cellSetKey) => {
-      if (!cellSetKey) {
-        return null;
-      }
-
-      const splitKey = cellSetKey.split('/')[1];
+    // If any key in group option is deleted, set option to null
+    Object.keys(comparisonGroup).forEach((type) => {
+      const groupDiff = {};
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: do we have a unit test for this case?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -199,23 +225,36 @@ const DiffExprCompute = (props) => {
...
+        // If comparisonType key is not available, add to group
+        if(!comparisonGroup[e.target.value]) {
+          dispatch(setComparisonGroup({
+            type: e.target.value,
+            cellSet: null,
+            compareWith: null,
+            basis: null,
+            ...comparisonGroup[e.target.value]
+          }))
+        }
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: I think we can go without dispatching `setComparisonGroup` -- what is the point of dispatching an action to save empty cell set selections?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: also, this code is quite hard to read and understand as it is. I would first extract the "check if a key in group option has deleted" part as a separate function. Then, in the use effect I would just have an if statement checking if that function returned any items that have to be nulled, and if it is true I would proceed with setting these keys to null

Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: My understanding is that this `groupDiff` is an object that contains the keys of all items that have to be made to null. If that is true, I suggest you rename `groupDiff` to something else that indicates better the purpose of that object, again for code readability purposes. For example, naming it `deletedKeys` would be maybe better.
Another thing is: does this `groupDiff` actually have to be an object? If all the keys in there are going to be set to `null`, maybe you could consider having it as a list and in that way you will have it lighter.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -50,42 +52,44 @@ const DiffExprCompute = (props) => {
...
     setHasMetadata(numMetadata > 0);
 
-    // Update the hierarchies if a previously selected set were to be deleted.
-    setSelectedGroups(_.mapValues(selectedGroups, (cellSetKey) => {
-      if (!cellSetKey) {
-        return null;
-      }
-
-      const splitKey = cellSetKey.split('/')[1];
+    // If any key in group option is deleted, set option to null
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: Also, I just tried doing that locally with these changes to see what is going to happen and I noticed that there is this console error appearing:

Screenshot 2021-01-21 at 10 12 10

Can you look into it and fix it? Errors and warnings on the console is something that needs looking at - ideally we should not have any errors or warnings.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -50,42 +52,44 @@ const DiffExprCompute = (props) => {
...
+    Object.keys(comparisonGroup).forEach((type) => {
+      const groupDiff = {};
 
-      if (cellSetKey === 'all' || cellSetKey === 'background' || splitKey === 'rest') {
-        return cellSetKey;
-      }
-
-      if (!properties[splitKey]) {
-        return null;
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: why is this code no longer needed?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -95,20 +99,40 @@ const DiffExprCompute = (props) => {
 
   // Validate form when the groups selected changes.
   useEffect(() => {
+
+    // Check if `compareWith` is in the same group as `cellSet`
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: why do we need to do this?

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -95,20 +99,40 @@ const DiffExprCompute = (props) => {
...
+        && comparisonGroup[selectedComparison].compareWith
+    ) {
+      const inSameGroup = comparisonGroup[selectedComparison].cellSet.split('/')[0] === comparisonGroup[selectedComparison].compareWith.split('/')[0]
+
+      if (!inSameGroup) {
+        dispatch(
+          setComparisonGroup({
+            ...comparisonGroup[selectedComparison],
+            type: ComparisonType.within,
+            compareWith: null,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: why do we need to pass in `null` in here?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -214,7 +230,11 @@ const VolcanoPlot = () => {
           description='Could not load differential expression data.'
           onClick={() => {
             dispatch(
-              loadDifferentialExpression(experimentId, config.diffExpData),
+              loadDifferentialExpression(
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:235]: great that you are fixing the volcano plot as well. Can you also remove the outdated cell set options from the initial state for the volcano plot? They are wrongly placed in the styling config, which is bad. This is the line that needs removing: https://github.com/biomage-ltd/ui/blob/master/src/redux/reducers/componentConfig/initialState.js#L88 Make sure that nothing reads from it, that the new part of the redux store that you are implementing is taken into an account only.

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx]: Previously, the click action changes state, which changes the comparison view. Now, the state is removed and state change is done through redux store. Actions dispatched to `mockStore`, does not change the state. So I removed this code and replaced the values in the store that is used in this test to a state after the radio button is clicked to simulate the clicking. However, this does not reflect the description of the test, which is to test the clicking. 

So, I recovered the code and simulate clicking better. Now the test checks for the action that is fired as the button is clicked, and then tests for the correct UI to be rendered.

Comment (ivababukova) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.test.jsx]: ok, that sounds good.

src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprManager.test.jsx]: Removed in new commit

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: It's fired when a non existent key is used to access an antd Tree component (? - it was fired by rc-tree, a library used by antd's Tree.js) e.g.`property[cellSetName]`. Checking for existence is replaced by using `property.hasOwnProperty(cellSetName)`, which doesn't fire the warning.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: changed in commit 11a74c2

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: Removed, as not relevant anymore per discussion with Vicky regarding in-group comparison

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: It's actually to initialize the `between` group. Actually it can be set as an initial value. However I was thinking for the case if we want to add another way to compare. But that may doesn't seem to be the case. I'll move it as an initial value.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: This was previous implementation for checking if a selected key was deleted. This is deleted because the way the selected keys are stored in has changed, from using state to redux and also a deeper object with each state (between/within) storing its own keys. Previously, there was only one.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: Per Vicky's comment on the needs to keep comparison in-group, not required anymore, so removed in latest commit.

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: To reset the value of the `compareWith`, select button (2nd select in within).

src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/components/differential-expression-tool/DiffExprCompute.jsx]: It's set as a key to make it easier to specifically add the key to the action payload by just spreading the object,

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:235]: Removed cellSets in initial config, checked no proble with the plot. However, because there is no cell selected, the plot's initial state is loading. I think this can be released as a new ticket.

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:235]: can you make a new ticket for that, Agi? Please send me a link to it once you make it.

src/redux/actionTypes/differentialExpression.js
```diff
@@ -16,6 +16,19 @@ const DIFF_EXPR_LOADED = `${DIFF_EXPR}/loaded`;
...
+
+/**
+ * Sets an error condition for a specific embedding.
+ */
+// const DIFF_EXPR_SET_COMPARISON_TYPE = `${}`
+
```
ivababukova: can you remove that commented out code?

aerlaut: deleted in newest commit

ivababukova (approved):
it's looking good, great job Agi!

PR #77, opened 2021-01-21
Embedding configuration to be saved to Redux

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-446

https://ui-wzpx3bx8u3r593588rdk5mwnmh.scp-staging.biomage.net/

N/A

Regarding the best way to store the default config, I have come up with a mixed approach of both considerations 1 and 2 in the ticket.

The UI will store the default state of the embedding config, but we will want to have a default state saved in DynamoDB as well. This is for two reasons:

(a) if the config were to change in a non-destructive way (e.g. a new field or parameter is added), we want to make sure we can automatically 'upgrade' experiments the next time the user does some pre-processing, and without a migration script;
(b) we will still load the state from DynamoDB for purposes of traceability. If in the future we were to change a default value to something else, experiments that have identical settings files in DynamoDB would run differently before and after the update, so we would have a reproducibility problem.

As such, the proposed way to implement this is for the reducer to perform the creation of the object in the order of `initial state` -> `current state` -> `state from DynamoDB`. This meets the requirements of (a). We should also persist the config in DynamoDB from the data ingest script, which meets the requirements of (b).

# Changes
### Code changes

Added a new redux store and appropriate actions to save embedding configurations globally.

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
image

### DIFF
--- a/src/redux/actionTypes/experimentSettings.js
+++ b/src/redux/actionTypes/experimentSettings.js
@@ -1,0 +1,25 @@
+const EXPERIMENT_SETTINGS = 'experimentSettings';
+
+/**
+ * Turns on the loading state for a given processing component.
+ */
+const EXPERIMENT_SETTINGS_PROCESSING_LOADING = `${EXPERIMENT_SETTINGS}/loadingProcessing`;
+
+/**
+ * Sets the state of the processing component to a state displayable by the system.
+ */
+const EXPERIMENT_SETTINGS_PROCESSING_LOADED = `${EXPERIMENT_SETTINGS}/loadedProcessing`;
+
+/**
+ * Updates the processing component's configuration.
+ */
+const EXPERIMENT_SETTINGS_PROCESSING_UPDATE = `${EXPERIMENT_SETTINGS}/updateProcessing`;
+
+const EXPERIMENT_SETTINGS_PROCESSING_ERROR = `${EXPERIMENT_SETTINGS}/errorProcessing`;
+
+export {
+  EXPERIMENT_SETTINGS_PROCESSING_LOADING,
+  EXPERIMENT_SETTINGS_PROCESSING_LOADED,
+  EXPERIMENT_SETTINGS_PROCESSING_UPDATE,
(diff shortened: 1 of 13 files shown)


ivababukova:
Love the animals :d In terms of the details about the changes that you listed, I will add us a ticket for the `We should also persist the config in DynamoDB from the data ingest script, which meets the requirements of (b).` part

marcellp:
> Love the animals :d In terms of the details about the changes that you listed, I will add us a ticket for the `We should also persist the config in DynamoDB from the data ingest script, which meets the requirements of (b).` part

My understanding is that it is already a ticket, BIOMAGE-450.

src/redux/actionTypes/experimentSettings.js
```diff
@@ -0,0 +1,25 @@
+const EXPERIMENT_SETTINGS = 'experimentSettings';
```
ivababukova, line 1: just to make sure I got it right, the design of these actions is with the thought to use them for any settings that will influence the data in the experiment, for instance we will use them for the data integration module, for the QC filters too. Is that correct?

marcellp: Correct, also see the `initialState` of the reducer. Idea is to add ones under `configureEmbedding` for all the different filters, like how it is in Agi's PR under `processingConfig`: https://github.com/biomage-ltd/api/pull/41/files#diff-0d6e67500a03df911efef7905b8624278ab319098301fed58c2709526ebde5fcR61

marcellp: I also envisioned this store to be used for more general experiment settings/metadata in the future, such as the name of the PI, team, organism data, last modified, name of experiment, what genome it was aligned against, etc.

ivababukova: great, this will be very needed in the future

marcellp:
Unit tests added.

ivababukova (approved):
looks great!

PR #78, opened 2021-01-22
Multiple grouping in heatmap

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-465?atlOrigin=eyJpIjoiNDM2OGQ5YTE0NTA0NDg0ZGE3ZjQ0MGY3YzY1ZTExYzIiLCJwIjoiaiJ9

https://ui-7br81kzwi543663r5k1zfdagqy.scp-staging.biomage.net/

- Dropdown to add/remove groupby's blocks the view of the underlying selection menu, perhaps display it to the side instead?

- Ordering looks weird when a cell belongs to more than one group in a same grouping (this is something from before this PR's changes), should we do something about that?

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/__test__/components/reorderable-list/ReorderableList.test.jsx
+++ b/src/__test__/components/reorderable-list/ReorderableList.test.jsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { mount, configure } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
+import ReorderableList from '../../../components/ReorderableList';
+
+configure({ adapter: new Adapter() });
+
+describe('ReorderableList', () => {
+  let onChangeCount;
+  let latestOnChangeListReceived;
+
+  const onChangeMock = (list) => {
+    onChangeCount += 1;
+    latestOnChangeListReceived = list;
+  };
+
+  let reorderableListMock;
+
+  const leftItemMock = (itemData) => (itemData.key);
+  const rightItemMock = () => (<br />);
+
+  let component;
+
+  beforeEach(() => {
+    onChangeCount = 0;
+    latestOnChangeListReceived = null;
+
+    reorderableListMock = [{ key: 1 }, { key: 2 }, { key: 3 }];
+
+    component = mount(
+      <ReorderableList
+        onChange={onChangeMock}
+        defaultList={reorderableListMock}
+        leftItem={leftItemMock}
+        rightItem={rightItemMock}
+      />,
+    );
+  });
+
+  it('renders correctly', () => {
+    const items = component.find('Item');
+
+    // renders the three items
+    expect(items.length).toEqual(3);
+
+    console.log('items.at(0)');
+    console.log(items.at(0).debug());
+
+    // in the correct order and with the correct items
+    expect(items.at(0).text()).toEqual('1');
+    expect(items.at(1).text()).toEqual('2');
+    expect(items.at(2).text()).toEqual('3');
+  });
+
+  it('reorders correctly', () => {
+    // when pressing button down on the first item
+    const item1 = component.find('Item').at(0);
+    act(() => (item1.find('Button').at(1).simulate('click')));
+
+    component.update();
+
+    const items = component.find('Item');
+    // the new list order is correct
+    expect(items.at(0).text()).toEqual('2');
+    expect(items.at(1).text()).toEqual('1');
+    expect(items.at(2).text()).toEqual('3');
+
+    // and an onChange is fired
+    expect(onChangeCount).toEqual(1);
+  });
+});
--- a/src/__test__/components/reorderable-list/ReorderableList.test.jsx
+++ b/src/__test__/components/reorderable-list/ReorderableList.test.jsx
@@ -44,9 +44,6 @@ describe('ReorderableList', () => {
     // renders the three items
     expect(items.length).toEqual(3);
 
-    console.log('items.at(0)');
-    console.log(items.at(0).debug());
-
     // in the correct order and with the correct items
     expect(items.at(0).text()).toEqual('1');
     expect(items.at(1).text()).toEqual('2');
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
@@ -1,11 +1,15 @@
 import React from 'react';
-import { shallow, configure, mount, key } from 'enzyme';
+import {
+  shallow, configure, mount,
+} from 'enzyme';
 import { Provider } from 'react-redux';
 import waitForActions from 'redux-mock-store-await-actions';
 import Adapter from 'enzyme-adapter-react-16';
 import thunk from 'redux-thunk';
 import configureMockStore from 'redux-mock-store';
-import { Button, Menu, Dropdown, Space } from 'antd';
+import {
+  Button, Dropdown,
+} from 'antd';
 import HeatmapGroupBySettings from '../../../../../../../pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings';
 import { UPDATE_CONFIG } from '../../../../../../../redux/actionTypes/componentConfig';
 
@@ -100,12 +104,14 @@ describe('HeatmapGroupBySettings', () => {
     // Should be rendered.
     expect(component.find('HeatmapGroupBySettings').length).toEqual(1);
 
-    // with two items
-    const groupByItems = component.find('Item');
-    expect(groupByItems.length).toEqual(2);
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
 
-    // the second item should be samples (the first one is the dropdown)
-    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
+    // with one ite
+    expect(groupByItems.length).toEqual(1);
+
+    // the item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
 
     // there's a dropdown
     const dropdown = component.find(Dropdown);
@@ -153,19 +159,21 @@ describe('HeatmapGroupBySettings', () => {
     store.clearActions();
     expect(store.getActions().length).toEqual(0);
 
-    // there should be 3 group by items
-    const groupByItems = component.find('Item');
-    expect(groupByItems.length).toEqual(3);
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
 
-    // the second item should be samples (the first one is the dropdown)
-    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
-    // the third item should be louvain (the first one is the dropdown)
-    expect(groupByItems.at(2).childAt(0).childAt(0).text()).toEqual('louvain clusters');
+    // there should be 2 group by items
+    expect(groupByItems.length).toEqual(2);
+
+    // the first item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    // the second item should be louvain
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('louvain clusters');
 
     // when the groupby is clicked again
     buttons.at(0).simulate('click');
     component.update();
-    
+
     await waitForActions(store, [UPDATE_CONFIG]);
 
     // The store should update.
@@ -176,11 +184,10 @@ describe('HeatmapGroupBySettings', () => {
     expect(secondAction).toMatchSnapshot();
 
     // there should be 2 group by items
-    const updatedGroupByItems = component.find('Item');
-    expect(groupByItems.length).toEqual(3);
+    expect(groupByItems.length).toEqual(2);
 
-    // and the second item should be samples (the first one is the dropdown)
-    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
+    // and the first item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
   });
 
   test('interacting with the groupby reorder options will trigger the appropriate actions', async () => {
@@ -206,23 +213,23 @@ describe('HeatmapGroupBySettings', () => {
     addButtons.at(0).simulate('click');
     component.update();
 
-    // there should be 3 group by items
-    const groupByItems = component.find('Item');
-    expect(groupByItems.length).toEqual(3);
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
+
+    // there should be 2 group by items
+    expect(groupByItems.length).toEqual(2);
 
     // clear store from this action (not being tested here)
     await waitForActions(store, [UPDATE_CONFIG]);
     store.clearActions();
 
-    // the second item should be samples (the first one is the dropdown)
-    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
-    // the third item should be louvain (the first one is the dropdown)
-    expect(groupByItems.at(2).childAt(0).childAt(0).text()).toEqual('louvain clusters');
+    // the first item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    // the second item should be louvain (the first one is the dropdown)
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('louvain clusters');
 
     // after a click for sample to go down
-    const reorderButtonsSample = groupByItems.at(1).find('Button');
-    console.log('reorderButtonsSample');
-    console.log(reorderButtonsSample.debug());
+    const reorderButtonsSample = groupByItems.at(0).find('Button');
     reorderButtonsSample.at(1).simulate('click');
 
     // the store should update.
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
@@ -1,12 +1,12 @@
 import React from 'react';
-import { configure, mount } from 'enzyme';
+import { shallow, configure, mount, key } from 'enzyme';
 import { Provider } from 'react-redux';
+import waitForActions from 'redux-mock-store-await-actions';
 import Adapter from 'enzyme-adapter-react-16';
 import thunk from 'redux-thunk';
 import configureMockStore from 'redux-mock-store';
-import { Radio } from 'antd';
+import { Button, Menu, Dropdown, Space } from 'antd';
 import HeatmapGroupBySettings from '../../../../../../../pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings';
-
 import { UPDATE_CONFIG } from '../../../../../../../redux/actionTypes/componentConfig';
 
 jest.mock('localforage');
@@ -100,19 +100,27 @@ describe('HeatmapGroupBySettings', () => {
     // Should be rendered.
     expect(component.find('HeatmapGroupBySettings').length).toEqual(1);
 
-    // With two buttons.
-    const buttons = component.find('Radio');
-    expect(buttons.length).toEqual(2);
+    // with two items
+    const groupByItems = component.find('Item');
+    expect(groupByItems.length).toEqual(2);
 
-    // With the right keys.
-    expect(buttons.at(0).props().value).toEqual('louvain');
-    expect(buttons.at(1).props().value).toEqual('sample');
+    // the second item should be samples (the first one is the dropdown)
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
 
-    // The selected one should be `sample`.
-    expect(component.find(Radio.Group).props().value).toEqual('sample');
+    // there's a dropdown
+    const dropdown = component.find(Dropdown);
+    const submenu = shallow(<div>{dropdown.prop('overlay')}</div>);
+    const subMenuItems = submenu.find('MenuItem');
+
+    // With two items.
+    expect(subMenuItems.length).toEqual(2);
+
+    // each of which should have the right names.
+    expect(subMenuItems.at(0).childAt(0).text()).toEqual('louvain clusters');
+    expect(subMenuItems.at(1).childAt(0).text()).toEqual('Sample');
   });
 
-  it('clicking on the other button will trigger the appropriate action', async () => {
+  test('interacting with the groupby add/remove options will trigger the appropriate actions', async () => {
     const store = mockStore({
       ...initialState,
     });
@@ -126,14 +134,101 @@ describe('HeatmapGroupBySettings', () => {
     // Should be rendered.
     expect(component.find('HeatmapGroupBySettings').length).toEqual(1);
 
-    // When a separate group by is selected...
-    const { onChange } = component.find(Radio.Group).props();
-    onChange({ target: { value: 'louvain' } });
+    const dropdown = component.find(Dropdown);
+    const submenu = shallow(<div>{dropdown.prop('overlay')}</div>);
+    const subMenuItems = submenu.find('MenuItem');
+
+    // When the other group by is clicked...
+    const buttons = subMenuItems.find(Button);
+    buttons.at(0).simulate('click');
+    component.update();
+
+    await waitForActions(store, [UPDATE_CONFIG]);
 
     // The store should update.
     expect(store.getActions().length).toEqual(1);
-    const action = store.getActions()[0];
-    expect(action.type).toBe(UPDATE_CONFIG);
-    expect(action).toMatchSnapshot();
+    const firstAction = store.getActions()[0];
+    expect(firstAction.type).toBe(UPDATE_CONFIG);
+    expect(firstAction).toMatchSnapshot();
+    store.clearActions();
+    expect(store.getActions().length).toEqual(0);
+
+    // there should be 3 group by items
+    const groupByItems = component.find('Item');
+    expect(groupByItems.length).toEqual(3);
+
+    // the second item should be samples (the first one is the dropdown)
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
+    // the third item should be louvain (the first one is the dropdown)
+    expect(groupByItems.at(2).childAt(0).childAt(0).text()).toEqual('louvain clusters');
+
+    // when the groupby is clicked again
+    buttons.at(0).simulate('click');
+    component.update();
+    
+    await waitForActions(store, [UPDATE_CONFIG]);
+
+    // The store should update.
+    expect(store.getActions().length).toEqual(1);
+    const secondAction = store.getActions()[0];
+
+    expect(secondAction.type).toBe(UPDATE_CONFIG);
+    expect(secondAction).toMatchSnapshot();
+
+    // there should be 2 group by items
+    const updatedGroupByItems = component.find('Item');
+    expect(groupByItems.length).toEqual(3);
+
+    // and the second item should be samples (the first one is the dropdown)
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
+  });
+
+  test('interacting with the groupby reorder options will trigger the appropriate actions', async () => {
+    const store = mockStore({
+      ...initialState,
+    });
+
+    component = mount(
+      <Provider store={store}>
+        <HeatmapGroupBySettings experimentId='123' width={200} height={200} />
+      </Provider>,
+    );
+
+    // Should be rendered.
+    expect(component.find('HeatmapGroupBySettings').length).toEqual(1);
+
+    const dropdown = component.find(Dropdown);
+    const submenu = shallow(<div>{dropdown.prop('overlay')}</div>);
+    const subMenuItems = submenu.find('MenuItem');
+
+    // Add a louvain group by
+    const addButtons = subMenuItems.find(Button);
+    addButtons.at(0).simulate('click');
+    component.update();
+
+    // there should be 3 group by items
+    const groupByItems = component.find('Item');
+    expect(groupByItems.length).toEqual(3);
+
+    // clear store from this action (not being tested here)
+    await waitForActions(store, [UPDATE_CONFIG]);
+    store.clearActions();
+
+    // the second item should be samples (the first one is the dropdown)
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('Sample');
+    // the third item should be louvain (the first one is the dropdown)
+    expect(groupByItems.at(2).childAt(0).childAt(0).text()).toEqual('louvain clusters');
+
+    // after a click for sample to go down
+    const reorderButtonsSample = groupByItems.at(1).find('Button');
+    console.log('reorderButtonsSample');
+    console.log(reorderButtonsSample.debug());
+    reorderButtonsSample.at(1).simulate('click');
+
+    // the store should update.
+    await waitForActions(store, [UPDATE_CONFIG]);
+    expect(store.getActions().length).toEqual(1);
+    const secondAction = store.getActions()[0];
+    expect(secondAction).toMatchSnapshot();
   });
 });
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
@@ -107,7 +107,7 @@ describe('HeatmapGroupBySettings', () => {
     const reorderableList = component.find('ReorderableList');
     const groupByItems = reorderableList.find('Item');
 
-    // with one ite
+    // with one item
     expect(groupByItems.length).toEqual(1);
 
     // the item should be samples
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
@@ -74,7 +74,7 @@ const initialState = {
   componentConfig: {
     interactiveHeatmap: {
       config: {
-        groupedTrack: 'sample',
+        groupedTracks: ['sample'],
         selectedTracks: ['louvain'],
       },
     },
--- a/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
+++ b/src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
@@ -97,7 +97,7 @@ describe('HeatmapGroupBySettings', () => {
 
     component = mount(
       <Provider store={store}>
-        <HeatmapGroupBySettings experimentId='123' />
+        <HeatmapGroupBySettings componentType='interactiveHeatmap' experimentId='123' />
       </Provider>,
     );
 
@@ -111,7 +111,7 @@ describe('HeatmapGroupBySettings', () => {
     expect(groupByItems.length).toEqual(1);
 
     // the item should be samples
-    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    expect(groupByItems.at(0).text()).toEqual('Sample');
 
     // there's a dropdown
     const dropdown = component.find(Dropdown);
@@ -122,8 +122,8 @@ describe('HeatmapGroupBySettings', () => {
     expect(subMenuItems.length).toEqual(2);
 
     // each of which should have the right names.
-    expect(subMenuItems.at(0).childAt(0).text()).toEqual('louvain clusters');
-    expect(subMenuItems.at(1).childAt(0).text()).toEqual('Sample');
+    expect(subMenuItems.at(0).find('div').text()).toEqual('louvain clusters');
+    expect(subMenuItems.at(1).find('div').text()).toEqual('Sample');
   });
 
   test('interacting with the groupby add/remove options will trigger the appropriate actions', async () => {
@@ -133,7 +133,7 @@ describe('HeatmapGroupBySettings', () => {
 
     component = mount(
       <Provider store={store}>
-        <HeatmapGroupBySettings experimentId='123' width={200} height={200} />
+        <HeatmapGroupBySettings componentType='interactiveHeatmap' experimentId='123' width={200} height={200} />
       </Provider>,
     );
 
@@ -159,6 +159,7 @@ describe('HeatmapGroupBySettings', () => {
     store.clearActions();
     expect(store.getActions().length).toEqual(0);
 
+    component.update();
     const reorderableList = component.find('ReorderableList');
     const groupByItems = reorderableList.find('Item');
 
@@ -166,9 +167,9 @@ describe('HeatmapGroupBySettings', () => {
     expect(groupByItems.length).toEqual(2);
 
     // the first item should be samples
-    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    expect(groupByItems.at(0).text()).toEqual('Sample');
     // the second item should be louvain
-    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('louvain clusters');
+    expect(groupByItems.at(1).text()).toEqual('louvain clusters');
 
     // when the groupby is clicked again
     buttons.at(0).simulate('click');
@@ -187,7 +188,7 @@ describe('HeatmapGroupBySettings', () => {
     expect(groupByItems.length).toEqual(2);
 
     // and the first item should be samples
-    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    expect(groupByItems.at(0).text()).toEqual('Sample');
   });
 
   test('interacting with the groupby reorder options will trigger the appropriate actions', async () => {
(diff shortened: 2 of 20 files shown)


src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx:1]: I added to this component a useRef to avoid the initial render triggering a `dispatch()` of an updatePlotConfig action.

I believe this is not really a big problem, but it does look a little bit weird now compared to before, because now each render of the heatmap plot is very easy to notice (due to the random sampling of cells from each cell set), so the plot would change visibly when these settings were opened initially.

I also did this in the groupBy settings, which had the same issue.

ivababukova:
@cosa65 I just checked the link in staging and the UX of "group by" does not look like what is displayed on the wireframes (slide 31). Are you sure that you deployed this PR in staging?

cosa65:
> @cosa65 I just checked the link in staging and the UX of "group by" does not look like what is displayed on the wireframes (slide 31). Are you sure that you deployed this PR in staging?

@ivababukova  About this, I redeployed and it has an updated version. I did another change with the groupby extra dropdown being opened by hover instead of click but for some reason even though ui is unpinned it didn't update, so I'll do another stagin environment and kill this one once the other is up.

ivababukova:
@vickymorrison can we have your eyes on this in terms of UX and usability?

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
```diff
@@ -1,12 +1,12 @@
 import React from 'react';
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx:1]: Martin, do you develop with the UI open as a workspace in VsCode, so that the code is auto-formatted and corrected after save? I just checked out your changes locally and this file looks full of linting errors (commas, formatting, etc.)

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -1,20 +1,23 @@
-import React from 'react';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:1]: Martin, do you develop with the UI open as a workspace in VsCode, so that the code is auto-formatted and corrected after save? I just checked out your changes locally and this file looks full of linting errors (commas, formatting, etc.)

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -29,26 +32,160 @@ const HeatmapGroupBySettings = () => {
...
+    if (index === -1 || index >= source.length - 1) {
+      return arr;
+    }
+
+    const el = arr[index];
+    arr[index] = arr[index + 1];
+    arr[index + 1] = el;
+
+    return arr;
+  };
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx]: this and lots of other code is essentially duplicated code from the `HeatmapMetadataTrackSettings` component in order to create this component with the up and down arrows that is used in here and in `HeatmapMetadataTrackSettings`. 
I would suggest that you extract this in a new, separate component. If possible, make it generic, like `ColorPicker`, `FocusButton` for example that do not depend on specific state in Redux. In that way, we would be able to easily reuse it anywhere in the UI in the future. If it can be made generic, place it in the place where we keep such components, which is in `src/components`. Make sure that you use this new components in `HeatmapGroupBySettings` and in  `HeatmapMetadataTrackSettings` files.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -29,26 +32,160 @@ const HeatmapGroupBySettings = () => {
...
-          <Radio style={radioStyle} value={cellSet.key}>
-            {cellSet.name}
-          </Radio>
-        ))}
-      </Radio.Group>
-    </div>
+      <Space direction='vertical'>
+        <Dropdown overlay={menu}>
+          <div style={{ padding: '7px', border: '1px solid rgb(238,238,238)' }}>
+            Select the parameters to group by
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: There is a usability issue with the functionality of this: After I have selected the parameters to group by, the menu does not disappear and show me my selections until I hover over the "group by" option in the main heatmap options menu. If I go out of the `Select the parameters to group by` window, the whole thing closes and then I have to go and click from scratch. I am not sure how exactly it should behave, but @marcellp usually has a good feel about these things, he could be helpful here with some advice.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx:1]: Good spot about unneeded triggering of `dispatch`, however this doesn't seem like the right way to avoid it. The two `useEffect` functions within which you have placed the `isInitialRenderRef.current` check to avoid executing, get triggered because whatever is in the square (`[]`) brackets for the respective useEffect has been changed, which in this case is `cellSets.hierarchy` and `trackData`. In order to prevent the trigger of the dispatch for `updatePlotConfig`, you should look at the relevant `useEffect` function and prevent `trackData` from changing on initial render. In that way, the dispatch will not happen.

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx:1]: I would recommend that you read a bit more about `useEffect` and initial component rendering, it will be very helpful for this task.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -18,23 +18,58 @@ const HeatmapMetadataTrackSettings = () => {
...
+    return _.unionBy(
+      trackData,
+      getCellSets(
+        ['cellSets', 'metadataCategorical'],
+      ).map(
+        (data) => ({ selected: selectedTracks.includes(data.key), key: data.key }),
+      ),
+      'key',
+    )
+  };
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx]: Looking at `getInitialTrackData` and the `getUpdatedTrackData` functions, they are almost the same apart from the union and `trackData` parameter. You could have only one function for getting trackData, which will be with the same contents as `getUpdatedTrackData` at the moment, and restore the initial default for trackData which was an empty list.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx:1]: Hmm, actually you are right, I had a read around that people tend to use useRef for this: https://stackoverflow.com/questions/53253940/make-react-useeffect-hook-not-run-on-initial-render, https://medium.com/swlh/prevent-useeffects-callback-firing-during-initial-render-the-armchair-critic-f71bc0e03536

In terms of the weirdness that you mention that is caused by random sampling, double check with Vicky if it is a problem.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+  const positionInGrouping = (cellSetNames, cell) => {
+    let position = 0;
+
+    for (const { key: cellSetName } of cellSetNames) {
+      if (cellsInSet(cellSetName).has(cell)) {
+        return position;
+      }
+
+      position += 1;
+    }
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: there is a warning displayed for this piece of code -- you should use array iterations instead of a loop?

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+
+    // If not in grouping set it in the end
+    return position + 1;
+  };
+
+  const cellsInSet = (cellSetName) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Since this is a getter function, it will get more readable if you call it  `getCellsInSet`.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
-    // Get how many cells we are sampling.
-    let total = 0;
+    let allCellsInSets = new Set();
 
-    // Create an object for storing the cells grouped by `groupBy`.
-    const groupedCells = {};
+    Object.values(arrayOfcellSetNamesByGroupByKey).forEach((arrayOfSetNames) => {
+      // Add each set of cells into 
+      arrayOfSetNames.forEach((setNameObj) => {
+        const cellSet = cellsInSet(setNameObj.key);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: calling this `cellSet` is very ambiguous - we already use that name to refer to the object that holds all cell set information. Maybe name `cellSetIds` would be better?

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+    return position + 1;
+  };
+
+  const cellsInSet = (cellSetName) => {
+    return properties[cellSetName].cellIds;
+  };
+
+  ///////////////////////////////////////////////////////////////////////////////////////////
+  // Returns a set of all cells relevant to the experiment (so the ones belonging to one or more of the enabled sets and not hidden).
+  const getAllRelevantCellIds = (arrayOfcellSetNamesByGroupByKey) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: instead of this name and the comment above, it will be more clean and easy to read if you just rename function `getAllRelevantCellIds` to `getNonHiddenCellIds`. Then the comment above won't be necessary.
The name `arrayOfcellSetNamesByGroupByKey` is confusing too -- I took a peak in the rest of the code and this is essentially a dictionary or each cell set and its children. I also noticed that this `cellSetNamesByGroupByKey` that is computed in line 213 in this file is not used anywhere else. In that case, I would not add it as a parameter to the `getAllRelevantCellIds` function at all and move line 213 to be executed here.
In general, always try to write the code in such a way, that writing comments is not necessary. For this to happen, naming variables clearly and consistently is key.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+        const cellSet = cellsInSet(setNameObj.key);
+        // Union of allCellsInSets and cellSet
+        allCellsInSets = new Set([...allCellsInSets, ...cellSet]);
+      });
+    });
 
-    // Find the `groupBy` root node.
-    const rootNode = hierarchy.filter((clusters) => clusters.key === groupBy);
+    // Only work with non-hidden cells.
+    const shownCells = new Set([...allCellsInSets].filter(x => !hiddenCellIds.has(x)));
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: for naming consistency, I would rename `shownCells` to `shownCellIds`, or even better `nonHiddenCellIds` so that it matches with the function name after you rename it

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
-      return [];
+    return shownCells;
+  }
+
+  const getCellSetNamesDict = (rootNodes) => {
+    const cellSetNamesByGroupByKey = {};
+    rootNodes.forEach((currentRootNode) => cellSetNamesByGroupByKey[currentRootNode.key] = currentRootNode.children);
+
+    return cellSetNamesByGroupByKey
+  };
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: there are lots of small functions in this file that are about massaging the cell sets data in various ways and forms in order to make their data structure efficient. This function is an example, but there are other functions in the other heatmap components too. These functions should be extracted into a separate file and only imported and called in here and other files - this is not functionality specific to the heatmap plot and it does not fit for these cell set data manipulation functions to be here. 
Take a look in `utils/cellSetOperations.js` and evaluate if this is a suitable place for these functions and if so move them there.

vickymorrison:
In terms of UX, I am not finding the transition from the dropdown manu back to the list view very easy. E.g. if the user has selected 3 items from the dropdown menu and then wants to view the list and reorder them it is not very easy to make the dropdown menu go away. I ended up coming out of the group by menu in order to do this. Can you make it so that the user can click on the dropdown menu to minimise it?

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
 
-      if (!cells) {
-        return;
+        if (onePosition < otherPosition) {
+          return -1;
+        }
+
+        if (onePosition > otherPosition) {
+          return 1;
+        }
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: what does this do? What is `onePosition` and `otherPosition`?

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+  }
+
+  const getCellSetNamesDict = (rootNodes) => {
+    const cellSetNamesByGroupByKey = {};
+    rootNodes.forEach((currentRootNode) => cellSetNamesByGroupByKey[currentRootNode.key] = currentRootNode.children);
+
+    return cellSetNamesByGroupByKey
+  };
+
+  const downsampled = (cellIds) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: this seems like a very heavy rewrite - I wonder whether there isn't an easier way to implement this that reuses the downsample function that we already have in master, but leverage it to be used across all clusters, not per cluster

cosa65:
> In terms of UX, I am not finding the transition from the dropdown manu back to the list view very easy. E.g. if the user has selected 3 items from the dropdown menu and then wants to view the list and reorder them it is not very easy to make the dropdown menu go away. I ended up coming out of the group by menu in order to do this. Can you make it so that the user can click on the dropdown menu to minimise it?
@vickymorrison 

so just to be sure, the issue with this menu could be fixed in two ways:
1 - it could be changed so that it is triggered by a click instead or
2 - Its hover function could be kept (since this is the default trigger on other menus of this kind) and its menu could be displayed on the side instead of over the other menu.

we are going with option 1 then?

vickymorrison:
> > In terms of UX, I am not finding the transition from the dropdown manu back to the list view very easy. E.g. if the user has selected 3 items from the dropdown menu and then wants to view the list and reorder them it is not very easy to make the dropdown menu go away. I ended up coming out of the group by menu in order to do this. Can you make it so that the user can click on the dropdown menu to minimise it?
> > @vickymorrison
> 
> so just to be sure, the issue with this menu could be fixed in two ways:
> 1 - it could be changed so that it is triggered by a click instead or
> 2 - Its hover function could be kept (since this is the default trigger on other menus of this kind) and its menu could be displayed on the side instead of over the other menu.
> 
> we are going with option 1 then?

On my screen (a laptop) there is no space to the right as the group by menu is already near the edge of the page. So I suggest going with option 1 where the user has to click the dropdown menu - like they do in the Differential Expression dropdown menus.
Screenshot 2021-01-25 at 15 06 25

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Correct, the appropriate way to do this would be:

```
cellSetNames.forEach(({key: cellSetName}) => { ... });
```

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
-    const sample = [];
+    // Transform the groupByRootNodes element into a dictionary/object to make its searches by key faster 
+    // (this is important since it will be used in the sorting comparator) 
+    const cellSetNamesByGroupByKey = getCellSetNamesDict(groupByRootNodes);
+
+    // Get all cellIds from cells that will show up (so those that aren't hidden and are in an enabled groupBy)
+    const cellIdsSet = getAllRelevantCellIds(cellSetNamesByGroupByKey);
+    const cellIdsArray = Array.from(cellIdsSet);
+
+    const cellIdsSample = downsampled(cellIdsArray);
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: I thought the idea here was that cells would be sampled randomly, but *after* sorting in a way such that each bottom-level grouping would have their proportions preserved, not just a literal random sampling of all cells, *followed by* the grouping.

If you sample all cells randomly, you will have most of the cells from the biggest clusters, and almost none from the smaller ones, which is exactly what we want to avoid. Maybe I'm misunderstanding, @vickymorrison can probably verify.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
+      var randomIndex = Math.floor(Math.random() * cellIds.length);
+      const element = cellIds.splice(randomIndex, 1);
 
-      total += shownCells.length;
+      downsampledCellIds.push(element[0]);
+    }
 
-      groupedCells[key] = shownCells;
-    });
+    return downsampledCellIds;
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: You can probably use [_.sampleSize](https://lodash.com/docs/4.17.15#sampleSize) in place of this custom function.

marcellp (dismissed):

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: I am not sure how we could use it across all groupBy's because we would have to keep in mind the intersections of cells between each group by, and if we went into each groupBy then we would run into many issues when trying to merge these different downsampled groups of cells into one downsample (like we can't guarantee that the merged downsampled set is actually less than maxCells so we would have to do a random downsampling after anyways).

This is a problem I ran into and asked about in the bioinformatics slack I believe, the random downsampling was what most people agreed with.

Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: I fell like a solution trying to use the same logic of the original downsampling function would actually involve a heavy rewrite anyways beause we would have to introduce some kind of relationship between the cell groups to know which cells to pick to keep representativeness.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:1]: I do, in the linux computer I got the auto formatting, but here it seems to not be working.

I'll try to make it work and fix this.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: I actually talked with vicky about how to go about this and I think she would prefer to just have the trigger action be click instead.

The other option I had thought about to deal with this was having the dropdown to open on the side instead of over the already existing component.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: yes, I was trying to rename everything from rootNode to this but it is probably too wordy.

By the way, it doesn't only filter out hidden cellIds, it also avoids picking cellIds that are not in any cellSet defined by the enabled groupBys, so nonHidden wouldn't really represent this function either.

I could just separate the functionality and have one function that filters out hidden cellIds and another with this name: `getCellIdsInGroupBys`.

Let me know what you think about this.

Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: and about the other comment you're right, it is better to just use the `getCellSetNamesDict` inside the function that needs it.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: it compares the ordering of both cells in the current groupby.
If both cells are in the same cellSet defined in the current groupby we move over to check their ordering according to the following groupby.

I am going to do some renaming on this comparator to make it more readable.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: I am not sampling completely randomly, first I am picking only the cells that belong to one or more of the enabled group bys and running the downsample over that.

That was what I interpreted could be done to reduce the impact of the randomness on how representative the downsampled set is.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:1]: Thanks Martin, the auto formatting is super helpful, make sure that you have it working.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: I think that click will work, provided that the menu for selecting parameters collapses back after click, so that the user can see easily what they have selected showing up/disappearing.

I was actually wondering why did we not go with exactly what was implemented for the "metadata tracks" menu with the switches (see screenshot). From my perspective at least, they seem much more intuitive and then the whole usability problem we are discussing here disappears. @vickymorrison what are the reasons for why there we have switches and here we don't?
Screenshot 2021-01-26 at 09 58 40

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Thanks for clarification Martin, I see why you used the `relevant` word in the name now. Maybe `getEnabledCellIds` would be better name then? Or anything else if you think about that is not generic and quite short

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Make sense, I understand your reasoning now and I agree that this is better. Good to hear that you have considered it, if it doesn't work well, that's fine.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (vickymorrison) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: My original wireframe design was exactly as you are suggesting, Iva. But @marcellp convinced me that the usability of that design was flawed - the switches imply that something is shown/hidden (like for the metadata tracks) and we thought that the function of the switches would be less clear and intuitive to users. This new design, as Martin is implementing, has much better usability and will also be better for large datasets with many metadata tracks.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -121,83 +122,148 @@ const HeatmapPlot = (props) => {
...
-    // Iterate over each child node.
-    children.forEach(({ key }) => {
-      // Only work with non-hidden cells.
-      const shownCells = Array.from(
-        properties[key].cellIds,
-      ).filter(
-        (id) => !hiddenCellIds.has(id),
-      );
+    // Take maxCells amount of cells, picking each one randomly
+    for (let i = 0; i < maxCells; i++) {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: Better use _.sampleSize as Marcell recommended below, but in general I wanted to say that for readability reasons we tend to use `forEach`, `filter` and `map` when traversing arrays and doing some manipulations in these. I don't know if @marcellp has any different opinion, but my feel is that this is the recommended way to do it in javascript instead of for loops and I would recommend you to use these instead of classical `for` loops. 
In terms of which one of them is more efficient, even though `for` is more efficient than `forEach`, browsers have become very good at optimising for `forEach` statements and practically, these often have better performance (I was reading this: https://stackoverflow.com/questions/43821759/why-array-foreach-is-slower-than-for-loop-in-javascript)

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: I see, I guess this is something that we will verify with the users. Are we having tow different designs for this for `metadata tracks` and `group by`, because we want to test with users which one is better?

Comment (vickymorrison) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: Yes, this is something that we can test and observe directly in usability studies at the open beta release.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: That is correct, new JavaScript tries to embrace functional paradigms as much as possible and I also think it makes the code easier to read given the number of anonymous functions and other functional elements in the language.

It is certainly the recommeded way to iterate over objects in modern JS (ES6).

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: In the end I am using some functions for sets, but since they are not particularly about cellSets and more about generic sets I will create a different file.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx]: I am going to refactor the reused code, but about the initial default there is actually a bit of a problem:

The reason I am setting the original default as `getInitialTrackData` is that before the way this menu was filled before was also triggering the first update to the heatmap here:
```

  useEffect(() => {
    // Prevent initial dispatch when object appears
    if (isInitialRenderRef.current) {
      return;
    }

    setTrackData(getUpdatedTrackData());
  }, [cellSets.hierarchy]);

```

Since now I am preventing this initial `setTrackData` that causes heatmapplot to render again, the menu is not being filled, so I need to have it filled by default.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: actually in the end I have made a big rewrite to this (marcell also made some comments about this in https://github.com/biomage-ltd/ui/pull/78#discussion_r563800792), so a lot of this is changed and it is more based on the original method of sampling.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: about the original issue, in the end I am leaving both hover and click actions as triggers, that way it is easy to close but also easy to open, let me know if it still feels weird and any other changes should be done.

Comment (vickymorrison) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: Has the staging updated? It doesn't seem to click to close the dropdown menu.

Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:161]: For some reason it is not updating when I do new commits, I just created a new staging environment and set it in the description of the PR, once it is up it should have the latest changes.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
```diff
@@ -16,6 +16,7 @@ import { loadCellSets } from '../../../../../../redux/actions/cellSets';
 import { loadComponentConfig } from '../../../../../../redux/actions/componentConfig';
 
 import { union } from '../../../../../../utils/cellSetOperations';
+import SetOperations from '../../../../../../utils/setOperations';
 
 const COMPONENT_TYPE = 'InteractiveHeatmap';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: I suggest that we directly use `COMPONENT_TYPE` instead of hardcoding `interactiveHeatmap` in all dispatch actions, unless there is some specific reason why we are doing it this way that I am not seeing. The code will be more extensible in the future and cleaner this way.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -1,50 +1,24 @@
...
-    }
-
-    dispatch(
-      updatePlotConfig('interactiveHeatmap', {
-        selectedTracks: trackData.filter((o) => o.selected).map((o) => o.key),
-      }),
-    );
-  }, [trackData]);
+  const selectedTracks = useSelector(
+    (state) => state.componentConfig.interactiveHeatmap.config.selectedTracks,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapMetadataTrackSettings.jsx]: We plan to use the `HeatmapMetadataTrackSettings` component in other parts of the platform, for instance in plots and tables heatmap in the near future. To make it easy to reuse, I would suggest that you don't hardcode `interactiveHeatmap` in here, but pass the component type as props and then use it for the dispatch actions. In this way, `HeatmapMetadataTrackSettings` and `HeatmapPlot` are no longer coupled.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -29,24 +34,105 @@ const HeatmapGroupBySettings = () => {
...
+
+      return;
+    }
+
+    if (cellSetsOrder.length === 0) {
+      return;
+    }
+
+    dispatch(
+      updatePlotConfig('interactiveHeatmap', {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx]: similarly to my comment in the `HeatmapMetadataTrackSettings` component, I suggest that you don't hardcode this here, but rather pass it as props, so that this component is easy to reuse in other parts of the platform.

src/components/ReorderableList.jsx
```diff
@@ -0,0 +1,103 @@
...
+      disabled={currentPosition === reorderableList.length - 1}
+      icon={<DownOutlined />}
+      style={{ marginRight: '5px' }}
+      onClick={() => {
+        onMoveDown(moveDown(reorderableList, key));
+      }}
+    />
+  );
+
+  const fullItem = (listItem, i) => (
```
ivababukova: I think we can do better in terms of descriptive naming in here. 
In terms of `leftItem` and `rightItem`: I took a look at the code, `rightItem` seems to be the actual item name, so I would just call it `itemName` instead of `rightItem`. `leftItem` is the switch that appears in metadata settings, but not in group by settings. It is hard to name that one, but I would maybe call it `itemProps`? This way the name signifies that it can be not only a `switch` but something else, and that whatever it is, belongs to the item. Happy to accept other naming suggestions. You can use the new names for `leftItem` and `rightItem` everywhere in the code, not only for this function.
In terms of the function `fullItem`, since its functionality is to bring the whole item together with all its required buttons, name, etc, I would probably call it `composeItem`.
Maybe renaming `listItem` to just `item` would be cleaner too?

src/components/ReorderableList.jsx
```diff
@@ -0,0 +1,103 @@
...
+      {upButton(listItem.key, i)}
+      {downButton(listItem.key, i)}
+
+      {rightItem(listItem, i)}
+    </div>
+  );
+
+  return (
+    <Space direction='vertical'>
+      {reorderableList.map((listItem, i) => (
```
ivababukova: I would just call this `item`

src/utils/setOperations.js
```diff
@@ -0,0 +1,11 @@
+class SetOperations {
```
ivababukova, line 1: that can fit nicely inside the already existing `cellSetOperations.js` file in that folder, could you move the function there?

ivababukova: nevermind, I just realised that this is for sets, not for cellSets

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
```diff
@@ -100,19 +104,29 @@ describe('HeatmapGroupBySettings', () => {
...
-    const buttons = component.find('Radio');
-    expect(buttons.length).toEqual(2);
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
+
+    // with one ite
+    expect(groupByItems.length).toEqual(1);
+
+    // the item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: We should use css selectors to find the correct element, instead of doing `groupByItems.at(0).childAt(0).childAt(0)`. In this way, if we change the positioning or anything in the `groupByItems`, this test is going to fail. In here you can use `find` and specify the actual component you are looking for, then get its text. Take a look how it is done in other tests, there are lots of other examples like `FocusButton.test.jsx` or `DiffExprResults.test.jsx` or `HideButton.test.jsx`

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
```diff
@@ -100,19 +104,29 @@ describe('HeatmapGroupBySettings', () => {
...
+    const subMenuItems = submenu.find('MenuItem');
 
-    // The selected one should be `sample`.
-    expect(component.find(Radio.Group).props().value).toEqual('sample');
+    // With two items.
+    expect(subMenuItems.length).toEqual(2);
+
+    // each of which should have the right names.
+    expect(subMenuItems.at(0).childAt(0).text()).toEqual('louvain clusters');
+    expect(subMenuItems.at(1).childAt(0).text()).toEqual('Sample');
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: similarly in here -- use css selectors

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
```diff
@@ -126,14 +140,102 @@ describe('HeatmapGroupBySettings', () => {
...
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
+
+    // there should be 2 group by items
+    expect(groupByItems.length).toEqual(2);
+
+    // the first item should be samples
+    expect(groupByItems.at(0).childAt(0).childAt(0).text()).toEqual('Sample');
+    // the second item should be louvain
+    expect(groupByItems.at(1).childAt(0).childAt(0).text()).toEqual('louvain clusters');
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: and here

src/components/ReorderableList.jsx
```diff
@@ -0,0 +1,103 @@
+import React from 'react';
```
ivababukova: could you add unit tests for `ReorderableList` component too?

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: In order to use this and keep the naming consistent I changed COMPONENT_TYPE from 'InteractiveHeatmap' to 'interactiveHeatmap', as far as I could see it is ok, but if you believe this could cause any issues let me know.

src/components/ReorderableList.jsx
cosa65: I changed `listItem` to `itemData` and `fullItem` to `composeItem`. 

About `leftItem` and `rightItem`, they can be any kind of component interchangeably, the only difference between them is that one is to the left of the up down buttons and the other one is to the right, that is why I feel that naming fits them better.

This way we can keep it more generic so that whoever is using it can choose whether it is going to have a name or something more complex and in any side they prefer. At the same time I don't feel this genericness really has a downside on its current usage.

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
Comment (cosa65) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: Yes I was already using selectors in other parts of the test, but in this case I thought I had no way to access the text in one of the child divs, but apparently with `.text()` on the parent I can access it already, so I'll just remove the `childAt`s.

Comment (ivababukova) [reply src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: sounds good!

src/components/ReorderableList.jsx
ivababukova: Sounds good Martin, let's keep it in the way you are proposing, what you are saying makes sense.

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapPlot.jsx]: provided that the group by and metadata settings work when used in the heatmap options, everything should be fine.

src/__test__/components/reorderable-list/ReorderableList.test.jsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  it('renders correctly', () => {
+    const items = component.find('Item');
+
+    // renders the three items
+    expect(items.length).toEqual(3);
+
+    console.log('items.at(0)');
+    console.log(items.at(0).debug());
```
ivababukova: no need for this

src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx
```diff
@@ -93,47 +97,148 @@ describe('HeatmapGroupBySettings', () => {
...
     // Should be rendered.
     expect(component.find('HeatmapGroupBySettings').length).toEqual(1);
 
-    // With two buttons.
-    const buttons = component.find('Radio');
-    expect(buttons.length).toEqual(2);
+    const reorderableList = component.find('ReorderableList');
+    const groupByItems = reorderableList.find('Item');
+
+    // with one ite
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.test.jsx]: forgot the `m` ;)

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -29,24 +38,108 @@ const HeatmapGroupBySettings = () => {
...
+                      if (positionInCellSetOrder > -1) {
+                        // If the cell is included in the cellSet, we have to remove it
+                        newCellSetsOrder.splice(positionInCellSetOrder, 1);
+                      } else {
+                        // If the cell is not included in the cellSet, we have to add it
+                        newCellSetsOrder.push(cellSet);
+                      }
+
+                      console.log('newCellSetsOrder');
+                      console.log(newCellSetsOrder);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx]: no need for that

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
```diff
@@ -29,24 +38,108 @@ const HeatmapGroupBySettings = () => {
...
+                </div>
+              </Menu.Item>
+            );
+          })
+      }
+    </Menu>
+  );
+
+  console.log('cellSetsOrder');
+  console.log(cellSetsOrder);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx]: no need for that

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: is that one fixed?

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -55,11 +55,11 @@ const ExplorationViewPage = () => {
     },
     Heatmap: {
       toolbarControls: [
-        <HeatmapSettings />,
+        <HeatmapSettings componentType={COMPONENT_TYPE} />,
         <RemoveButton />,
       ],
       component: (width, height) => (
-        <HeatmapPlot experimentId={experimentId} width={width} height={height} />
+        <HeatmapPlot componentType={COMPONENT_TYPE} experimentId={experimentId} width={width} height={height} />
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: why are you passing `COMPONENT_TYPE` in here? It is not used, `HeatmapPlot` uses `COMPONENT_TYPE` directly

vickymorrison (approved):

src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-exploration/components/heatmap/HeatmapGroupBySettings.jsx:128]: yes, in the end we left it with both hover and click as trigger and vicky told me on slack it was ok.

ivababukova (approved):

PR #86, opened 2021-01-27
Refactor the 'EmbeddingPreview' UI component to support multiple views of the embedding

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-470

https://ui-scemtqcwy5xrmxw94mkxw6orsc.scp-staging.biomage.net/

N/A

The wireframes Vicky posted had an error, as inversion is only an option for the continuous embedding, not the categorical. I assumed it was a mistake, so I fixed it accordingly. It is not possible to invert colors easily via a continuous embedding using the schema we have.

# Changes
### Code changes
We used the code base for the existing continuous/categorical embedding schema in Plots and Tables to design the embedding. I reused the components from those components.

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers
- [x] (UX changes) Approved by @vickymorrison

### Just before merging:

### Optional
image


aerlaut (dismissed):
The code looks good to me. I understand your reasoning about categorical data not having an invert option.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -17,487 +16,261 @@ import PlotStyling from '../../filter-cells/components/PlotStyling';
...
-              <img
-                alt=''
-                src={plot1Pic}
-                style={{
-                  height: '100px',
-                  width: '100px',
-                  align: 'center',
-                  padding: '8px',
-                  border: '1px solid #000',
+  if (!config) {
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:149]: I just realized this part of the code is a duplicate of the one above (line 147). Can you please remove this duplicate code?

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -17,487 +16,261 @@ import PlotStyling from '../../filter-cells/components/PlotStyling';
...
-      this.updatePlotWithChanges({ plotTitle: 'cell doublet score' });
+  };
+
+  const updatePlotWithChanges = (obj) => {
+    const newConfig = _.cloneDeep(config);
+    _.merge(newConfig, obj);
+    setConfig(newConfig);
+  };
+
+  const renderPlot = () => {
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:132]: renderPlot is actually just a helper function to step through the ifs to render the plot. Would it be better to just use a set of ternary operators e.g. `err ? ( <PlatformError /> ) : !config : ( <Spin /> ) : <Vega >` in place of `renderPlot`?

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -17,487 +16,261 @@ import PlotStyling from '../../filter-cells/components/PlotStyling';
...
+import LegendEditor from '../../../plots-and-tables/components/LegendEditor';
+import LabelsDesign from '../../../plots-and-tables/components/LabelsDesign';
+
+const { Panel } = Collapse;
+
+// TODO: when we want to enable users to create their custom plots,
+// we will need to change this to proper Uuid
+// const plotUuid = 'embeddingCategoricalMain';
+// const plotType = 'embeddingCategorical';
+// const embeddingType = 'umap';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: is the plotUuid used now `embeddingCategorical` and `embeddingContinuous`, as per `initialConfig` fields set in each entry in `plots` below?

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -17,487 +16,261 @@ import PlotStyling from '../../filter-cells/components/PlotStyling';
...
-  updatePlotWithChanges(obj) {
-    this.setState((prevState) => {
-      const newState = _.cloneDeep(prevState);
-
-      _.merge(newState.config, obj);
-
-      return newState;
+    const spec = plots[selectedSpec].specGenerator(config);
+    generateData(spec);
+    console.log(config, spec);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: no need for console log

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -17,487 +16,261 @@ import PlotStyling from '../../filter-cells/components/PlotStyling';
...
+        s.values = [];
+
+        UMAP.forEach((cell, i) => {
+          s.values[cell.cluster_id] = {
+            name: `${cell.cluster_id}`,
+            cellSetId: cell.cluster_id,
+            cellIds: s.values[cell.cluster_id]?.cellIds ? [...s.values[cell.cluster_id].cellIds, i] : [i],
+            color: colorProvider.getColor(),
+          };
+          // cluster_id
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: this comment does not look very useful, you can remove it

src/redux/reducers/componentConfig/initialState.js
```diff
@@ -1,3 +1,46 @@
+const embeddingPreviewInitialConfig = {
```
ivababukova: where is this spec used? Is it only for plots and tables?

vickymorrison:
I don't understand the comment about colour inversion - it is an option in the categorical embedding in Plots and Tables.

ivababukova:
This is a super weird thing - when I open the link in staging and go to configure embedding, when I click between the 4 views, the colours of the cell sets change at some point. See the video:

https://user-images.githubusercontent.com/65414651/106112587-bc08c700-614d-11eb-8303-c400608436c8.mov

Just to confirm -- this is happening because we haven't connected the view with real cell sets data?

marcellp:
> This is a super weird thing - when I open the link in staging and go to configure embedding, when I click between the 4 views, the colours of the cell sets change at some point. See the video:
> 
>  Screen.Recording.2021-01-28.at.09.46.28.mov 
> Just to confirm -- this is happening because we haven't connected the view with real cell sets data?

Correct. This happens because the system right now generates new colors each time the config changes. This will not be the case once the system fetches data from Redux, for now it was the easiest thing to do.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: Fixed.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: Fixed.

src/redux/reducers/componentConfig/initialState.js
marcellp: It was not used for anything, fixed.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:132]: I would keep it as is. Chaining ternary operators is not exactly easy to read, and I think for the purposes of this logic it's probably easier to keep it as a helper.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:149]: Good point. I removed it from the other place, because the config now is used for other things other than rendering the plot, so halting the whole page load until config is ready is a better strategy in my opinion.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx]: It isn't, that comment is not useful right now. We will have to do something with the UUIDs and the type of plots when we persist the configs to Redux, but for now it's not relevant to the code at all.

marcellp:
> I don't understand the comment about colour inversion - it is an option in the categorical embedding in Plots and Tables.

Fair point. I think there is a bug in the way I generate colors right now that somehow messes with the inversion, as it also changes the colors of the clusters and not just the axes. The same issue does not appear in the categorical plot in plots and tables. This will be resolved once we get data from Redux like with everything else.

Either way, the feature is added.

ivababukova (approved):

PR #87, opened 2021-01-27
Fix root cause of banding problem

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-478

TBD

N/A

N/A

# Changes
### Code changes
Made the downsampling less aggressive and actually fixed the root cause of our downsampling problem.

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
image

### DIFF
--- a/src/utils/heatmapSpec.js
+++ b/src/utils/heatmapSpec.js
@@ -183,6 +183,7 @@ const spec = {
         },
         update: {
           fill: { field: 'color' },
+          stroke: { field: 'color' },
         },
       },
     },
@@ -217,6 +218,10 @@ const spec = {
         },
         update: {
           fill: {
+            scale: 'color',
+            field: 'expression',
+          },
+          stroke: {
             scale: 'color',
             field: 'expression',
           },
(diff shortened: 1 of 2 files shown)


src/utils/heatmapSpec.js
```diff
@@ -220,6 +221,10 @@ const spec = {
             scale: 'color',
             field: 'expression',
           },
+          stroke: {
+            scale: 'color',
+            field: 'expression',
+          },
```
marcellp, line 227: I had a shower thought which made me realized that this was causing the banding all along. I was right about the Moire pattern and it being related to space, I just had the wrong idea about what causes it.

The scale for where we position our rectangles on the heatmap is determined by taking the width of the entire canvas and subdividing it into as many equal bands as there are cells. Most of the time this results in a "band" being a non-integer number of pixels (say, 1.034011 pixels/rectangle). As the position of the nth rectangle will be `n*1.034011`, at some point there is guaranteed to be some loss of floating-point precision, at which point the rectangle gets aligned to the nearest whole pixel, causing a tiny gap to appear. The cycle then repeats. As this is a periodic function, it happens at regular intervals, causing visible banding.

This is fixed by simply adding a border of 1px size and filling it with the same color as the inside. The rectangles are drawn on top of each other, so if there is overlap the banding disappears.

ivababukova: that's a great insight, and a very neat fix. Thanks for explaining and this sounds like a very productive shower!

ivababukova (approved):

PR #95, opened 2021-02-03
Feature/redux style config

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-463

#### Link to staging deployment URL 

https://ui-5y8c6t8rpqyi9xwznk1euadomo.scp-staging.biomage.net/

#### Links to any Pull Requests related to this

N/A

#### Anything else the reviewers should know about the changes here

- Modify plot components to use the new Redux schema
- Refactor common variables used by plot styling components out into base styles, which are then destructured in plot's `initialConfig`s.
- Wire up configureEmbedding to use data from Redux store.

# Changes
### Code changes

- Moved diff object to update data styling object from inside plot styling components to props.
- Where applicable, renamed plot styling variables to resemble its Vega variable.
- Created 4 new configs for each of the 4 plots in `embeddingPreview` (suffixed with `embeddingPreview` in `componentConfing/initialState`).

# Definition of DONE
Your changes will be ready for merging after each of the steps below have been completed:

### Testing
- [X] Unit tests written
- [X] Tested locally with Inframock
- [X] Deployed to staging

To set up easy local testing with inframock, follow the instructions here: https://github.com/biomage-ltd/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/biomage-ltd/biomage-utils

### Documentation updates
Is all relevant documentation updated to reflect the proposed changes in this PR?

- [ ] Relevant Github READMEs updated
- [ ] Relevant wiki pages created/updated

### Approvers
- [x] Approved by a member of the core engineering team
- [ ] (UX changes) Approved by vickymorrison (this is her username, tag her if you need approval)

### Just before merging:
- [x] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional
- [ ] Photo of a cute animal attached to this PR

PR #97, opened 2021-02-03
Implement the embedding cell set and metadata/sample views

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-451

https://ui-embedding-load-real-data.scp-staging.biomage.net/

N/A

Changed some additional things, such as the layout of the data processing page (it's a card so it looks more put together) as well as using a carousel for the filters. I also refactored some of the loading logic from the config component to the main (plot) component.

# Changes
### Code changes
The plots now get their data from Redux, as required.

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers
- [ ] (UX changes) Approved by vickymorrison

### Just before merging:

### Optional


marcellp:
Staging now available at https://ui-embedding-load-real-data.scp-staging.biomage.net/.

package.json
```diff
@@ -18,7 +18,7 @@
     "start": "NODE_ENV=development node server.js",
     "build": "next build",
     "analyse": "NODE_ENV=analyse next build",
-    "prod": "NODE_ENV=production node server.js",
+    "prod": "NODE_ENV=development node server.js",
```
ivababukova: why?

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
```diff
@@ -39,20 +42,25 @@ const CalculationConfig = (props) => {
   }, 1500), []);
 
   const updateSettings = (diff) => {
-    // Always update the settings in the Redux store.
-    dispatch(updateProcessingSettings(
-      experimentId,
-      FILTER_UUID,
-      diff,
-    ));
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:47]: it looks like this is still done in both cases, why is it refactored and put under each of the if statements?

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
```diff
@@ -63,14 +71,6 @@ const CalculationConfig = (props) => {
     dispatch(loadEmbedding(experimentId, embeddingMethod));
   };
 
-  useEffect(() => {
-    if (!experimentId) {
-      return;
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:68]: why is this removed?

package.json
marcellp: There is a bug in staging that I tried to fix, I figured out what the problem is and it's done now.

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -64,7 +64,7 @@ const DataProcessingPage = () => {
     {
       id: 'comptueEmbeddingFilter',
       name: 'Compute embedding',
-      render: () => <EmbeddingPreview experimentId={experimentId} />,
+      render: (expId) => <EmbeddingPreview experimentId={expId} />,
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx:67]: better call it `experimentId` instead of `expId`, so that it is consistent with the naming that we use everywhere else

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:68]: It's not removed, it's just refactored into EmbeddingPreview, see https://github.com/biomage-ltd/ui/pull/97/files#diff-afd1e0b0de5b6da0610c8bd7fbb9865fec7ffc38aaf347805381ff303ea967bbR81-R86.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:47]: Correct. The reason is that in the former case, the dispatch is executed as a callback, notice that it is embedded as the second argument of `setChangesOutstanding`.

When you use `useState()`, there is no guarantee that the setter will be called synchronously with the rest of the code, i.e. it is possible for `changesOutstanding` to be `false` by the time the update Redux action is called. This caused some timing issues, e.g. it would think that the user pressed `Apply` and therefore the embedding could be reloaded when that was not the case.

The way it's resolved is to put the dispatch into a callback, which is provided as the second argument, e.g.:

```
setValue(false, () => console.log('ran after value set to false'))
```

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:67]: It cannot be called that, because `experimentId` is shadowed as a prop.

The reason for this way of rendering is that on initial load (when the variable would be created), `experimentId` would be `undefined`. This causes staging to break (as it does right now) because the data would try to be loaded initially for an undefined experiment. The way this can be resolved is to pass the `experimentId` just when the component is about to render.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -81,23 +78,51 @@ const EmbeddingPreview = () => {
...
   };
 
+  // Start loading by fetching the settings and cell sets.
+  useEffect(() => {
+    dispatch(loadCellSets(experimentId));
+    dispatch(loadProcessingSettings(experimentId, FILTER_UUID));
+  }, []);
+
+  // Once the config is loaded, we can also load the embedding.
+  useEffect(() => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:88]: we might need an if statement in here to check if `embeddingMethod` is not defined, because if it is not, we do not want to send that dispatch request

Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:88]: You're right, this is now fixed.

marcellp:
I understand staging is currently broken. I have some idea as to why, I'm currently pushing some fixes to make sure it works.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -39,20 +34,22 @@ import LabelsDesign from '../../../plots-and-tables/components/LabelsDesign';
 
 const { Panel } = Collapse;
 
-const EmbeddingPreview = () => {
+const EmbeddingPreview = (props) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:37]: it will be worth looking at Agi's PR and chatting with him - he has done lots of changes to this component in his PR too.

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:67]: I see, yeah you are right. Good call regarding the error

ivababukova:
It is showing an empty plot on the samples view -- it will be more user-friendly if there was an indication about why the plot is empty - i.e. "The plot is empty because your dataset consists of only 1 sample"
Screenshot 2021-02-04 at 13 53 15

marcellp:
Yup, this used to fail on staging because of a bug, and now that bug is fixed. I will add some empty state to this one accordingly.

marcellp:
@ivababukova -- this is now fixed.

ivababukova:
Thanks Marcell, it looks better now, although I noticed a weird issue with the rendering of the cell sets view. I made a video that illustrates it, but basically for about half a second the embedding renders as empty. There were similar issues appearing in the frequency plot, I wonder if it is a vega thing. Do you have an idea what is the reason for this?
https://user-images.githubusercontent.com/65414651/106900562-f4347a80-66f6-11eb-8840-d43a6e294cc3.mov

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:47]: thanks for the clarification, makes sense now

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -1,35 +1,30 @@
...
   InfoCircleOutlined,
 } from '@ant-design/icons';
 import _ from 'lodash';
 import { Vega } from 'react-vega';
-import useSelection from 'antd/lib/table/hooks/useSelection';
 import plot1Pic from '../../../../../../../static/media/plot9.png';
 import plot2Pic from '../../../../../../../static/media/plot10.png';
-import PlotStyling from '../../filter-cells/components/PlotStyling';
 import CalculationConfig from './CalculationConfig';
 import UMAP from './new_data.json';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:15]: you can remove this file import and also delete the file itself if it is no longer used everywhere

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -81,23 +78,55 @@ const EmbeddingPreview = () => {
...
-  // If the user toggles to a different embedding, set the config to be the initial
-  // state for that type of plot.
   useEffect(() => {
-    setConfig(plots[selectedSpec].initialConfig);
+    // If the user toggles to a different embedding, set the config to be the initial
+    // state for that type of plot.
+    setConfig(_.cloneDeep(plots[selectedSpec].initialConfig));
   }, [selectedSpec]);
 
   // Quick and dirty function to massage prepared data into a good shape.
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:132]: @aerlaut you are deleting this function as part of your PR. Is this change in here going to clash very badly with your changes?

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:132]: Yes. with the refactor, `embeddingPreview` becomes very trimmed down. The latest changes can be seen here : https://github.com/biomage-ltd/ui/pull/95/files#diff-afd1e0b0de5b6da0610c8bd7fbb9865fec7ffc38aaf347805381ff303ea967bb

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:37]: Latest changes pushed to #95. I don't think there'll be additional major changes, just unit tests left

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:132]: Correct, although I can see that you basically implemented a large chunk of the ticket that I was assigned as well -- this makes sense given your scope I guess, and I can see that we can overwrite most of my changes with this of yours. That said, I think we should merge yours after mine to make sure that nothing that I verified as working as part of my ticket was somehow left out by yours.

marcellp:
> Thanks Marcell, it looks better now, although I noticed a weird issue with the rendering of the cell sets view. I made a video that illustrates it, but basically for about half a second the embedding renders as empty. There were similar issues appearing in the frequency plot, I wonder if it is a vega thing. Do you have an idea what is the reason for this?
> https://user-images.githubusercontent.com/65414651/106900562-f4347a80-66f6-11eb-8840-d43a6e294cc3.mov

It is incrementally loading. First it loads just the spec with no data, then as the data becomes available, it loads it with the data. If you want, I can make it not do that, although given Agi's changes it may be superfluous.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx
```diff
@@ -136,23 +186,40 @@ const EmbeddingPreview = () => {
...
       );
     }
 
+    if (selectedSpec === 'sample'
+      && plotSpec.data.find((d) => d.name === 'cellSets').values.length === 0) {
+      return (
+        <Empty description='Your project has only one sample.' />
+
+      );
+    }
```
Comment (marcellp) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/EmbeddingPreview.jsx:221]: @aerlaut -- this change should be incorporated into your MR to ensure the user sees a default state when they have only one sample.

ivababukova (approved):

PR #96, opened 2021-02-03
Add navigator between plots with preview

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-515?atlOrigin=eyJpIjoiZTVjYWU0NWJlOTgwNGJkM2JiNzU4ZWM5ZDNiMzJhZGMiLCJwIjoiaiJ9

#### Link to staging deployment URL 
https://ui-3i4ir19b7bfj9isbi1xi7dpn1a.scp-staging.biomage.net/

#### Anything else the reviewers should know about the changes here

# Changes
### Code changes

# Definition of DONE
Your changes will be ready for merging after each of the steps below have been completed:

### Testing
- [x] Unit tests written
- [x] Tested locally with Inframock
- [x] Deployed to staging

To set up easy local testing with inframock, follow the instructions here: https://github.com/biomage-ltd/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/biomage-ltd/biomage-utils

### Documentation updates
Is all relevant documentation updated to reflect the proposed changes in this PR?

- [ ] Relevant Github READMEs updated
- [ ] Relevant wiki pages created/updated

### Approvers
- [ ] Approved by a member of the core engineering team
- [ ] (UX changes) Approved by vickymorrison (this is her username, tag her if you need approval)

### Just before merging:
- [ ] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional
- [ ] Photo of a cute animal attached to this PR

PR #99, opened 2021-02-04
Create data integration UI mockups

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-525

https://ui-8dgcnnqrgrm7c5bi4yxzph3ymo.scp-staging.biomage.net/

N/A

No functionality has been built into the page

# Changes
### Code changes

Added visual components according to mockup.

# Definition of DONE

### Testing
- [X] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -30,36 +48,153 @@ const defaultStylingConfig = {
...
+                </Panel>
+                <Panel header='Axes and Margins' key='axes'>
+                  <AxesDesign config={config} onUpdate={updatePlotWithChanges} />
+                </Panel>
+                <Panel header='Colours' key='colors'>
+                  <ColourInversion
+                    config={config}
+                    onUpdate={updatePlotWithChanges}
+                  />
+                  {(selectedPlot === 'sample' || selectedPlot === 'frequency') ? (
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: on line 171 you have the same conditions, maybe merge those ternary operators into a single one?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -1,43 +1,175 @@
...
+                ))
+              }
+            </Select>
+          </Form.Item>
+          <Form.Item label='# of genes:'>
+            <InputNumber
+              defaultValue={2000}
+              min={1}
+              onPressEnter={() => { }}
+              onChange={() => { }}
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Do we need to have these events there, that do nothing before functionality is added?

StefanBabukov (dismissed):

src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx
```diff
@@ -0,0 +1,107 @@
...
+    expect(form.length).toBeGreaterThan(0);
+  });
+
+  it('renders correctly when the data is in the store', () => {
+    const store = mockStore(storeState);
+
+    const component = mount(
+      <Provider store={store}>
+        <CalculationConfig
+          experimentId='1234'
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx]: Why are experimentId, width and length set as props? There is probably a good reason that I don't know about because I have seen them in many other tests too.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -30,36 +48,153 @@ const defaultStylingConfig = {
...
 };
 
+const { Panel } = Collapse;
+
 const DataIntegration = () => {
-  const [plotConfig, setPlotConfig] = useState(defaultStylingConfig);
+  const router = useRouter();
+  const { experimentId } = router.query;
+
+  const [selectedPlot, setSelectedPlot] = useState('sample');
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:63]: The functionality for the different plots and switching between each of them is actually part of a different ticket (in [this PR](https://github.com/biomage-ltd/ui/pull/96)), the addition of something like this here too might create some unnecessary merge conflicts.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -30,36 +48,153 @@ const defaultStylingConfig = {
...
         </Col>
+
         <Col span={7}>
-          <CalculationConfig />
-          <PlotStyling
-            config={plotConfig}
-            onUpdate={updatePlotWithChanges}
-            singlePlot
-          />
+          <Collapse defaultActiveKey={['data-integration']}>
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:190]: Could you move the styling elements over to a different component? Something that can replace `PlotStyling` perhaps

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Those are templates for future work, so that the person doing the next ticket to implement the functionality can continue from there. Maybe I should put `// TODO` or something to signify that.

src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx]: I believe most of them are just copy pasted from other test codes 😆 They aren't read by the components because they aren't used. I'll take them out.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:63]: I see I see. I was using it to test for the if statements to check if conditionals for the different parts of the styling (if they show up properly or not). I think that justifies the code to be there.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:190]: Yea, this was something I talked about with @ivababukova  for #95 . This will be investigated and looked into.

There are a lot of redundant duplication of codes. However, as the PR shows, refactoring something that affects a large part of the codebase can be complicated, especially with multiple merge requests being done at the same time (merge conflicts).

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Maybe @ivababukova or @marcellp has something to say about this practice? Is it good to leave templates / todo behind for future works? Or just keep the codebase clean for the PR?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: The conditions are located under different `<Panel />` scope. If I were to group them, there would be some chunks of duplicated code :

```
{(selectedPlot === 'sample' || selectedPlot === 'frequency') ? (
  <Panel header='Colours' key='colors'>
   <ColourInversion>
      <Alert ... />
   </ColourInversion>
  </Panel>
<Panel header='Legend' key='legend'>
                    <LegendEditor ... />
                  </Panel>
) : (
<Panel header='Colours' key='colors'>
   <ColourInversion />
  </Panel>
)
```

By writing this way, I try to minimize code duplication, but as you can see, it comes with more if statements. Some of this is a matter of personal preference, but I think each has its own merits. Maybe @ivababukova and @marcellp can also weigh in on this?

src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx
Comment (ivababukova) [reply src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx]: The experiment id is needed for all components that dispatch redux actions. The reason why they are needed is because when dispatching a redux action, the ui usually sends a request to the backend, and for that request we need to have the experiment id.
In data-epxloration components we have to have width and height as props because of the tiled component (called react mosaic) that we use to render the data exploration page into tiles. If you look at the code in here: https://github.com/biomage-ltd/ui/blob/master/src/pages/experiments/%5BexperimentId%5D/data-exploration/index.jsx#L54 the mosaic framework passes width and height to each of the components in the `TILE_MAP` object, which are calculated from the mosaic framework based on the screen width and height and the percentage split of how much proportion of the screen each component should take. 
If you are more interested, you can take a look in here too: https://github.com/biomage-ltd/ui/blob/master/src/redux/actions/layout.js

Comment (ivababukova) [reply src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx]: so in short, width and height are not needed for components outside of data exploration. Experiment id is needed for every component that has to send a request to the backend at some point.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: It depends. It is usually best to leave the code clean and without todos, commented out code is not a best practice either. If the todo is big enough, make a ticket. If the todo is small enough, implement it. The reason is that even if you put a todo in the code, it is very likely that it won't get done, but just stay there, which is not great. 

In this case however, since you are saying that this will get implemented in the next ticket, there is little chance that the code will be forgotten and left like this. In this case either way is fine -- having the empty functions or not.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:190]: yup, we have a ticket to clean these things up for next sprint

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: Since there are lots of if statements here the code feels quite messy either way. The way I would probably do it is to first define the three different views, one for sample, one for elbow and one for frequency. These can be put in the `plots` object, so something like this:

```
const plots = {
    sample: {
      title: 'Samples',
      imgSrc: cellBySamplePic,
      stylingSpecifics: (config) => (<Panel header='Colours' key='colors'>
   <ColourInversion>
      <Alert ... />
   </ColourInversion>
  </Panel>
<Panel header='Legend' key='legend'>
                    <LegendEditor ... />
                  </Panel>)
    },
```
and then in here, I would just plug in the styling specifics for the relevant plot

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Ok... I think it's better to erase the hanging functions.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: Factored out into plot-specific option under `plots`

src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/data-processing/data-integration/CalculationConfig.test.jsx]: Deleted in new commit

cosa65:
A little bit nitpicky so feel free to let me know if you don't agree with this, but there seems to be quite a bit of vertical margin between each of the (Data Integration) settings components compared to the mockups

aerlaut:
> A little bit nitpicky so feel free to let me know if you don't agree with this, but there seems to be quite a bit of vertical margin between each of the (Data Integration) settings components compared to the mockups

Yea, there is a difference. This space is due to each input being wrapped in their own `<Form.Item />` . The size is different from the one in the mockup, but consistent with other inputs in other forms, So I think it's better to not change it. 

If you look at the [docs](https://ant.design/components/form/#Form.Item), by wrapping components in `<Form.Item />`, their behaviour changes. However, I don't think we're using this feature,

cosa65 (approved):

vickymorrison:
A few things to change please:
1. the '# of genes' is duplicated - it should only appear directly under the data integration method dropdown.
2. the '% variation explained' should not be able to be altered by the user - this will be provided as information to the user based on the selection in the 'Number of Principal Components' box.
3. the arrows in the '# of genes' field should move up/down in increments of 100 rather than 1.
4. a question - are the plot styling menu options part of another ticket?

aerlaut:
> A few things to change please:
> 
>     1. the '# of genes' is duplicated - it should only appear directly under the data integration method dropdown.
> 
>     2. the '% variation explained' should not be able to be altered by the user - this will be provided as information to the user based on the selection in the 'Number of Principal Components' box.
> 
>     3. the arrows in the '# of genes' field should move up/down in increments of 100 rather than 1.
> 
>     4. a question - are the plot styling menu options part of another ticket?

1. Fixed 
2. Fixed  -> made into readOnly, instead of disabled.
3. Fixed
4. Yes, they're being refactored in another ticket

vickymorrison:
The staging doesn't seem to be updated so I can't check the changes.

vickymorrison (approved):

PR #102, opened 2021-02-09
resolution and minDistance fires event only after change

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-542
https://biomage.atlassian.net/browse/BIOMAGE-545

https://ui-kocjfd3mq4p54bjrfz1iggw66h.scp-staging.biomage.net/

N.A.

- `resolution` only fires action when finished changing the value
-  `minDistance` changes value only after step (clicking the arrow) or `onBlur` when the user leaves focus of the input
- Prevented config to run when user presses enter when inputting `minDistance`

# Changes
### Code changes

- Created local states to handle 'during change' transitions
- Moved dispatch out of `setChangesOutstanding` so that it fires

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
+++ b/src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import _ from 'lodash';
 import { useDispatch, useSelector } from 'react-redux';
 import {
@@ -37,21 +37,36 @@ const CalculationConfig = (props) => {
   const { umap: umapSettings, tsne: tsneSettings } = data?.embeddingSettings.methodSettings || {};
   const { louvain: louvainSettings } = data?.clusteringSettings.methodSettings || {};
 
-  const dispatchDebounce = useCallback(_.debounce((f) => {
+  const [resolution, setResolution] = useState(null);
+  const [minDistance, setMinDistance] = useState(null);
+
+  useEffect(() => {
+    if (!resolution && louvainSettings) {
+      setResolution(louvainSettings.resolution);
+    }
+  }, [louvainSettings]);
+
+  useEffect(() => {
+    if (!minDistance && umapSettings) {
+      setMinDistance(umapSettings.minimumDistance);
+    }
+  }, [umapSettings]);
+
+  const dispatchDebounce = _.debounce((f) => {
     dispatch(f);
-  }, 1500), []);
+  }, 1500);
 
   const updateSettings = (diff) => {
     if (diff.embeddingSettings) {
       // If this is an embedding change, indicate to user that their changes are not
       // applied until they hit Apply.
-      setChangesOutstanding(true, () => {
-        dispatch(updateProcessingSettings(
-          experimentId,
-          FILTER_UUID,
-          diff,
-        ));
-      });
+
+      setChangesOutstanding(true);
+      dispatch(updateProcessingSettings(
+        experimentId,
+        FILTER_UUID,
+        diff,
+      ));
     } else {
       // If it's a clustering change, debounce the save process at 1.5s.
       dispatchDebounce(saveProcessingSettings(experimentId, FILTER_UUID));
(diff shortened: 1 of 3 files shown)


src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
```diff
@@ -86,21 +101,25 @@ const CalculationConfig = (props) => {
...
-            },
-          )}
+          onChange={(value) => setMinDistance(value)}
           onStep={(value) => updateSettings(
             {
               embeddingSettings:
                 { methodSettings: { umap: { minimumDistance: parseFloat(value) } } },
             },
           )}
+          onPressEnter={(e) => e.preventDefault()}
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:114]: So the plot does not reload when user press enter when entering the number.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
```diff
@@ -45,13 +60,13 @@ const CalculationConfig = (props) => {
...
-          diff,
-        ));
-      });
+
+      setChangesOutstanding(true);
+      dispatch(updateProcessingSettings(
+        experimentId,
+        FILTER_UUID,
+        diff,
+      ));
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:69]: Dispatch is moved out so that it'll be called. `setState` does not have a callback,

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
```diff
@@ -37,6 +37,21 @@ const CalculationConfig = (props) => {
   const { umap: umapSettings, tsne: tsneSettings } = data?.embeddingSettings.methodSettings || {};
   const { louvain: louvainSettings } = data?.clusteringSettings.methodSettings || {};
 
+  const [resolution, setResolution] = useState(null);
+  const [minDistance, setMinDistance] = useState(null);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:41]: Introduced local states as "buffer" before firing dispatch to redux when state is ultimately set.

src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx
Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:69]: You're right, this is a bug. Keep in mind, though: there is no guarantee that `changesOutstanding` will be set *before* the dispatch is called, as the setter for `useState` is asyc.

If the timing of the two events (`setChangesOutstanding` & `dispatch`) does not matter, you can keep it as is. Otherwise, they should be linked via `useEffect`.

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:69]: Yes, I'm aware of the racing of the two conditions. 

As they involve components only in the browser (state and Redux), if everything goes well, they should resolve in a couple of milliseconds. That should be enough time for the states to resolve before the user can click Apply. However, as you said, there is no guarantee.

Ideally, `setChangesOutstanding` is fired after dispatch finishes changing. So, I think it'll be better to implement it with `useEffect`?

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:69]: I can fix this by adding a variable under `useRef` to skip the first run and `setChangesOutstanding` using `useEffect`. However, that causes 2 tests to not be able to run, because then the Alert requires Redux store change (which is not yet supported by `mockStore`). The code below basically sums up the changes


```
const enableChangesOutstanding = useRef(false);

  useEffect(() => {
    if (enableChangesOutstanding.current && !changesOutstanding) {
      setChangesOutstanding(true);
    }
  }, [umapSettings, tsneSettings, embeddingMethod]);

const updateSettings = (diff) => { ....
   enableChangesOutstanding.current = true;
...}
```

Personally, I don't think the changes is worth making, as it disables test too. The tests that are disabled are the one that checks if alert pops up when there is a change, and if the alert disappears again after clicking save.

Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/configure-embedding/components/CalculationConfig.jsx:69]: I agree, for the purposes of this it doesn't matter. We can ignore it for now.

marcellp (approved):

PR #105, opened 2021-02-10
Redux store for data integration config

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-529

https://ui-5twj5effbbfp1ija7rt8h51ueh.scp-staging.biomage.net/

N.A.

- Removed alert beside "Run" button and show the one on top, to make it consistent with configure embedding

# Changes
### Code changes

- Used `experimentSettings` mechanism for updating store, so the are no new actions
- Modified `experimentSettings` reducer so that array in object is replaced with new array (instead of merged)
- Refactor method options for data integration into a separate component
- Delete `data-integration/index.jsx` because it is not used anymore. `DataIntegration.jsx` is used directly in `data-processing/index.jsx`, bypassing the `index.jsx`

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
-
<image src="https://user-images.githubusercontent.com/2862362/107448941-3f2a1400-6b75-11eb-9b3d-b1dd5f56b056.png" width="400" />


src/redux/reducers/experimentSettings/updateProcessingSettings.js
```diff
@@ -5,7 +5,19 @@ import initialState from './initialState';
...
+    }
+  };
+
+  const newConfig = _.cloneDeep(
+    _.mergeWith(
+      state.processing[settingName],
+      configChange,
+      arrayMerge,
+    ),
+  );
```
aerlaut, line 20: Change from `_.merge` to `_mergeWith` to handle array in config object. The default behaviour of `_.merge` will merge arrays in object, which means `_.merge({ a : [1]}, {a : [] })` results in `{a : [1]}`, while what is wanted is a replacement of the array : `{a : []}`

src/redux/reducers/experimentSettings/initialState.js
```diff
@@ -4,6 +4,23 @@ export default {
...
+          },
+        },
+      },
+      dimensionalityReduction: {
+        method: 'rpca',
+        numPCs: 30,
+        variationExplained: 91,
+        excludeGeneCategories: [],
+      },
+    },
```
aerlaut, line 23: Created new schema for storing data integration. This is different from the schema in the database, where previously it was only `dimensionalityReduction`. This has to be taken into account when implementing mechanism for storing this in the database.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/SeuratV4Options.jsx
```diff
@@ -0,0 +1,65 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+import { Form, InputNumber, Select } from 'antd';
+
+const { Option } = Select;
+
+const MethodOptions = (props) => {
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/SeuratV4Options.jsx]: New method-specific component to facilitate changes in options when different component method is chosen

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -157,21 +143,23 @@ const CalculationConfig = () => {
...
           <Form.Item label='Number of Principal Components'>
             <InputNumber
-              value={settings.dimensionalityReduction.numPCs}
-              onChange={(value) => updateSettings({ dimensionalityReduction: { numPCs: value } })}
+              value={numPCs}
+              onChange={(value) => setNumPCs(value)}
+              onPressEnter={(e) => e.preventDefault()}
               onStep={(value) => updateSettings({ dimensionalityReduction: { numPCs: value } })}
+              onBlur={(e) => updateSettings({ dimensionalityReduction: { numPCs: parseInt(e.target.value, 0) } })}
             />
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Used a local state (`numPCs`) to buffer changes in the UI, and only fire updates to store when finished changing (`onBlur`, `onStep`). `onPressEnter` is disabled to prevent triggering of `Apply` button when user presses enter in this field.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -56,38 +68,25 @@ const CalculationConfig = () => {
...
-      noGenes: 2000,
-      normalisation: 'logNormalise',
-    },
-    dimensionalityReduction: {
-      numPCs: 30,
-      variationExplained: 91,
-      excludeGeneCategories: [],
-      method: 'rpca',
-    },
-  };
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx:71]: Moved to redux

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from 'react';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:1]: there is a console error for this file, see screenshot:

Screenshot 2021-02-10 at 10 09 52

Can you fix it? We should be looking out for those errors and fixing them during development -- they can potentially break production in unexpected ways, and are also a pain for local development and testing

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -1,4 +1,5 @@
...
+import { useDispatch, useSelector } from 'react-redux';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:2]: another console error when moving from doublet filter step to data integration:

Screenshot 2021-02-10 at 10 13 01

Can you fix it too?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:1]: this one occurs when I am on the data integration step during the first time when I click to the second elbow plot view

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:2]: this `useDispatch` doesn't seem to be used -- if it is not used, can you remove it from the import?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -84,9 +87,17 @@ const DataIntegration = () => {
...
   const { experimentId } = router.query;
 
+  const calculationConfig = useSelector((state) => state.experimentSettings.processing.dataIntegration);
+
   const [activePlotKey, setActivePlotKey] = useState('elbowPlot');
   const [config, setCurrentConfig] = useState(persistedConfigs.elbowPlot);
 
+  useEffect(() => {
+    if (experimentId && !calculationConfig) {
+      loadProcessingSettings(experimentId);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: actually, you should use the imported `dispatch` in here

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
```diff
@@ -84,9 +87,17 @@ const DataIntegration = () => {
   const router = useRouter();
   const { experimentId } = router.query;
 
+  const calculationConfig = useSelector((state) => state.experimentSettings.processing.dataIntegration);
+
   const [activePlotKey, setActivePlotKey] = useState('elbowPlot');
   const [config, setCurrentConfig] = useState(persistedConfigs.elbowPlot);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:94]: These ones are basically having a similar function in terms of state updates and and because of that I think that we can remove the `[config, setCurrentConfig]` state variable. If we keep `[activePlotKey, setActivePlotKey]` (which I recommend to rename so that the name for it matches the name for the state variable that we use for that in the configure embedding for clearness and consistency. In general it is a good idea to aim to have uniform naming for variables that provide similar functionality across modules for easier code readability, refactoring and clearness), then `config` will just be equal to `persistedConfigs[activePlotKey]`.
This will simplify the code, because now you only need to worry about updating `activePlotKey`, the `config` will then be automatically taken care of.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -182,20 +170,37 @@ const CalculationConfig = () => {
...
+            <Row>
+              <Col span={6}>
+                <Button
+                  size='small'
+                  type='primary'
+                  htmlType='submit'
+                  disabled={!changesOutstanding}
+                  onClick={applyDataIntegrationSettings}
+                >
+                  Apply
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: It seems a bit weird how the apply button is disabled initially. I understand why it is disabled initially -- because there are no changes outstanding, however it does take a few brain-cycles to realise that. My concern is that users who will have much less context about this system won't realise why this button is disabled initially and they would think that they can't change the settings. To mitigate this, I would add a tooltip that appears on hover over this button when there are no changes outstanding that states why the button is disabled and that the user needs to change the configuration settings to enable the button. This issue also appears in Configure Embedding, we can add that tooltip there as well. @vickymorrison what do you think? Is there anything better that we can do here?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -97,7 +96,7 @@ const CalculationConfig = () => {
         {changesOutstanding && (
           <Form.Item>
             <Alert
-              message='Your changes are not yet applied. To update the plots, click Apply.'
+              message='Your changes are not yet applied. To re run data integration, click Apply.'
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: usability issue: this message does not trigger after I change the numbers for `# of genes` or the `number of principal components` -- in order for the alert to appear, I actually have to click on the screen itself.
Also, if I change a setting and then unchange it, the message that my settings are not applied dos not appear, which is not accurate. Now I understand that this is a hard one to fix, but if we don't fix it here on the UI side, it would cause unnecessary requests to the backend and unnecessary reruns of the future step functions pipeline, which will be an even bigger pain.

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: I also actually, just fyi, no actions for this specific change, but because of this issue I just realised that we might need to disable the caches for data processing entirely because otherwise users will have discrepancies in their count matrix files if they revert back to previously selected settings (I added a comment here: https://docs.google.com/document/d/1cVqHMt9gqyS3vYnCdhcZrKeokXDAQDlwlCF5lmNYf8w/edit#heading=h.3etmldd107r6).

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:2]: used `useDispatch` in next commit

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx]: Yea, I missed that. Added in next commit.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:1]: Changes to fix this is related to the plot styling config. There will be another ticket for the plot styling config, which should remove this,

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:2]: Fixed in next commit

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (vickymorrison) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Iva's suggestion is a good immediate solution.
It raises the bigger question of whether users will realise that Data Integration has already been performed. I know that Adam's wireframes for the filters have a toggle at the top with 'default' and 'manual' options which make it clear that default settings have already been applied to the data. Iva and I will discuss whether to apply this to Data Integration and Configure Embedding too.
But for now, adding the tooltip is a good way forward.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:94]: @cosa65 can you handle this in your changes? if I fix this in this commit, I'll have to update the plot stylings.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Ok. I've added the tooltip both in Data Integration and Configure Embedding

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: This only happens when you type in the number right?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/DataIntegration.jsx:94]: Actually removing this useState is part of what I am doing in my current ticket, it was meant to be a temporary thing to allow to test basic styling functionality until we started working on redux.

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Fixed in latest commit

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: yup, only when I type in the number

Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: I don't think your comment will be an issue -- added a comment to the google doc @ivababukova

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: thanks Marcell, I will check there and respond there directly

Comment (marcellp) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx]: Can you change `re run` to `rerun`?

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
```diff
@@ -19,11 +20,22 @@ import {
   QuestionCircleOutlined,
 } from '@ant-design/icons';
 
+import SeuratV4Options from './SeuratV4Options';
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx:23]: why don't we rename this file to `MethodOptions` to match the name of the component?

src/redux/reducers/experimentSettings/initialState.js
ivababukova: thanks for raising that, @cosa65 do we have a ticket for updating the API schema about this? If so, we will also need to update this file too: https://github.com/biomage-ltd/api/blob/master/src/api/route-services/mock-data.json

src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx:23]: It should be the other way around, I'll rename the component to `SeuratV4Options`

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx:23]: makes sense!

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/data-integration/components/CalculationConfig.jsx:23]: changed in new commit

src/redux/reducers/experimentSettings/updateProcessingSettings.js
ivababukova: so is `mergeWith` essentially equivalent to replacing the old config values for keys with the new values?

ivababukova (approved):

PR #120, opened 2021-02-18
Implement load and save progress data functionality

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-535?atlOrigin=eyJpIjoiNTBmNzM2N2ZiNmEyNGFjN2JlNGI0ZDhlYjkxMjQ5ZmYiLCJwIjoiaiJ9

https://ui-qpyacjy41t7rt8nghga8367g8h.scp-staging.biomage.net/

Modified existing code so that when opening data processing we automatically open on the latest pending step

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


cosa65:
Added package-lock and package files that were updated when I ran npm install (without doing any direct change to the files)

src/redux/reducers/experimentSettings/initialState.js
```diff
@@ -1,6 +1,6 @@
 export default {
   processing: {
-    meta: {
+    processingConfig: {
```
aerlaut: Can you please return this key back to `meta`? It's purposefully called `meta` to describe metadata about file data processing (e.g. progress, status (loaded, error), etc.)

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -66,17 +66,34 @@ const DataProcessingPage = ({ experimentId, experimentData, route }) => {
...
+    };
+
+    const completeProcessingStepIfAdvanced = () => {
+      if (stepIdx > completedSteps.size) {
+        dispatch(completeProcessingStep(experimentId, steps[stepIdx - 1].key, steps.length));
+      }
+    };
+
+    goToStepIdx();
+    completeProcessingStepIfAdvanced();
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx:102]: Any reason why `goToStepIdx()` and `completeProcessingStepIfAdvanced()` are implemented as functions instead of just blocks of code?

Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:102]: just for readability, and differentiation of responsibilities of each of the blocks of code, we are doing two completely different things that are based on the value that `stepIdx` takes

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -66,17 +66,34 @@ const DataProcessingPage = ({ experimentId, experimentData, route }) => {
...
 
   const dispatch = useDispatch();
-  const [stepIdx, setStepIdx] = useState(0);
 
   const completedPath = '/experiments/[experimentId]/data-exploration';
 
-  const completedSteps = useSelector((state) => state.experimentSettings.processing.meta.stepsDone);
+  const completedSteps = useSelector((state) => state.experimentSettings.processing.processingConfig.stepsDone);
+
+  const [stepIdx, setStepIdx] = useState((completedSteps?.size ?? 0) % steps.length);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx]: Why have to modulo (%)? can `completedSteps.size` be greater than `steps.length`?

Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx]: completedSteps.size can be equal to steps.length, while the highest value stepIdx should be able to take is steps.length - 1

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:102]: Hmm... the body of the functions are quite simple (functions wrapped in if statements). I don't think you need to put them in functions... If you want to explain what they do, I think you can put a line of comment on top of each of the function,

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx]: Oh I see..

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:102]: I think I did it also to show the difference in functionality between both of them, is there any issue with having them as auxiliary functions inside the useEffect they are being used in?

I also would prefer to avoid comments when it is possible to have it expressed in the code

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx]: it was to have the latest step be the one that is opened as soon as we arrive to the page, but there is an issue with this that has been fixed in the newest version by adding an `initialState` value and a `useEffect`.

src/redux/actions/experimentSettings/completeProcessingStep.js
```diff
@@ -1,6 +1,34 @@
...
+
+  await fetch(
+    `${getApiEndpoint()}/v1/experiments/${experimentId}/processingConfig`,
+    {
+      method: 'PUT',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify([{
+        name: 'meta',
+        body,
+      }]),
```
ivababukova, line 28: you should add a try/catch block in here and handle the case when there is an exception (dispatch `EXPERIMENT_SETTINGS_PROCESSING_ERROR` action in that case), so that the UI can update appropriately if saving the completed step did not work out.

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:102]: no, not really a problem

src/redux/actions/experimentSettings/completeProcessingStep.js
cosa65: In the case it didn't work out should I add a reducer for that and an error message here too? In that case I should probably show an error message when it can't load the steps too.

ivababukova: the reducer for the EXPERIMENT_SETTINGS_PROCESSING_ERROR action should already be implemented. In here you will only need to dispatch a `EXPERIMENT_SETTINGS_PROCESSING_ERROR` action. In the component part of data processing, you might need to add handling if the state is errored, but you have to check, maybe it is already implemented, I am not sure. Does that make sense?

cosa65: yes, I seems it is not created

cosa65: yes, `EXPERIMENT_SETTINGS_PROCESSING_ERROR` is only being used to dispatch it, but nothing in reducers takes it right now.

ivababukova: yeah, I just took a look and noticed that although there is a reducer for it: `processingSettingsError`, it is not hooked up in `redux/reducers/experimentSettings/index.js`, so you will need to do that too.

ivababukova (approved):

PR #127, opened 2021-02-24
Add configure embedding mitochondrial plot

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-490

https://ui-agi-mitochondrial-plot.scp-staging.biomage.net/

https://github.com/biomage-ltd/api/pull/53

- Implemented plot for pulling mitochondrial content data

# Changes
### Code changes

- Created a new plot for mitochondrial content : `MitochondrialContentPlot`
- Created a new store for storing cell metadata. This store can also be used to fetch other cell metadata, like doublet score

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/plots/MitochondrialContentPlot.jsx
```diff
@@ -0,0 +1,104 @@
...
+  );
+  const embedding = useSelector((state) => state.embeddings[embeddingSettings.method]) || {};
+
+  const mitochondrialContent = useSelector((state) => state.cellMeta?.mitochondrialContent) || {};
+  const cellSets = useSelector((state) => state.cellSets);
+  const [plotSpec, setPlotSpec] = useState({});
+
+  useEffect(() => {
+    if (cellSets.loading && !cellSets.error) {
+      dispatch(loadCellSets(experimentId));
```
cosa65, line 32: This is more to discuss, so let me know what you think about it.

This is something I noticed differs with other implementations of plots. In these cases, the plots are making the dispatches themselves while also managing the rendering directly, which means they refresh themselves on their own, while for example the elbow plot is just setting up the spec and receiving the data to display.

I feel like this option results in more repeated code since, for example, the `cellSets` loading part is always the same while perhaps by having the plot receive them as props we could avoid having it in all the embedding plots code.

And something similar goes for the processing settings, loadEmbedding (which differs only in the method) and others.

What im proposing would probably take a lot of refactoring on ConfigureEmbedding too though, so if you think this is worth trying then we could add a ticket for it or perhaps implement something less refactor-y for now like just separating the dispatches and useSelector/useState's by putting them in a wrapper that performs the loading and then passes the data and config over to the actual vega plot generating component.

cosa65: Something nice about this option too could be avoiding having one embedding for the each type of data it displays, instead having one loader for each type of data and only one single embedding plot generator

src/redux/reducers/cellMeta/cellMetaLoaded.js
```diff
@@ -0,0 +1,14 @@
+const cellMetaLoaded = (state, action) => {
+  const { metaName, data } = action.payload;
+
+  return {
+    ...state,
+    [metaName]: {
+      ...state[metaName],
+      loading: false,
```
cosa65, line 8: This should set error: false too.

src/components/plots/MitochondrialContentPlot.jsx
aerlaut: Yea, it would be good to be able to reuse plots with similar functionality. Generating the plot is also related with the plot's specGenerator and dataGenerator.  There may be a way to better implement these separate functions if we want to refactor the plot element.

I believe this can be put out as a ticket, as it requires some investigation how to best refactor the components.

aerlaut: On separate loading, it's good to move the dispatches outside the plots. However, this means any data loading the plot requires should be done in the data i.e. parent has to know the dependencies in child data. This can be alleviated with a good interface from the child, so it forces the parent to provide the required data. I think this goes hand in hand with refactoring of the plot components.

aerlaut: So yea I think we can move this refactoring out for another ticket.

src/redux/reducers/cellMeta/cellMetaLoaded.js
aerlaut: fixed

cosa65 (approved):

PR #129, opened 2021-02-24
Add error handling to steps configuration

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-587?atlOrigin=eyJpIjoiYmI3Nzk5YWQzNzFjNDhkOGI3MzdjZDZkN2E5YjEwMzYiLCJwIjoiaiJ9

https://ui-phdwdw9fgm7thd6pp7fo4w9bae.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


cosa65:
Something important about this is that the error handling in the case saveProcessingSettings fails is just an error message, there are other options that might be more ideal:

Automatic retries of the request:
- Making the application automatically retry saving until it is successful, this has the downside that it can cause some race conditions if the user re applies the configuration in the meantime, so to handle this we would need to either disable the apply button until there isn't an error anymore or have some flag that is set when the user clicks apply again and it causes the retries to stop.

Manual retries of the request:
- Reenabling the 'Apply button' so that the user can run the changes again. This is not that good an idea because it would also mean rerunning an embedding or data integration that is potentially already running. Even though it might (probably) have failed too, the failure of this request shouldnt be the cause of a re run of a different request.
- Something along the lines of the previous idea: Once there is an error: create a modal window or unhide a button somewhere in the calculationConfigs that allows the user to attempt to trigger a save again. 
- Another option along these lines could be modifying the NotificationManager allowing it to also receive an optional function that causes the error message to display a button to run the function.

marcellp:
I guess the main question is when we'd expect such a thing to happen. If it's when the network link is broken, I think it's fine to not do anything to try to manually resolve this as long as the user is aware of how they may be able to resolve it themselves.

In addition, I don't know yet how this system will fit into the QC pipeline as a whole, so I think it's probably better to go with the simpler solution and work on the rest as things become more clear.

Any suggestions @vickymorrison?

src/redux/actions/experimentSettings/loadProcessingSettings.js
```diff
@@ -34,10 +37,19 @@ const loadProcessingSettings = (experimentId) => async (dispatch, getState) => {
 
     throw new Error('HTTP status code was not 200.');
   } catch (e) {
+    dispatch(
+      pushNotificationMessage(
+        'error',
+        'We couldn\'t connect to the server to load processing settings from the previous session, please try refreshing the page',
+        5,
```
marcellp: The user doesn't need to know the technical background behind what caused the error to occur.

We can just say:

We couldn't load your processing settings. Please check your internet connection and try refreshing the page.

src/redux/actions/experimentSettings/completeProcessingStep.js
```diff
@@ -17,23 +24,53 @@ const completeProcessingStep = (
...
+      payload:
+        { experimentId, settingName, numSteps },
+    });
+  } catch (e) {
+    dispatch(
+      pushNotificationMessage(
+        'error',
+        'We couldn\'t connect to the server to save the current progress, please try again',
+        3,
+      ),
```
marcellp: Similarly, we can just say:

We couldn't save your processing settings. Please check your internet connection and try again.

vickymorrison:
> I guess the main question is when we'd expect such a thing to happen. If it's when the network link is broken, I think it's fine to not do anything to try to manually resolve this as long as the user is aware of how they may be able to resolve it themselves.
> 
> In addition, I don't know yet how this system will fit into the QC pipeline as a whole, so I think it's probably better to go with the simpler solution and work on the rest as things become more clear.
> 
> Any suggestions @vickymorrison?

I asked Martin the same question - are there other situations when this would happen or only when the user's connection fails? If the latter, then I suggest a slight re-wording to "We couldn't connect to the server to save the current progress. Check your internet connection and try again." Otherwise, it sounds like the issue is on our side rather than the user's side.

marcellp:
@vickymorrison I asked Martin to make it not refer to the "server" at all, as I thought it was extraneous detail that the user does not need, i.e. going with:

> We couldn't load your processing settings. Please check your internet connection and try refreshing the page.

> We couldn't save your processing settings. Please check your internet connection and try again.

What do you think?

vickymorrison:
> @vickymorrison I asked Martin to make it not refer to the "server" at all, as I thought it was extraneous detail that the user does not need, i.e. going with:
> 
> > We couldn't load your processing settings. Please check your internet connection and try refreshing the page.
> 
> > We couldn't save your processing settings. Please check your internet connection and try again.
> 
> What do you think?

That sounds good @marcellp

marcellp (approved):
LGTM after the error messages are fixed.

PR #134, opened 2021-03-02
Do cell set clustering from the platform

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-567

https://ui-martin-cell-set-clustering.scp-staging.biomage.net/

https://github.com/biomage-ltd/worker/pull/77

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


aerlaut:
Looks good to me, but would like to see the staging working. Is the deployment facing the same error has Juanlu?

ivababukova:
> Looks good to me, but would like to see the staging working. Is the deployment facing the same error has Juanlu?

The reason why the link is not working is because the liveness checks are failing, see the screenshot from Lens. During starting of a pod, Kubernetes uses liveness and readiness checks to figure out if a pod has started successfully and is ready for traffic or not (basically these are types of health checks if you are interested read more here: https://cloud.google.com/blog/products/containers-kubernetes/kubernetes-best-practices-setting-up-health-checks-with-readiness-and-liveness-probes). What these checks should be for a particular service is defined by us in the kubernetes deployment templates. For this deployment, the liveness and readiness checks are hitting the api endpoint ending on `/v1/heath` and getting a non-error response. You can see what the working health check endpoint in production returns: https://api.scp.biomage.net/v1/health
Here is where the deployment template for Martin's failing staging deployment is defined: https://github.com/biomage-ltd/iac/blob/master/releases/staging/martin-cell-set-clustering.yaml#L68 
Essentially, for some reason the api endpoint in staging ending on `/v1/health` returns 500 error. @marcellp can you help by telling us where and what to check to get any idea of what the possible causes for this may be?

Screenshot 2021-03-03 at 09 21 25

src/redux/reducers/cellSets/cellSetsClusteringUpdated.js
```diff
@@ -0,0 +1,15 @@
+import { createHierarchyFromTree, createPropertiesFromTree } from './helpers';
+
+const cellSetsClusteringUpdated = (state, action) => {
+  const { data } = action.payload;
+
+  return {
+    ...state,
+    loading: false,
+    updatingClustering: false,
+    hierarchy: createHierarchyFromTree(data),
```
ivababukova: have you tested in here that if the user reclusters, the Scratchpad cluster entry and the clusters under it, if they are any, do not disappear? We should also have a unit test for it

src/redux/reducers/cellSets/initialState.js
```diff
@@ -4,6 +4,7 @@ const initialState = {
   selected: [],
   loading: true,
   error: false,
+  updatingClustering: false,
```
ivababukova, line 7: It would be better to reuse `loading` - the purpose of it is to indicate if any parts of the data is currently under a change. Higher-level components check what is the state of 'loading' and if it is true, they know that they have to show the spinner. Is there any reason why we can't use it in here?

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -5,15 +5,19 @@ import {
...
+  if ((!loading && !error) || updatingClustering) {
     return null;
   }
   // There is only two way this action dispatcher does anything. Either
   // it is called after an error condition, or in a loading state. In the former
   // state, dispatching the loading action is unnecessary.
+
+  // Apart from this, if the clustering is being updated,
+  // this action dispatcher ends too since we will get a more up to date when the task is done
+
```
ivababukova: These comments are not great:
The first one is hard to read and super confusing - I had to read it like 3 times in order to understand what it talks about. And when I understood I was like "cool, so what?". For the bit that says `In the former state, dispatching the loading action is unnecessary.` if it is unnecessary, then why are we dispatching it anyway? The code here is quite short and simple, I don't think we need a comment here at all. Moreover, I think that this first comment only adds more confusion as to what is happening - makes it look like something very complex is going on, when it actually isn't. If I was you, I would just delete it.

For the second comment - I would move the comment to be before the if statement on line 11, because this is technically when the action will end by returning null.

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -5,15 +5,19 @@ import {
 
 const loadCellSets = (experimentId) => async (dispatch, getState) => {
   const {
-    loading, error,
+    loading, error, updatingClustering,
   } = getState().cellSets;
 
-  if (!loading && !error) {
+  if ((!loading && !error) || updatingClustering) {
```
ivababukova, line 11: why are we calling this action at all when doing cell set reclustering? where are we calling it from? Wouldn't it be easier to just prevent it from calling when `updatingClustering` is being called?

src/redux/actions/cellSets/loadCellSets.js
cosa65: Yes, I found the previous comments confusing too, I'll delete it.

About my own added comment, should I delete it too since it is nothing particularly complex?

src/redux/reducers/cellSets/initialState.js
cosa65: I think we need two separate properties.

If someone triggers a `loadCellSets` action while the clustering is being updated then we will probably get this new action to finish before `updateCellSetsClustering` does, which will set the state to `CELL_SETS_LOADED`.

Later on, `updateCellSetsClustering` will actually finish but in the ui it will look as if everything had already finished updating from before.

src/redux/reducers/cellSets/cellSetsClusteringUpdated.js
cosa65: You're right, I'll add handling for that.

src/redux/actions/cellSets/loadCellSets.js
ivababukova: yup, you can delete yours too

src/redux/reducers/cellSets/initialState.js
ivababukova: Yeah, I understand now why you decided to add it. However, looking at the action for `loadCellSets`, specifically at this piece of code:

```
if ((!loading && !error) || updatingClustering) {
    return null;
}
```
This means that if cell sets is in loading state, the action will return null anyway. Therefore, if we made `updateCellSetsClustering` use `loading` instead of `updatingClustering`, it will technically work. Or am I missing something?

cosa65: It returns null if cell sets is **not** in loading state. I think the reason that this is so is because the cell sets are in loading state when they begin, so that's when we would probably want to load (and avoid loading again after).

I didn't want to change that logic so that's why I added this.

src/redux/actions/cellSets/loadCellSets.js
cosa65: It is a case I ran into when testing the application with no louvain clustering.

Both configure embedding and some of the plots call loadCellSets, so instead of handling it everywhere I felt it was better to add a guard to the already existing ones here.

src/redux/reducers/cellSets/initialState.js
ivababukova: I see, somehow i skipped the `not` bit. Ok, I am convinced, sounds like the most sensible thing to do is to have this `updatingClustering` boolean

src/redux/actions/cellSets/loadCellSets.js
ivababukova: yeah, adding the guard here sounds more sensible than having to modify each of the individual ones

ivababukova (approved):

PR #135, opened 2021-03-03
Create modal for file upload

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-572

https://ui-135-agi.scp-staging.biomage.net/

N.A.

- Mocked creation of projects
- Implemented MIME type file validation
- Instructions default to uploading folder with files in it.
- File upload not yet implemented
- Create new project modal up automatically opens when there is no project
- Upload samples modal automatically open if project has no samples

# Changes
### Code changes

- Implemented Dropzone with simple file-type validation

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers
- [ ] (UX changes) Approved by @vickymorrison (this is her username, tag her if you need approval)

### Just before merging:

### Optional


src/pages/data-management/components/FileUploadModal.jsx
```diff
@@ -0,0 +1,210 @@
...
+
+  useEffect(() => {
+    setCanUpload(filesList.length);
+  }, [filesList]);
+
+  // Handle on Drop
+  const onDrop = (acceptedFiles) => {
+    const newList = [];
+
+    const acceptedFilesRegexp = `(${techOptions[selectedTech].acceptedFiles.join('|')})$`;
```
aerlaut, line 61: Enforce file naming convention here., as decribed under `techOptons[techName].acceptedFiles`, and informed to the user through `inputInfo`

src/pages/data-management/index.jsx
```diff
@@ -25,34 +27,46 @@ const DataManagementPage = ({ experimentId, experimentData, route }) => {
...
       setNewProjectModalVisible(false);
     }
   }, [projectsList]);
 
-  const createNewProject = () => {
+  const createNewProject = (newProjectName) => {
+    const newProject = {
+      name: newProjectName,
+      createdDate: moment().local().format('DD MMM YYYY, HH:mm:ss [GMT]Z'),
+      lastModified: moment().local().format('DD MMM YYYY, HH:mm:ss [GMT]Z'),
```
aerlaut, line 42: Shows date creation time according to local timezone. However, later on when the data is saved, it should be saved in UTC.

src/pages/data-management/index.jsx
```diff
@@ -25,34 +27,46 @@ const DataManagementPage = ({ experimentId, experimentData, route }) => {
...
+    };
+
+    setProjectsList([...projectsList, newProject]);
+    setActiveProjectIdx(projectsList.length);
     setNewProjectModalVisible(false);
   };
 
   const TILE_MAP = {
     'Projects List': {
-      toolbarControls: [<RemoveButton />],
```
aerlaut, line 41: Remove the ability for users to close the projects list window, so it's always there.

src/pages/data-management/index.jsx
```diff
@@ -25,34 +27,46 @@ const DataManagementPage = ({ experimentId, experimentData, route }) => {
...
           </Button>
-        </div>
+          <ProjectsList projects={projectsList} height={height} activeProjectIdx={activeProjectIdx} />
+        </Space>
       ),
     },
     'Data Management': {
-      toolbarControls: [
-        <RemoveButton />,
-      ],
```
aerlaut, line 53: This also remove the ability for users to close the project management window

vickymorrison:
I suggest updating the wording on the modal as follows:

Only 10x Chromium datasets are currently supported.

For each sample, the following three files are required:
    • barcodes.tsv or barcodes.tsv.gz
    • features.tsv or features.tsv.gz
    • matrix.mtx or matrix.mtx.gz

Drag and drop the folder containing these three files. The folder should be named with the sample ID.

Further guidance on supported file types and formats is available here.

The "here" should provide a link to the following PDF: https://drive.google.com/file/d/1qX6no9od4pi-Wy87Q06hmjnLNECwItKJ/view?usp=sharing

vickymorrison:
If the user attempts to upload invalid file types (as I just tried), should the "upload" button be disabled?

vickymorrison (approved):

aerlaut:
> If the user attempts to upload invalid file types (as I just tried), should the "upload" button be disabled?

Disable upload if there is invalid file, add option for user to delete the files and try to upload again.

src/pages/data-management/components/FileUploadModal.jsx
```diff
@@ -0,0 +1,229 @@
...
+import { CheckCircleTwoTone, CloseCircleTwoTone, DeleteOutlined } from '@ant-design/icons';
+import Dropzone from 'react-dropzone';
+
+const { Text, Title, Paragraph } = Typography;
+const { Option } = Select;
+
+const NewProjectModal = (props) => {
+  const { visible, onUpload, onCancel } = props;
+
+  const initialSelected = '10X Chromium';
```
cosa65: This const could be removed and we just pass '10X Chromium' to `useState`.

src/pages/data-management/components/FileUploadModal.jsx
```diff
@@ -0,0 +1,229 @@
...
+          <Title level={4} style={{ display: 'inline-block' }}>
+            Technology:
+            <span style={{ color: 'red', marginRight: '2em' }}>*</span>
+          </Title>
+          <Select style={{ width: 250 }} defaultValue={selectedTech}>
+            {Object.keys(techOptions).map((val, idx) => (
+              <Option key={idx} value={val}>{val}</Option>
+            ))}
+          </Select>
+          <Text type='secondary'><i>Only 10x Chromium datasets is currently supported</i></Text>
```
cosa65: I would change "is" for "are"

src/pages/data-management/components/FileUploadModal.jsx
```diff
@@ -0,0 +1,229 @@
...
+
+NewProjectModal.propTypes = {
+  visible: PropTypes.bool,
+  onUpload: PropTypes.func,
+  onCancel: PropTypes.func,
+};
+
+NewProjectModal.defaultProps = {
+  visible: true,
+  onUpload: null,
```
cosa65, line 224: I think I wouldn't consider `onUpload` as a prop that could be null or optional, I would set it as required.

`onCancel` perhaps could be optional but with an empty function as default? Because otherwise we would get an error, right?

cosa65:
I think there is a bug: In the initial module opening I get the disabled "Create project button" which is ok, but if I create a proejct and then open the modal again to create another project I get an enabled button by default,so I can create projects with empty project names

image

src/pages/data-management/components/ProjectsList.jsx
```diff
@@ -0,0 +1,85 @@
+import React, { useState, useEffect } from 'react';
```
cosa65: I think I would give this component a different name, something (`ProjectsListDisplayer`?) to differentiate itself from the actual list of projects (`projectsList`) that we have for example in `data-management/index.jsx`

src/pages/data-management/index.jsx
cosa65: Is this UTC task added to a ticket already?

src/pages/data-management/index.jsx
cosa65: Why do we want to remove this ability?

src/pages/data-management/components/FileUploadModal.jsx
aerlaut: it's fine to put the `onCancel` to `null` even though we require it as a function in the proptype : https://stackoverflow.com/questions/37842868/how-can-i-declare-a-proptype-corresponding-to-a-nullable-number

Regarding `onUpload`, it's not a prop that's required for rendering the component. That's why I set it to optional, i.e. the component can render even if it doesn't function yet. I think this is a pattern that's also used in UI components library, so we can quickly render the components without having to implement the functionalities yet.

src/pages/data-management/components/ProjectsList.jsx
aerlaut: I'll rename it `ProjectsListContainer`

aerlaut:
Reply to https://github.com/biomage-ltd/ui/pull/135#discussion_r586728582

I remove the ability for users to remove the display so that they will not be able to close the projects list / data management UI. These are the only 2 components that are in the page, and the `Data Management` part relies on `Projects List` to choose the currently opened project. Right now, there is no mechanism to reopen the `Projects List` and `Data Management` and to reload it users have to refresh the page. By removing the remove button, we'd essentially prevent the chance of missclicks from happening.

PR #136, opened 2021-03-04
Create redux store for projects

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-573

https://ui-agi-ui136-ehafobhdsdaaui6w.scp-staging.biomage.net/data-management

N.A.

- implemented actions, actionTypes and reducers for `projects`

# Changes
### Code changes

- Moved logic for project creation inside `ProjectListContainer`

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/redux/actions/projects/createProject.js
```diff
@@ -0,0 +1,28 @@
...
+
+const createProject = (
+  projectName,
+) => async (dispatch) => {
+  const createdAt = moment().local().format('DD MMM YYYY, HH:mm:ss [GMT]Z');
+
+  const newProject = {
+    ...projectTemplate,
+    name: projectName,
+    uuid: hash(projectName + moment().format('DDMMYYY')).slice(0, 10),
```
aerlaut: take only the first 10 character of generated hash, otherwise it becomes too long and unwieldy. Thoughts?

src/pages/data-management/components/ProjectsListContainer.jsx
```diff
@@ -1,23 +1,23 @@
...
 import PropTypes from 'prop-types';
 import { Card, Space } from 'antd';
 import { blue } from '@ant-design/colors';
 
 import FileUploadModal from './FileUploadModal';
+import { setActiveProject } from '../../../redux/actions/projects';
 
 const ProjectsListContainer = (props) => {
-  const { height, projects, activeProjectIdx } = props;
+  const { height } = props;
```
aerlaut, line 13: Move `projects` and `activeProjectIdx` to Redux store

src/redux/actions/projects/createProject.js
marcellp: Clipping hashes is kind of a bad idea because it does not guarantee randomness /and/ it is also hard to read.

If we don't care about possible collisions, we can use an adjective-animal system for the projects (https://github.com/a-type/adjective-adjective-animal).

```
generate(3).then(console.log);
// "mortified-industrious-albatross"
```

These are easy to remember and also cute. If we want to avoid collisions altogether we should use UUIDs (there is a package for it in the UI already) and just accept that they are long.

src/redux/actions/projects/createProject.js
```diff
@@ -0,0 +1,28 @@
...
+
+import {
+  PROJECTS_CREATE,
+} from '../../actionTypes/projects';
+import { projectTemplate } from '../../reducers/projects/initialState';
+
+const createProject = (
+  projectName,
+) => async (dispatch) => {
+  const createdAt = moment().local().format('DD MMM YYYY, HH:mm:ss [GMT]Z');
```
marcellp: I recommend changing this from a UX and accessibility point of view.

Using custom date strings such as this one hardcoded into Redux is not great. Many languages/locales don't use the type of format that is in here (`04 Mar 2021, 16:13:40 GMT+00:00`). It is also not standard, so it is hard to get an actual date object from this string in the future.

Second, it is in general not a good idea to tell people actual dates for when things happened, relative dates are easier to parse and think about, and the actual proper data usually becomes visible on hover (You can see this yourself even on this page). This also resolves the issue you had with timezones.

In general, the best practice would mean the following: for createdAt and lastModified, you can simply use the ISO string. This ensures cross-compatibility with different languages, and a universal way to store dates as text.

```
> const createdAt = moment().toISOString()
undefined
> createdAt
'2021-03-04T16:20:20.373Z'
```

Now, you need some way to display this on the UI. It is almost always a good idea when displaying dates on the UI to always display relative dates as the main string, but to also show the proper date if you hover over the text:

```
> moment(createdAt).fromNow()
'a minute ago'
```

For displaying dates in general in UI that are not relative (such as the full date on hover), it is a good idea to always use the localized version of the date so we will be able to later support different locales/regions. You can see [here](https://momentjs.com/docs/#/displaying/) that you can use `LLLL` as a format string that will automatically format the date according to the locale of the user. On my computer the default locale is `en-US` so I get (with no settings):

```
> moment(createdAt).format('LLLL')
'Thursday, March 4, 2021 4:20 PM'
```

But if we were to support say Polish, British English, Hungarian, or Indonesian in the future, we can do that too, and the same string adjusts to the way dates are usually written in that region:

```
> moment.locale('pl-PL')
'pl'
> moment(createdAt).format('LLLL')
'czwartek, 4 marca 2021 16:20'

> moment.locale('en-GB')
'en-gb'
> moment(createdAt).format('LLLL')
'Thursday, 4 March 2021 16:20'

> moment.locale('hu-HU')
'hu'
> moment(createdAt).format('LLLL')
'2021. március 4., csütörtök 16:20'

> moment.locale('id-ID')
'id'
> moment(createdAt).format('LLLL')
'Kamis, 4 Maret 2021 pukul 16.20'
```

src/redux/actions/projects/updateProject.js
```diff
@@ -0,0 +1,20 @@
+import _ from 'lodash';
+
+import {
+  PROJECTS_UPDATE,
+} from '../../actionTypes/projects';
+
+const createProject = (
```
marcellp: The name here still says `createProject`.

Isn't the `lastModified` value meant to also update here?

src/redux/actions/projects/createProject.js
aerlaut: Avoding clashes is the priority. I'll change it to using UUID.

src/redux/actions/projects/updateProject.js
aerlaut: oh yea :S fixed in commit d78e4f3

src/redux/actions/projects/createProject.js
aerlaut: Added in commit d78e4f3

src/redux/actionTypes/projects.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Save project.
+ */
+const PROJECTS_CREATE = `${PROJECTS}/create`;
+
+/**
+ * Update project.
+ */
+const PROJECTS_UPDATE = `${PROJECTS}/update`;
```
ivababukova, line 11: how about deleting a project?

src/redux/actions/projects/createProject.js
```diff
@@ -0,0 +1,28 @@
...
+
+const createProject = (
+  projectName,
+) => async (dispatch) => {
+  const createdAt = moment().toISOString();
+
+  const newProject = {
+    ...projectTemplate,
+    name: projectName,
+    uuid: uuidv4(),
```
ivababukova, line 17: what is the relationship between this `uuid` in here and `experimentId`? Where and what for is `uuid` used? Where do we record the experiment id?

src/redux/actions/projects/createProject.js
```diff
@@ -0,0 +1,28 @@
+import moment from 'moment';
+import { v4 as uuidv4 } from 'uuid';
+
+import {
+  PROJECTS_CREATE,
+} from '../../actionTypes/projects';
+import { projectTemplate } from '../../reducers/projects/initialState';
+
+const createProject = (
```
ivababukova, line 9: what do we do with the rest of the redux store when we have multiple projects? How do we know if, for example, the contents of differential expression in redux are relevant for the selected project, or are actually for a different project?

src/redux/actionTypes/projects.js
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Update project.
+ */
+const PROJECTS_SET_ACTIVE = `${PROJECTS}/setActive`;
```
ivababukova, line 16: why is there a second action for updating a project?

aerlaut: This one is for setting the active project

src/redux/actionTypes/projects.js
aerlaut: Deleting and updating a project will be separated into different tickets as they are more involved. Deleting a project requires deleting the related samples and also files. Likewise with updating.

src/redux/actionTypes/projects.js
ivababukova: can you update the comment on top of it in this case?

src/redux/actionTypes/projects.js
ivababukova: ok, makes sense!

src/redux/actions/projects/createProject.js
aerlaut: Oh yea, I'll add an array to store the experiments related to the project. Each experiment will also have a `projectUuid` to point back to the parent project,

src/redux/actions/projects/createProject.js
aerlaut: The contents of for data processing, data exploration and plots and tables are actually related to experiments, not projects. So I imagine after selecting an experiment, the contents of the relevant stores will updated according to the experimentId.

src/redux/actions/projects/createProject.js
aerlaut: updated in c5355d9

src/redux/actions/projects/createProject.js
ivababukova: Yes, they are related to experiments, this is true. However, at the moment the contents of the redux store are not associated with an experiment, which means that if a user has viewed experiment `asd` before and decided now to view another experiment `abc`, redux will not dispatch new actions to update the data in the store. Is this something that we are planning to fix for the MVP, or are we going to support only 1:1 mapping between experiments and projects for now?

aerlaut: Will have it ready for MVP in a separate ticket

ivababukova (approved):

PR #140, opened 2021-03-05
Create Redux store to track QC status

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-608

TBD

N/A

As agreed with Agi, the only purpose of this ticket is to track whether or not the pipeline is currently running. A Redux store did not need to be created because we can query the API directly for this data. In addition, this data may be subject to change relatively quickly and it is more difficult to block high-level components like pages based on the contents of the Redux store.

It also includes a sample component on how to block out pages and paths if a pipeline is running. This is partially in scope for another ticket, BIOMAGE-498, but it was necessary in order to test properly if the alternative way I proposed meets the requirements for the downstream tickets.

# Changes
### Code changes

Added a check in `_app.jsx` that automatically queries the state of the pipeline on page load. Depending on the data, it can trigger ContentWrapper to disable or enable pages from loading. We can also potentially use this data downstream in BIOMAGE-498 directly, depending on the implementation there.

# Definition of DONE

### Testing
- [x] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
image


marcellp:
Changes requested by @aerlaut are done.

src/components/ContentWrapper.jsx
```diff
@@ -16,25 +15,49 @@ import {
...
+    dispatch(loadPipelineStatus(experimentId));
+    updateSocket.current = initUpdateSocket(experimentId, (res) => { console.log(res); });
   }, [experimentId]);
 
+  useEffect(() => {
+    if (pipelineStatusRequested) {
+      return;
+    }
+
+    setPipelineStatusRequested(true);
```
marcellp, line 63: This is so we prevent pages from loading between page load and the time that `loadPipelineStatus` triggers `pipelineLoading ` to be set to `true`.

If we've already been to the website, `pipelineLoading` will be `false` and so the page would immediately load until the loading event triggers, at which point it would change to the loading state. To avoid this race condition, this variable is used.

package.json
```diff
@@ -125,6 +124,7 @@
     "vega": "^5.13.0",
     "vega-lite": "^4.13.0",
     "vitessce": "^0.2.1",
-    "webpack": "^4.43.0"
+    "webpack": "^4.43.0",
+    "react-dropzone": "^11.3.1"
```
ivababukova, line 128: was this an intended move?I am just asking because of the recent socket io issues we had in master

src/components/ContentWrapper.jsx
```diff
@@ -16,25 +15,54 @@ import {
...
+  const updateSocket = useRef(null);
   useEffect(() => {
-    if (experimentId) {
-      updateSocket.current = initUpdateSocket(experimentId, (res) => { console.log(res); });
+    if (!experimentId) {
+      return;
     }
+
+    dispatch(loadPipelineStatus(experimentId));
+    updateSocket.current = initUpdateSocket(experimentId, (res) => { console.log(res); });
```
ivababukova, line 55: you can just call the `initUpdateSocket` function, no need for the console log

ivababukova: Also, I don't see this `updateSocket` variable used anywhere, is the plan to use it later? And how would we use it?

ivababukova:
Does this PR mean that the `meta` part in the `experimentSettings` store is no longer needed?

package.json
marcellp: Yes, it is not necessary to have it in dependencies, it is not directly in the dependency graph of `_app.jsx`

src/components/ContentWrapper.jsx
marcellp: > you can just call the `initUpdateSocket` function, no need for the console log

There is a reason, it is so we can test changes we make to the pipeline end-to-end until the UI updates are done.

marcellp: > Also, I don't see this `updateSocket` variable used anywhere, is the plan to use it later? And how would we use it?

That is the socket that will relay pipeline updates back to the UI or anything that is not directly associated with a work request submitted to the workers. This was created based on https://biomage.atlassian.net/browse/BIOMAGE-516 and the code was just moved right now.

marcellp:
> Does this PR mean that the `meta` part in the `experimentSettings` store is no longer needed?

I think it may not be required, but it's worth waiting with this until Xavier is done with his API changes at https://biomage.atlassian.net/browse/BIOMAGE-513 and we get to the UX design of reacting to updates. We will probably end up refactoring that code slightly, but it's worth keeping it in for now imho.

src/components/ContentWrapper.jsx
ivababukova: ok, all these make sense, thanks for explaining!

ivababukova (approved):
looks good to me

marcellp:
Will wait for @aerlaut if this is sufficient for meeting the criteria of his ticket as he asked for some refactoring. If he is satisfied I will merge.

aerlaut:
@marcellp Looks good to me 👍

PR #149, opened 2021-03-15
Mirror heatmap controls

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-467
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-626
https://ui-plsworkffs-ui149.scp-staging.biomage.net/

# Changes
The selected genes from the heatmap are now stored in genes/expression/views/[plotUuid]
During developing I realised if we don't store the selected genes in the config anymore, they won't be persisted. For now the selected genes are also stored in the config so that we can save them.
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


ivababukova:
I just played around with it locally and I noticed that the heatmap in data exploration doesn't render properly - every time I add a gene, there are write lines all over the heatmap (see example screenshot). The problem could be that you are not making the heatmap enter loading state when we are adding/removing/overriding genes. Try adding the loading state back and this will probably get fixed.

[](url) 
Screenshot 2021-03-15 at 11 08 05

StefanBabukov:
> I just played around with it locally and I noticed that the heatmap in data exploration doesn't render properly - every time I add a gene, there are write lines all over the heatmap (see example screenshot). The problem could be that you are not making the heatmap enter loading state when we are adding/removing/overriding genes. Try adding the loading state back and this will probably get fixed.
> 
> 
> Screenshot 2021-03-15 at 11 08 05

How do I reproduce it, I haven't been able to get those white lines.

ivababukova:
I was just adding genes to the heatmap. But now I noticed that it appears on the master version of localhost, so maybe it is not from your PR. However, you do still need to trigger the heatmap into loading state when changing what genes to display.

src/components/ContentWrapper.jsx
```diff
@@ -47,9 +47,10 @@ const ContentWrapper = (props) => {
   const pipelineStatusKey = pipelineStatus.pipeline?.status;
   const pipelineErrors = ['FAILED', 'TIMED_OUT', 'ABORTED'];
 
-  const pipelineRunning = pipelineStatusKey === 'RUNNING';
-  const pipelineRunningError = pipelineErrors.includes(pipelineStatusKey);
-
+  // const pipelineRunning = pipelineStatusKey === 'RUNNING';
+  // const pipelineRunningError = pipelineErrors.includes(pipelineStatusKey);
+  const pipelineRunning = false;
+  const pipelineRunningError = false;
```
ivababukova: @cosa65 do we need that merged in? Or is its purpose just to enable testing on staging?

StefanBabukov: I changed this temporarily so that staging can be opened I intended to revert it before merging.

ivababukova: sounds good, just make sure you don't forget about it

src/utils/plotSpecs/generateHeatmapSpec.js
```diff
@@ -1,45 +1,45 @@
...
-        symbolSize: { value: 200 },
-        encode: {
-          labels: {
-            update: {
-              text: { scale: 'cellSetIDToName', field: 'label' },
-            },
+    },
+    {
+      fill: 'cellSetColors',
+      title: 'Cluster ID',
```
ivababukova: why hardcoding the title here?

src/utils/plotSpecs/generateHeatmapSpec.js
```diff
@@ -1,45 +1,45 @@
...
-        type: 'gradient',
-        orient: 'bottom',
-        direction: 'horizontal',
-        title: ['Intensity'],
-        labelFont: { value: config.fontStyle.font },
-        titleFont: { value: config.fontStyle.font },
-        gradientLength: {
-          signal: 'width',
-        },
+const generateSpec = (config) => {
```
ivababukova: is this spec used for both heatmaps (data exploration and on plots and tables)?

src/redux/reducers/componentConfig/updateConfig.js
```diff
@@ -4,7 +4,14 @@ const updateConfig = (state, action) => {
...
 
   const newConfig = _.cloneDeep(state[plotUuid].config);
-  _.merge(newConfig, configChange);
+
+  const mergeConfig = (objValue, srcValue) => {
+    if (_.isArray(objValue) && srcValue) {
+      return srcValue;
+    }
+  };
+
```
ivababukova, line 12: why was this change necessary?

src/utils/plotSpecs/generateHeatmapSpec.js
StefanBabukov: Because the legend displays the clusterIds that are shown (from metadata tracks) or does this need to be variable changed?
image

src/utils/plotSpecs/generateHeatmapSpec.js
StefanBabukov: No the data-exploration heatmap spec is different (has tooltips, no second legend, mouseOver events, no config-set properties)

src/redux/reducers/componentConfig/updateConfig.js
StefanBabukov: Because before if you have a property e.g `something:['a',' b', 'c']` and you want to set it to ['a'] - the config change. The way that merge works (not an intersection) it won't overwrite the previous value of **something** to be equal to just ['a'] but instead it will remain unchanged. So this way we check if the value is an array and if the new value exists we set the previous one to the new one instead of doing a normal merge.

cosa65:
There's merge conflicts.

src/components/plots/helpers/populateHeatmapData.js
```diff
@@ -0,0 +1,216 @@
...
+import { union } from '../../../utils/cellSetOperations';
+
+const populateHeatmapData = (cellSets, config, expression, selectedGenes, downsampling = false) => {
+  const { hierarchy, properties, hidden } = cellSets;
+  const {
+    selectedTracks, groupedTracks, expressionValue,
+  } = config;
+  const maxCells = 1000;
+  const getCellsInSet = (cellSetName) => properties[cellSetName].cellIds;
+  // const selectedGenes = expression.views[componentType].data;
```
cosa65: Remove this

src/utils/plotSpecs/generateHeatmapSpec.js
cosa65: We are using the heatmap in data integration with the titles like this: https://docs.google.com/presentation/d/1-6mtMrZF14GRAYtLjh3F7FmhyE4o5WRpwyxt5gISGDo/edit#slide=id.gae6b21872b_1_12

(with coloring by sample), before it was set up to allow for the title to be set according to what we need, is this causing problems on that side?

src/redux/reducers/componentConfig/updateConfig.js
ivababukova: I didn't understand your answer, can you give me a specific example of a case when that happens?

src/utils/plotSpecs/generateHeatmapSpec.js
ivababukova: no, I meant whether this file "src/plotSpecs/generateHeatmapSpec.js" is used for generating both the data exploration and the plots and tables heatmap

src/utils/plotSpecs/generateHeatmapSpec.js
ivababukova: It is always better to use a variable rather than hardocding that value. When you hardocde it like this, in the future if we wanted to display something else, not cluster id, we would have to change the config and adapt to it. Hardcoding values makes the code more fragile

src/utils/plotSpecs/generateHeatmapSpec.js
StefanBabukov: It is only for plots and tables heatmap

src/redux/reducers/componentConfig/updateConfig.js
StefanBabukov: if we have a config field `{something: ['a', 'b','c']}` and we want to change it to `{something:['a']}` before it was `_.merge({something: ['a', 'b','c']}, {something:['a']}) ` - this would still return` {something: ['a', 'b','c']}`, not  `{something:['a']}`  due to how merge works https://lodash.com/docs/4.17.15  Example would be when you add and remove metadata tracks or selected genes - it was impossible to remove a selected gene before because the array would stay the same.

src/utils/plotSpecs/generateHeatmapSpec.js
StefanBabukov: I didn't get what you mean, I don't see a heatmap in the slides?

cosa65 (approved):

src/redux/reducers/componentConfig/updateConfig.js
ivababukova: thanks Stefan! I understand now

ivababukova (approved):

PR #156, opened 2021-03-18
Integration with pipeline

ogibson:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-640

https://ui-martinfosco-ui156.scp-staging.biomage.net/

This PR is meant to allow receiving data from the pipeline, effectively replacing the previous mock data in the pipeline plots. Nevertheless, take into account that for the new data to get to the plots a filter must be run, and the data received from the pipeline. 

# Changes
### Code changes

Created a new function to update pipeline plot data
Changed the way plot data is updated to modify the data when we have new plot data. 

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X ] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
- [X ] Photo of a cute animal attached to this PR

image

### DIFF
--- a/src/redux/actions/componentConfig/updatePlotData.js
+++ b/src/redux/actions/componentConfig/updatePlotData.js
@@ -1,10 +1,10 @@
 import { UPDATE_CONFIG } from '../../actionTypes/componentConfig';
 
-const updatePlotData = (plotUuid, dataChange) => (dispatch) => {
+const updatePlotData = (plotUuid, dataChanges) => (dispatch) => {
   dispatch({
     type: UPDATE_CONFIG,
     payload:
-      { plotUuid, dataChange },
+      { plotUuid, dataChanges },
   });
 };
 
--- a/src/redux/actions/componentConfig/updatePlotData.js
+++ b/src/redux/actions/componentConfig/updatePlotData.js
@@ -1,0 +1,11 @@
+import { UPDATE_CONFIG } from '../../actionTypes/componentConfig';
+
+const updatePlotData = (plotUuid, dataChange) => (dispatch) => {
+  dispatch({
+    type: UPDATE_CONFIG,
+    payload:
+      { plotUuid, dataChange },
+  });
+};
+
+export default updatePlotData;
--- a/src/redux/reducers/componentConfig/updateConfig.js
+++ b/src/redux/reducers/componentConfig/updateConfig.js
@@ -1,14 +1,14 @@
 import _ from 'lodash';
 
 const updateConfig = (state, action) => {
-  const { plotUuid, configChange = null, dataChange = null } = action.payload;
+  const { plotUuid, configChanges = null, dataChanges = null } = action.payload;
   let newConfig = state[plotUuid].config;
-  if (configChange) {
+  if (configChanges) {
     newConfig = _.cloneDeep(state[plotUuid].config);
-    _.merge(newConfig, configChange);
+    _.merge(newConfig, configChanges);
   }
 
-  const newData = dataChange ?? state[plotUuid].plotData;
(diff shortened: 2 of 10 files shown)


src/redux/actions/componentConfig/updatePlotData.js
```diff
@@ -0,0 +1,11 @@
+import { UPDATE_CONFIG } from '../../actionTypes/componentConfig';
+
+const updatePlotData = (plotUuid, dataChange) => (dispatch) => {
```
ivababukova: why couldn't the already existing `updatePlotConfig` action be reused here? Why are we creating a new action for this?

src/redux/reducers/componentConfig/updateConfig.js
```diff
@@ -1,16 +1,21 @@
 import _ from 'lodash';
 
 const updateConfig = (state, action) => {
-  const { plotUuid, configChange } = action.payload;
+  const { plotUuid, configChange = null, dataChange = null } = action.payload;
```
ivababukova: my understanding is that `configChange` is a boolean, is that the case? If that is the case, shouldn't the default value for it be either `true` or `false`?

src/redux/reducers/componentConfig/updateConfig.js
```diff
@@ -1,16 +1,21 @@
...
+  let newConfig = state[plotUuid].config;
+  if (configChange) {
```
ivababukova: what does this if statement evaluate to when `configChange` is null?

src/redux/actions/componentConfig/updatePlotData.js
cosa65: Because `updatePlotConfig` was being used for the config side.
I could refactor `updatePlotConfig` to take `configChange` and `DataChange` instead though, and reuse it in `updatePlotData`, let me know which way would be better. 
Or something like `{configChange:..., plotDataChange: ...}`

src/redux/reducers/componentConfig/updateConfig.js
cosa65: configChange is an object with all the changes in the plot's styling config, since now we need to be able to update the plotData too then we added dataChange.

src/redux/reducers/componentConfig/updateConfig.js
cosa65: If it is null it doesn't merge `newConfig` with the previous config.

src/redux/actions/componentConfig/updatePlotData.js
ivababukova: I see, I think that there are valid arguments for either option.
Looking at the name of the action type(`UPDATE_CONFIG`), it applies to both `configChange` and `plotDataChange`. From that point of view, it might make more sense to do what you proposed in your comment here. 
On the other hand, it could be a pain to implement this, because you might need to change a lot of places in the UI to pass in `configChange: ...` or `plotDataChange: ....`. Also, having separate actions for them (`updatePlotData` and `updatePlotConfig`) makes the code clearer and more explicit. 
Thinking about it, probably what you have implemented in this PR is the better option.

src/redux/reducers/componentConfig/updateConfig.js
ivababukova: with risk of being a bit pedantic, if you rename it to `configChanges` and `dataChanges`, that might make it much clearer what these are straight-away

src/redux/reducers/componentConfig/updateConfig.js
ivababukova: fair enough

ivababukova (approved):

PR #194, opened 2021-04-07
Implement data upload mechanism (locally)

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-575

 https://ui-martinfosco-ui194.scp-staging.biomage.net/

https://github.com/biomage-ltd/inframock/pull/38

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/processUpload.js
```diff
@@ -1,8 +1,88 @@
...
+              status: UploadStatus.UPLOAD_ERROR,
+            }));
+          });
+      },
+      async () => {
+        dispatch(updateSampleFile(sample.uuid, {
+          ...file,
+          status: UploadStatus.UPLOAD_ERROR,
+        }));
+      },
```
aerlaut: These dispatched functions look like the `actions` functions that we usually implement, dispatching `actionTypes`. Any reason why you decide to implement it this way instead of using the common pattern?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -232,17 +234,24 @@ const ProjectDetails = ({ width, height }) => {
...
-      matrix: _.sample(statuses),
-      species: 'dataMissing',
-    }));
-
+    const newData = projects[activeProjectUuid].samples.map((sampleUuid, idx) => {
+      const sampleFiles = samples[sampleUuid].files;
+
+      const barcodesStatus = sampleFiles['barcodes.tsv.gz']?.status;
+      // TODO:  File name for genes can be one of these, should we rename to have only one possible name before putting it in S3?
+      const genesStatus = (sampleFiles['genes.tsv.gz'] ?? sampleFiles['features.tsv.gz'])?.status;
```
aerlaut: It's ok to leave it. It's already handled in data-ingest: https://github.com/biomage-ltd/data-ingest/blob/cc303b3e2b942ff0b0e323c04671c98ffb17968b/src/1_Preproc.r#L35-L39

src/pages/_app.jsx
```diff
@@ -20,6 +25,54 @@ Router.events.on('routeChangeStart', () => NProgress.start());
 Router.events.on('routeChangeComplete', () => NProgress.done());
 Router.events.on('routeChangeError', () => NProgress.done());
 
+const setupAmplify = () => {
```
aerlaut: Would it be cleaner to refactor this into it's own "init" file?

aerlaut:
Where is the data upload triggered?

src/utils/processUpload.js
cosa65: I'm not sure I understand the question, the `updateSampleFile` was already being used somewhere else too, it is a normal action function.

If the question is more about using the `UploadStatus` enum for the statuses I believe it is a clean way to standardize keys, which helps in making the code less prone to errors because instead of using the strings everywhere we use the constants storing them.

This way in the future we only need to modify the strings in the enum to change those keys for all code that depends on it without fear of breaking anything.

cosa65: Oh I understand now I think, you mean moving all this logic to an action, you're right that it doesn't fit that well with the logic followed elsewhere.

I guess since all of this was running from a function and not directly from a component I got kind of in a grey area in this respect, you're right that it makes sense to move all the logic of uploading and interacting with amplify to a particular dispatch.

aerlaut: Yea, I mean instead of using `updateSampleFile` and passing in `status: UploadStatus`, would it be cleaner to have `UploadStatus`as an action type, and then create new action to handle that e.g. `fileUploaded`, `fileError`

src/pages/_app.jsx
```diff
@@ -33,6 +86,8 @@ const WrappedApp = ({ Component, pageProps }) => {
     if (router.route.includes('experimentId')) {
       setExperimentId(router.query.experimentId);
     }
+
+    setupAmplify();
```
cosa65: move this somewhere else, it doesn't depend on the experimentId and leave comments on why it is here

src/utils/processUpload.js
cosa65: I mainly followed the logic in the whole processUpload file which does actually update other properties following this same pattern, so we would have to probably move most of this function over to redux if we wanted to follow a different pattern.

Furthermore, moving this to an action would involve moving the Storage.put() to an action too, which might be shifting part of this function's functionality over to a redux action instead of keeping it here.

src/pages/_app.jsx
cosa65: turns out the checks in this useEffect are useful for setupAmplify, so it might be ok to keep it here

aerlaut (approved):
LGTM, tested locally

PR #196, opened 2021-04-08
Handle naming collision

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-609

https://ui-agi-ui196.scp-staging.biomage.net/

N.A.

- Implemented check for same project name when creating new project
- Implemented check when renaming project names and sample names
- Show what's wrong with the project / sample names.

# Changes
### Code changes

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/components/EditableField.jsx
+++ b/src/components/EditableField.jsx
@@ -1,20 +1,33 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import {
-  Button, Input, Space, Tooltip,
+  Button, Input, Space, Tooltip, Typography,
 } from 'antd';
 
 import {
   EditOutlined, DeleteOutlined, CheckOutlined, CloseOutlined,
 } from '@ant-design/icons';
 
+const { Text } = Typography;
+
 const EditableField = (props) => {
   const {
-    value, deleteEnabled, showEdit, onAfterSubmit, onAfterCancel, renderBold, defaultEditing,
+    value,
+    deleteEnabled,
+    showEdit,
+    onAfterSubmit,
+    onAfterCancel,
+    renderBold,
+    defaultEditing,
+    validationFunc,
+    errorText,
   } = props;
 
   const [editing, setEditing] = useState(defaultEditing);
   const [editedValue, setEditedValue] = useState(value);
+  const [isValid, setIsValid] = useState(true);
+  const [errorCode, setErrorCode] = useState(null);
+
   useEffect(() => {
     setEditedValue(value);
   }, [value]);
@@ -36,17 +49,28 @@ const EditableField = (props) => {
   const onChange = (e) => {
     const { value: newValue } = e.target;
 
+    // Returns true on valid, else error code
+    const validation = validationFunc(newValue);
+    if (validation !== true) {
+      setErrorCode(validation);
+      setIsValid(false);
+    } else {
+      setIsValid(true);
+    }
+
     setEditedValue(newValue);
   };
 
   const onSubmit = (e) => {
     e.stopPropagation();
+    if (!isValid) return null;
     onAfterSubmit(editedValue);
     toggleEditing(e);
   };
 
   const onCancel = (e) => {
     e.stopPropagation();
+    if (!isValid) setIsValid(true);
     setEditedValue(value);
     toggleEditing(e);
     onAfterCancel();
@@ -97,17 +121,28 @@ const EditableField = (props) => {
   };
 
   return (
-    <Space>
-      {renderEditState()}
-      {
-        deleteEnabled
-          ? (
-            <Tooltip placement='bottom' title='Delete' mouseLeaveDelay={0}>
-              <Button size='small' shape='circle' icon={<DeleteOutlined />} onClick={deleteEditableField} />
-            </Tooltip>
-          ) : <></>
-      }
-    </Space>
+    <>
+      <Space direction='vertical'>
+        <Space>
+          {renderEditState()}
+          {
+            deleteEnabled
+              ? (
+                <Tooltip placement='bottom' title='Delete' mouseLeaveDelay={0}>
+                  <Button size='small' shape='circle' icon={<DeleteOutlined />} onClick={deleteEditableField} />
+                </Tooltip>
+              ) : <></>
+          }
+        </Space>
+        {!isValid ? (
+          <Text type='danger'>
+            {typeof errorText === 'string'
+              || React.isValidElement(errorText) ? errorText
+              : errorText[errorCode]}
+          </Text>
+        ) : <></>}
+      </Space>
+    </>
   );
 };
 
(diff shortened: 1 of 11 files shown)


src/components/EditableField.jsx
```diff
@@ -97,24 +119,37 @@ const EditableField = (props) => {
...
+      </Space>
+    </>
   );
 };
 
 EditableField.defaultProps = {
   onAfterSubmit: () => null,
   onAfterCancel: () => null,
   onDelete: () => null,
+  validationFunc: null,
```
ivababukova: if it is a func, I think a more correct way to have is as `validationFunc: () => null`

src/components/EditableField.jsx
```diff
@@ -97,24 +119,37 @@ const EditableField = (props) => {
...
+  errorFunc: () => null,
+  errorText: '',
```
ivababukova: Do we actually need this `errorText` variable? I didn't see it set anywhere in the code

src/components/data-management/ProjectDetails.jsx
```diff
@@ -158,6 +166,8 @@ const ProjectDetails = ({ width, height }) => {
         value={text}
         onAfterSubmit={(name) => dispatch(updateSample(el.uuid, { name }))}
         onDelete={() => dispatch(deleteSamples(el.uuid))}
+        validationFunc={(name) => validateSampleName(name, sampleNames)}
+        errorFunc={(name) => validateSampleName(name, sampleNames)}
```
ivababukova: do we need to differentiate between `validationFunc` and `errorFunc`? I see that in practise, they are both equal to the same thing

src/utils/validateProjectName.js
```diff
@@ -0,0 +1,39 @@
+const defaultErrorMessages = {
+  MIN_8: 'Min 8 characters',
+  MIN_2_CHAR: 'Min 2 characters in sequence',
+  INVALID_CHARS: 'Invalid characters',
+  NAME_EXISTS: 'A project with the same name exists',
```
ivababukova: are we going to display this text to the user upon error in validation? If that will be the case, all of these apart from `NAME_EXISTS` will need rephrasing to be clearer

src/utils/validateSampleName.js
```diff
@@ -0,0 +1,29 @@
+const errorMessages = {
+  MIN_1_CHAR: 'Min 1 alphanumeric character',
+  INVALID_CHARS: 'Invalid characters',
+  NAME_EXISTS: 'A sample with the same name exists',
+};
+
+const validateSampleName = (input, sampleNames) => {
```
ivababukova: Where is the validation difference between sample and project names coming from and why that difference?

src/components/data-management/ProjectDetails.jsx
aerlaut: In this case, the validation function returns the error message, that's why it's set like that. I imagine there are cases where the validation function is quite simple `((name) => name !== 'projects')`, and does not return the text, then the `errorText` can be used

src/components/EditableField.jsx
aerlaut: `errorText` can be used together with `validationFunc` in the case where `validationFunc` does not return a value, and the `errorText` is quite simple. But yea, I think you're right, this can also be simply implemented as `errorFunc : () => 'Error message' `

aerlaut: I'll fix this up

src/components/data-management/ProjectDetails.jsx
ivababukova: are we going to have a case like this any time soon?

src/components/EditableField.jsx
ivababukova: sounds good!

src/components/EditableField.jsx
aerlaut: I'll have to test this, because there's the check if `validationFunc` is set or not, and having it as `() => null` means the check will always evaluate to true. Maybe it's better to set it as `validationFunc: undefined`

src/utils/validateProjectName.js
aerlaut: Yes, but it couldn't be too long because it's used in tight spaces. Do you have any suggestions to improve it?

ivababukova: Where do we use it? Maybe you can send me a screenshot?

ivababukova: I will actually take a look at staging

src/utils/validateSampleName.js
aerlaut: The project name validation is enforced so that the project name will be more verbose and descriptive, like 'Project 1'

The sample name, on the other hand, can be simpler, e.g. `WT1`, `KO`, 'WT2', because they're already scoped inside the project.

ivababukova:
I just checked staging and I made a video of a tiny usability issue -- after I click on the edit button to change the name of the project, after I click on the part of the name that I want to start typing characters from, the "Technology" window pops up, which doesn't look and feel very right. The user should be able to click on and highlight any part of that text without that window showing up

https://user-images.githubusercontent.com/65414651/114041524-e6c97700-9884-11eb-8928-cd5d925509c1.mov

src/utils/validateProjectName.js
ivababukova: yes, I took a look, it looks fine as it is

src/utils/validateSampleName.js
ivababukova: hmm I understand now the rationale, makes sense

src/components/EditableField.jsx
ivababukova: `if validationFunc is set or not` <-- if `validationFunc` is a function, how can it be set? It can be either called or not, right?

aerlaut: `validationFunc` should only be provided to the component if we want validation enabled, so it should be called if it is provided. Therefore, this requires a check if `validationFunc` is passed in (not undefined). Setting its default value to `validationFunc : () => null` results in `if (validationFunc)` always evaluating to true and validation always on, even if `validationFunc` is not passed in. So, I think it's default value should be `undefined` to signify that it's not passed to the component.

src/components/data-management/ProjectDetails.jsx
aerlaut: Hmm.. not really. I guess we can force implementors to return error string from `validationFunc`. This would enforce the function to provide an error message during validation handling too.

aerlaut:
> I just checked staging and I made a video of a tiny usability issue -- after I click on the edit button to change the name of the project, after I click on the part of the name that I want to start typing characters from, the "Technology" window pops up, which doesn't look and feel very right. The user should be able to click on and highlight any part of that text without that window showing up
> 
>  Screen.Recording.2021-04-08.at.16.07.54.mov

Hmm... yea. This behaviour is actually triggered when the user selects on a project without samples. There is already a button to add samples (on the right), so having this behaviour is confusing. I'll remove it,

ivababukova (approved):
lgtm

PR #197, opened 2021-04-08
Authorization header

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/seclocalure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-744
UI - https://github.com/biomage-ltd/ui/pull/197
API - https://github.com/biomage-ltd/api/pull/89
Inframock - https://github.com/biomage-ltd/inframock/pull/39

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,6 @@
     "babel-plugin-dynamic-import-node": "^2.3.3",
     "babel-plugin-import": "^1.9.0",
     "babel-plugin-lodash": "^3.3.4",
-    "better-fetch": "^1.1.2",
     "cross-env": "^7.0.3",
     "cssnano": "^4.1.10",
     "dotenv": "^8.2.0",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
     "babel-plugin-dynamic-import-node": "^2.3.3",
     "babel-plugin-import": "^1.9.0",
     "babel-plugin-lodash": "^3.3.4",
+    "better-fetch": "^1.1.2",
     "cross-env": "^7.0.3",
     "cssnano": "^4.1.10",
     "dotenv": "^8.2.0",
--- a/src/components/FeedbackButton.jsx
+++ b/src/components/FeedbackButton.jsx
@@ -47,6 +47,7 @@ const FeedbackButton = () => {
     try {
       const r = await fetch(atob(HOOK_URL), {
         method: 'POST',
+        headers: { Authorization: 'Bearer admin' },
         body: JSON.stringify(feedbackData),
       });
 
--- a/src/components/FeedbackButton.jsx
+++ b/src/components/FeedbackButton.jsx
@@ -47,7 +47,6 @@ const FeedbackButton = () => {
     try {
       const r = await fetch(atob(HOOK_URL), {
         method: 'POST',
-        headers: { Authorization: 'Bearer admin' },
         body: JSON.stringify(feedbackData),
       });
 
--- a/src/components/FeedbackButton.jsx
+++ b/src/components/FeedbackButton.jsx
@@ -6,7 +6,6 @@ import { useDispatch } from 'react-redux';
 import { CommentOutlined, DownOutlined } from '@ant-design/icons';
 import messages from './notification/messages';
 import pushNotificationMessage from '../redux/actions/notifications';
-import fetchAPI from '../utils/fetchAPI';
 
 const { TextArea } = Input;
 
@@ -46,7 +45,7 @@ const FeedbackButton = () => {
     };
 
     try {
-      const r = await fetchAPI(atob(HOOK_URL), {
+      const r = await fetch(atob(HOOK_URL), {
         method: 'POST',
         body: JSON.stringify(feedbackData),
       });
--- a/src/components/FeedbackButton.jsx
+++ b/src/components/FeedbackButton.jsx
@@ -6,6 +6,7 @@ import { useDispatch } from 'react-redux';
 import { CommentOutlined, DownOutlined } from '@ant-design/icons';
 import messages from './notification/messages';
 import pushNotificationMessage from '../redux/actions/notifications';
+import fetchAPI from '../utils/fetchAPI';
 
 const { TextArea } = Input;
 
@@ -45,7 +46,7 @@ const FeedbackButton = () => {
     };
 
     try {
-      const r = await fetch(atob(HOOK_URL), {
+      const r = await fetchAPI(atob(HOOK_URL), {
         method: 'POST',
         body: JSON.stringify(feedbackData),
       });
--- a/src/redux/actions/cellSets/loadCellSets.js
+++ b/src/redux/actions/cellSets/loadCellSets.js
@@ -18,7 +18,9 @@ const loadCellSets = (experimentId) => async (dispatch, getState) => {
   }
 
   try {
-    const response = await fetch(`${getApiEndpoint()}/v1/experiments/${experimentId}/cellSets`);
+    const response = await fetch(`${getApiEndpoint()}/v1/experiments/${experimentId}/cellSets`, {
+      headers: { Authorization: 'Bearer admin' },
+    });
     const json = await response.json();
     dispatch({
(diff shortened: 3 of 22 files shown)


aerlaut (commented):
So the changes here are just adding auth headers to every request from the UI right?

kafkasl (dismissed):
I think we shouldn't add the headers manually to each request. We will have to update many places if some change is required and it's easy to forget to add them. Instead we should have a middleware that just adds the headers to all requests. I haven't done that in react/redux but it should be something like this https://gist.github.com/benpickles/b465db90ca11adc9ea00

kafkasl:
@StefanBabukov probably a fetch wrapper like this is easier https://swizec.com/blog/a-tiny-es6-fetch-wrapper-that-makes-your-life-easier. I'm not a js expert, the goal is to have a single place with headers, etc... that is used by all requests.

StefanBabukov:
@kafkasl Maybe better-fetch is the better option seems much more cleaner.

kafkasl:
@StefanBabukov another option that I've just been told about is https://stackoverflow.com/questions/43051291/attach-authorization-header-for-all-axios-requests

I don't have a strong opinion, maybe someone more knowledgeable in JS can give you better advice (although they are now busy, not sure if you are working on this right now or you can wait)

StefanBabukov:
@kafkasl I changed it, I used better-fetch. Seemed a great solution. What does everyone else think?

package.json
```diff
@@ -50,6 +50,7 @@
     "babel-plugin-dynamic-import-node": "^2.3.3",
     "babel-plugin-import": "^1.9.0",
     "babel-plugin-lodash": "^3.3.4",
+    "better-fetch": "^1.1.2",
```
StefanBabukov: This module allows to add default headers to each api request.

xverges:
@StefanBabukov I do not know if I have a problem with `better-fetch` or with importing it as `fetch`: the end result is that, before @kafkasl 's request it was easy to forget to add a header, and now it is easy to forget to add the correct import and call plain `fetch` (without headers) instead of `better-fetch`'s implementation (with our desired headers).

I think that I would go for having our own wrapper, because 
* it would make more obvious what headers we are adding and when we are adding them
* it could also wrap the call to `getApiEndpoint()` that we have all over the place

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -1,4 +1,6 @@
+import fetch from 'better-fetch';
```
ivababukova: it looks like it is not used in here, so you can delete the import

ivababukova: nevermind, saw it!

kafkasl:
@StefanBabukov I tried adding it only to store.js but it's not picked up globally. If it can be defined in a single place would be ideal. 

The other thing I've been playing with is [this](https://gist.github.com/benpickles/b465db90ca11adc9ea00) which is great because you define it only once (the middleware). However, that requires that the fetch calls are made differently which could replace the ` getApiEndpoint()` call that xavier mentions.

However, @xverges I think that having middleware / better-fetch point is to hide the header details not make them explicit. They should always be added and that's it. Developers should not worry about which headers are being added (as they are the same for all calls)

xverges:
Selecting the approach, beaware that not all api calls are happening from actions.

@kafkasl 
What I tried to convey with

> it would make more obvious what headers we are adding and when we are adding them

is that, with `better-help`, if I need to know what headers are being set, I will need to go to its docs, learn that they are set with `setDefaultHeaders`, search in our code, and find the call setting the headers. With a home-made wrapper, I'll look into the home-made wrapper code, and I'll see the headers we are adding.

StefanBabukov:
Okay, so what approach do you think I should go for, I started doing the one @xverges proposed - writing a custom wrapper around fetch and using it instead of the normal fetch

package.json
xverges: Not needed anymore. You can also revert the changes to `package-lock.json`

src/utils/fetchAPI.js
```diff
@@ -0,0 +1,21 @@
...
+  const headers = params.headers ? params.headers : {};
+  const parameters = {
+    ...params,
+    headers: {
+      ...headers,
+      Authorization: 'Bearer admin',
+    },
+  };
+  let url = path;
+  if (!path.includes(getApiEndpoint())) {
```
xverges: IMO, we should not do this check. The parameter `path` should be a path, and not a full url, and remove the `getApiEndpoint()` calls that `fetchAPI` callers are doing now.

src/utils/fetchAPI.js
```diff
@@ -0,0 +1,21 @@
+import getApiEndpoint from './apiEndpoint';
+
+const fetchAPI = async (path, params = {}) => {
+  const headers = params.headers ? params.headers : {};
+  const parameters = {
+    ...params,
+    headers: {
+      ...headers,
+      Authorization: 'Bearer admin',
```
xverges: Our extra headers needs to be used both here and in our socket requests. I would define them in a third file, and import it from there.

src/utils/sendWork.js
```diff
@@ -22,6 +23,9 @@ const sendWork = async (experimentId, timeout, body, requestProps = {}) => {
     uuid: requestUuid,
     socketId: io.id,
     experimentId,
+    extraHeaders: {
+      Authorization: 'Bearer admin',
```
xverges: See previous comment about getting the header by importing it from another file.

src/components/FeedbackButton.jsx
```diff
@@ -45,7 +46,7 @@ const FeedbackButton = () => {
     };
 
     try {
-      const r = await fetch(atob(HOOK_URL), {
```
xverges: Here we want a regular fetch, since we are not calling the API.

xverges (changes requested):
Thanks!
Let's take advantage of this PR to remove all those calls `getApiEndpoint()` + some other requests.

src/utils/fetchAPI.js
StefanBabukov: Sometimes the url has the endpoint included - see getFromApiExpectOk.js file, maybe that can be changed too

xverges: You are absolutely right. And that brings the point that we missed that `getFromApiExpectOK` should sometimes call plain `fetch()` and others `fetchAPI()`, depending on the param it gets.

src/utils/getFromApiExpectOK.js
```diff
@@ -1,8 +1,13 @@
 import CustomError from './customError';
+import fetchAPI from './fetchAPI';
 
 const getFromApiExpectOK = async (url) => {
-  const response = await fetch(url);
-
+  let response = null;
+  if (url.includes('experiments')) {
```
StefanBabukov: I guess check the URL for this and we know its for our api? Hope its adequate

xverges: I would go for checking if it is a path and not and url instead of checking for `experiments`, since pretty soon we will want other places, like `data-management`.

src/utils/fetchAPI.js
```diff
@@ -0,0 +1,18 @@
+import getApiEndpoint from './apiEndpoint';
+import authorizationHeader from './authorizationHeader';
+
+const fetchAPI = async (path, params = {}) => {
+  const headers = params.headers ? params.headers : {};
+  const parameters = {
+    ...params,
+    headers: {
+      ...headers,
+      Authorization: authorizationHeader.Authorization,
```
xverges: isn't `...authorizationHeader,` enough? (also for `sendWork`)

xverges (changes requested):
Almost there :-)

xverges (approved):

PR #222, opened 2021-04-27
Minor initialisation fixes

xverges:
# Background
#### Link to issue 

N/A - As a result of using a broken local development environment that failed to load cellSets, I got several loud errors. This PR prevents them from throwing.
 
https://ui-xavier-ui222.scp-staging.biomage.net/

# Changes
### Code changes

* Set the error flag instead of the loading when fetching cellSets or samples returns an error status.
* Set better guards for Continuous Embedding plot and Violin plot

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


aerlaut (dismissed):
Changes look good, I've tested it in local. In non-`development` environment, these changes will enable the platform to load without the presence of `cellSets`, as `ContentWrapper` is bypassed. However, in production / staging, I believe there is a guarantee that `cellSets` will be available as data-ingest / pipeline is run?

kafkasl (dismissed):
why would we want to hide errors? this broken local env is temporary as we merge the cellsets changes. I'm worried that hiding these errors will make it harder to debug in the future. Shouldn't the cellsets be always present? if they are not we need to fix the env, not hide errors. Or am I missing something?

xverges:
> why would we want to hide errors?

`cellSets`, specifically their `properties` and `hierarchy` attributes, are loaded asynchronously, so we cannot blindly assume that they are already loaded. The broken environment brought us to this situation, but could have got here by means of a network glitch.

kafkasl (commented):
ok, your comment makes sense

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
```diff
@@ -184,7 +184,7 @@ const generateSpec = (config, plotData) => {
 };
 
 const filterCells = (cellSets, selectedCellSet) => {
-  let newCellSets = cellSets.hierarchy.find(
+  let newCellSets = cellSets.hierarchy?.find(
```
marcellp: This is probably not the place to do this, as there is no guarantee that `filterCells` and `generateData` will re-run once the cell sets become available.

Are you sure this is a problem at all? Can you tell me how you managed to produce the bug? I checked `CategoricalEmbeddingPlot` and it seems that there is a check to ensure this plot does not load unless cell sets are available (see [here](https://github.com/biomage-ltd/ui/blob/master/src/components/plots/CategoricalEmbeddingPlot.jsx#L47-L51)).

xverges: I was looking there also, until I realised that the issue came from https://github.com/biomage-ltd/ui/blob/57d9e8df13e877690e49f9ac06b4598651efb1aa/src/components/data-processing/ConfigureEmbedding/ConfigureEmbedding.jsx#L336

xverges:
The core reason for the failures was that the status of the fetch was not being checked, so `cellSets.error` was not set. I hav undone some of the previous changes. Now looking at the guards in `useEffect()` for the violin plot and the continuous embedding plot, because in this situation they get put the browser into a crazy CPU load.

ivababukova (approved):

PR #227, opened 2021-04-28
Data in Redux gets outdated after pipeline reruns

marcellp:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-869

https://ui-marcellp-ui227.scp-staging.biomage.net/

N/A

N/A

# Changes
### Code changes
Created a new action that, when dispatched, resets all experiment-related stores to their initial state.

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
image


src/redux/reducers/genes/index.js
```diff
@@ -20,35 +22,30 @@ const genesReducer = (state = initialState, action) => {
...
       return genesExpressionLoaded(state, action);
     }
-
     case GENES_EXPRESSION_ERROR: {
       return genesExpressionError(state, action);
     }
-
+    case EXPERIMENT_SETTINGS_PIPELINE_START: {
+      return initialState;
+    }
```
ivababukova, line 48: why in the genes reducer though?

ivababukova: ah never mind, got it

src/redux/reducers/cellInfo/index.js
```diff
@@ -9,15 +11,15 @@ const cellInfoReducer = (state = initialState, action) => {
...
       return cellInfoFocus(state, action);
     }
-
     case CELL_INFO_UNFOCUS: {
       return cellInfoUnfocus(state, action);
     }
-
+    case EXPERIMENT_SETTINGS_PIPELINE_START: {
+      return initialState;
+    }
```
ivababukova, line 22: I understand your approach here, but I have a concern that in the future if we write another part of the redux store, the developer who writes it might forget about adding these lines to invalidate the store. Do you agree that this can be a concern and if so do you have any ideas of what we can do about it?

marcellp: Yes, good point. The problem is that you cannot just reset /everything/ to its initial state. If that was the case, I would have just done it in the root reducer and be done with it, but the Redux store also includes things like the layout of the pages, experiment details, authentication credentials, etc. You cannot automatically check for the presence of this because whether you need to reset a certain store (or sometimes, only part of a store) is dependent on context.

ivababukova (approved):

PR #236, opened 2021-05-03
Set max value for min #UMIS in cell

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-889

https://ui-martinfosco-ui236.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/components/data-processing/CellSizeDistribution/CalculationConfig.jsx
+++ b/src/components/data-processing/CellSizeDistribution/CalculationConfig.jsx
@@ -12,12 +12,14 @@ import {
 } from '@ant-design/icons';
 import SliderWithInput from '../../SliderWithInput';
 
+const MIN_CELL_SIZE_PLACEHOLDER = 10800;
+
 const CellSizeDistributionConfig = (props) => {
   const {
     config, disabled, updateSettings, highestUmi,
   } = props;
 
-  const withinRange = (cellSize) => Math.max(Math.min(cellSize, highestUmi ?? 17000), 0);
+  const withinRange = (cellSize) => Math.max(Math.min(cellSize, highestUmi), 0);
 
   return (
     <>
@@ -35,10 +37,10 @@ const CellSizeDistributionConfig = (props) => {
             onPressEnter={(e) => {
               updateSettings({ minCellSize: withinRange(e.target.value) });
             }}
-            placeholder={10800}
+            placeholder={MIN_CELL_SIZE_PLACEHOLDER}
             step={100}
             disabled={disabled}
-            max={highestUmi ?? 17000}
+            max={highestUmi}
             min={0}
           />
         </Space>
--- a/src/components/data-processing/CellSizeDistribution/CalculationConfig.jsx
+++ b/src/components/data-processing/CellSizeDistribution/CalculationConfig.jsx
@@ -14,7 +14,7 @@ import SliderWithInput from '../../SliderWithInput';
 
 const CellSizeDistributionConfig = (props) => {
   const {
-    config, disabled, updateSettings,
+    config, disabled, updateSettings, highestUmi,
   } = props;
 
   return (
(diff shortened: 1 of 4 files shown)


src/components/data-processing/CellSizeDistribution/CellSizeDistribution.jsx
```diff
@@ -94,6 +97,16 @@ const CellSizeDistribution = (props) => {
     (state) => state.componentConfig[plots[selectedPlot].plotUuid]?.plotData,
   );
 
+  const histogramPlotData = useSelector(
+    (state) => state.componentConfig[plots.histogram.plotUuid]?.plotData,
+  );
+
+  const highestUmiRef = useRef(0);
```
xverges: I am unfamiliar with `useRef`. Is there a reason that makes it a better choice than `useState()` for this value?

src/components/plots/CellSizeDistributionHistogram.jsx
```diff
@@ -46,11 +46,13 @@ CellSizeDistributionHistogram.propTypes = {
     PropTypes.bool,
     PropTypes.object,
   ]),
+  highestUmi: PropTypes.number,
 };
 
 CellSizeDistributionHistogram.defaultProps = {
   plotData: null,
   actions: true,
+  highestUmi: 17000,
```
xverges: Shouldn't we make this value and the one in the `InputNumber.placeholder` equal?

src/components/data-processing/CellSizeDistribution/CellSizeDistribution.jsx
cosa65: they are used for different things: useRef is a way to persist the value inside across components re-renders, so keep it across the full lifetime of the component. Since we know that this value doesn't need to be recalculated whenever we re-render I thought it would be a valid use of useRef.

useState is for when we want to have a value that will affect what gets rendered in the component, but in this case it isn't causing re renders in this component but rather its children (we are passing it as props).

src/components/plots/CellSizeDistributionHistogram.jsx
cosa65: They are both being used but for different things, in the `InputNumber` we are setting the `minCellSize`, that value is the one the user can control.

The highestUmi is a value that is calculated by finding the maximum `u` out of all the data points received for the histogram and we use this value to decide the x axis extent of the plot, before this it was hardcoded at 17000 for some reason (which is part of what we noticed with oliver yesterday).

cosa65: We could perhaps cap the inputNumber placeholder at `max(highestUmi)`, is that what you meant?

src/components/data-processing/CellSizeDistribution/CellSizeDistribution.jsx
xverges: Thanks! Very useful!

src/components/plots/CellSizeDistributionHistogram.jsx
xverges: I mean that in one place we set the default to 17000 and in another to 10800, and it puzzles me

xverges: My comments were dumb. @cosa65 , very sorry for wasting your time. I was mistaking the highest value of the range (17000/`highestUmi`) with the default value suggested (10800).

Can we define one constant for 17000 and one for 10800 so that the future me is less unlikely to make this mistake again?

src/components/data-processing/CellSizeDistribution/CalculationConfig.jsx
```diff
@@ -17,6 +17,8 @@ const CellSizeDistributionConfig = (props) => {
     config, disabled, updateSettings, highestUmi,
   } = props;
 
+  const withinRange = (cellSize) => Math.max(Math.min(cellSize, highestUmi ?? 17000), 0);
```
xverges: Thanks! I was not aware of the `??` operator!
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

cosa65: Its very useful to avoid ifs!

src/components/plots/CellSizeDistributionHistogram.jsx
cosa65: constants make a lot of sense in this, you are right

xverges (approved):

ogibson (approved):

PR #238, opened 2021-05-04
fix single quote

alexvpickering:
`&apos;` doesn't seem to work inside quotes. Viewed on chrome:

image


xverges (approved):
Interesting...

PR #286, opened 2021-05-28
new project modal state set as closed to avoid fluttering in & out

kafkasl:
Setting up the new project modal as closed from the beginning. Otherwise, it pops up briefly and then it's auto-closed when the projects are loaded which looks bad.


ivababukova (approved):

ivababukova:
Good idea, it was annoying

PR #290, opened 2021-05-31
[BIOMAGE-1026] Improved error messages 

xverges:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1026
https://ui-xavier-ui290-api140.scp-staging.biomage.net/
https://github.com/biomage-ltd/api/pull/140

**Initial goal**
Error messages get to the user as notfication pop ups and as errors stored in redux and later displayed to the end user by visible components (e.g. GeneListTool or the NewProjectModal). We need to hide from the users messages that are more likley to confuse them than to help.

**Bonus goal 1**
Centralize in a single file error/progress messages that were in several files.

**Bonus goal 2**
Partially centralize the error management of fetch operations made from `redux` action

**Bonus goal 3**

`await` on dispatch of `redux` actions from other `redux` actions.

----

I understand that this is quite a boring review :-(

Some hints to move faster understanding the changes.

* You'll see over and over this line `import { isServerError, throwWithEndUserMessage } from '../../../utils/fetchErrors';`, so maybe it is a [good file to start the review with](https://github.com/biomage-ltd/ui/pull/290/files#diff-e8edb7fe0d9a9de07c17d7517e0e9b84e66da8fa753c236d10ffd9cfcefd9163R1)
* `utils/endUserMessages.js` is a fatter version of the now removed `components/notification/messages` 
* 3 `errorType.js` files contained progress/error messages.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/pages/data-management/index.jsx
```diff
@@ -125,7 +128,7 @@ const DataManagementPage = ({ route }) => {
         title='Data Management'
       />
       <LoadingModal
-        visible={projectSaving || sampleSaving}
+        visible={!!(projectSaving || sampleSaving)}
```
cosa65: I assume these double exclamations are for more readibility, but in that case wouldn't it be better to use `Boolean()`?

src/pages/data-management/index.jsx
```diff
@@ -125,7 +128,7 @@ const DataManagementPage = ({ route }) => {
...
         message={projectSaving || sampleSaving || ''}
```
cosa65: In this case it might be better to use `??` instead of `||`, (we can and are using `??` in the ui, the api doesn't support it)

src/utils/fetchErrors.js
```diff
@@ -0,0 +1,20 @@
+class ServerError extends Error {
+}
+
+const throwWithEndUserMessage = (response, json, friendlyMessage) => {
+  if (!response.ok) {
+    let messageToSend = json.message;
+    if (response.status === 500) {
+      console.error(`Status 500 fetching ${response.url}. Hidding full error from user: ${messageToSend}`);
```
cosa65: I would just write `Status 500 fetching ${response.url}. Full error: ${JSON.stringify(messageToSend)}

src/utils/fetchErrors.js
```diff
@@ -0,0 +1,20 @@
+class ServerError extends Error {
+}
+
+const throwWithEndUserMessage = (response, json, friendlyMessage) => {
+  if (!response.ok) {
```
cosa65, line 7: I would change the name of this function, it conveys the idea that `throwWithEndUserMessage` throws, but what it does first of all is check if it should throw.

I would name it something like `throwIfRequestFailed`, `checkResponse` or something else that makes it clearer it is actually doing a check before throwing

src/utils/fetchErrors.js
```diff
@@ -0,0 +1,20 @@
...
+    let messageToSend = json.message;
```
cosa65: I would call `messageToSend` something like `responseMessage` which make sit clearer where it comes from t oavoid confusions between the message sent to the api and the one received.

src/utils/fetchErrors.js
```diff
@@ -0,0 +1,20 @@
...
+      messageToSend = friendlyMessage;
+    }
+    throw new ServerError(messageToSend);
+  }
+};
+
+const isServerError = (error) => (error instanceof ServerError);
+
+export {
+  throwWithEndUserMessage,
```
cosa65: Very nitpicky so feel free to ignore: `throwWithUserMessage` is less wordy

src/utils/endUserMessages.js
```diff
@@ -0,0 +1,24 @@
+export default {
+  connectionError: 'Could not connect to the server. Check your internet connection and refresh the page.',
```
cosa65: I think for constants we have lately been using full uppercase like CONNECTION_ERROR

src/utils/endUserMessages.js
```diff
@@ -0,0 +1,24 @@
+export default {
```
cosa65, line 1: A lot of these messages were changed, you should go over them with @vickymorrison and check that the new ones are ok.

cosa65 (changes requested):

src/pages/data-management/index.jsx
xverges: Nice. I forced the boolean type becaiuse there was a warning about the prop having the wrong type.

src/utils/fetchErrors.js
xverges: I have removed the `JSON.stringify()` from your proposal, since `message` should be a plain string. I have left it as
```
      console.error(`Status 500 fetching ${response.url}. Error in response: ${message}`);
```

src/utils/fetchErrors.js
xverges: I have gone with plain `message`. Is it ok?

xverges:
👋  @vickymorrison 

This is what user that are not signed in will see until we have our own sign in page. Is the wording ok?
Screenshot 2021-06-03 at 08 23 27

Error/progress messages have been moved to a single file: https://github.com/biomage-ltd/ui/pull/290/files#diff-b1858551f03aa3558df0c46dcc039ecf748b51cc1efb306800e8779034c186cdR2

Most significant changes/questions:
* hide non-muggle messages coming from the server and replace them with those in the file
* removed some granularity in messages related to saving failures (i.e. '_Error saving experiment_' and '_Error saving project_' and '_Error saving..._' have been collapsed into a generic '_The server returned an error while saving your data_'). Do we need to recover the previous granularity?
* Does this talk about _the server_ make any sense at all? Would it be more appropriate to talk about _We had an error..._ or _We could not..._

vickymorrison:
I suggest a slight change to the message wording: "You need to sign in. Click the button on the top right."
About your comments/questions:
1. Sounds good.
2. The generic save error may be OK. The only reason we would want to give the user more details about the error is if it affects what they do to resolve the problem. If all they can do is wait / refresh / check their connection and then try again then I think a generic message is fine. Marcell may have thoughts here though.
3. It makes sense to me but I can't speak for other users! Let's leave it as is for now and we can change the messages if users report that they are confused.

marcellp (dismissed):
Thanks for asking for feedback on this. Here's my two cents:

* The message should be 'You need to sign in first. Click Sign in on the top right to begin."
* We should avoid referring to "the server" in user facing messages as this is useless information for the user and is technical for no reason. Let's stick with something like "(We) couldn't save your data."

src/utils/fetchErrors.js
cosa65: sounds good to me

src/utils/endUserMessages.js
```diff
@@ -3,18 +3,18 @@ export default {
...
+  ERROR_FETCHING_CELL_SETS: 'We couldn\'t get the list of cell sets.',
+  ERROR_FETCHING_EXPERIMENTS: 'We couldn\'t get the list of experiments.',
+  ERROR_FETCHING_PLOT_CONFIG: 'We couldn\'t get the plot settings.',
+  ERROR_FETCHING_PROCESSING: 'We couldn\'t get the filter settings.',
+  ERROR_FETCHING_PROJECTS: 'We couldn\'t get the list of projects.',
+  ERROR_FETCHING_SAMPLES: 'We couln\'t get your samples.',
+  ERROR_FETCHING_STATUS: 'We couldn\'t get the status of pending jobs.',
+  ERROR_NOT_SIGNED_IN: 'You need to sign in first. Click Sign in on the top right to begin.',
+  ERROR_SAVING: 'We couldn\'t save your data.',
+  ERROR_STARTING_PIPLELINE: 'We couldn\'t get our server to start working on your data.',
```
cosa65: Based on the changes suggested I think this could be "We couldn\'t start working on your data.", other than that looks good

cosa65 (approved):

PR #293, opened 2021-05-31
User profile change details

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-770
https://ui-grasp6-ui293.scp-staging.biomage.net/settings/profile

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/pages/settings/profile/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+        await Auth.changePassword(user, oldPassword, newPassword)
+          .then((response) => pushNotificationMessage('success', messages.detailsUpdated, 3))
+          .catch((error) => { setOldPasswordError("Doesn't match old password"); });
+      }
+    }
+    currentUser();
+    setChanges(initialState);
+  };
+
+  if (user) {
```
kafkasl, line 82: add comment, the user might not be loaded already

kafkasl:
old & new password can have the same value when changing them (not sure if it's ok).

vickymorrison:
I feel that there should be a 'cancel' button as well as the 'save changes' button. This should take the user back to whatever page they were viewing before they navigated to the user settings page. Currently, it's not clear how to go back without making any changes.

src/components/notification/messages.js
```diff
@@ -6,4 +6,5 @@ export default {
   feedbackSuccessful: 'Your feedback has been successfully sent. Thank you!',
   feedbackError: 'We couldn\'t process your feedback. Please try submitting again.',
   signInError: 'We couldn\'t sign you in. Please try again.',
+  detailsUpdated: 'Account details updated successfully',
```
xverges: This file has been deleted 😮 
You'll need to create a new message in https://github.com/biomage-ltd/ui/blob/develop/src/utils/endUserMessages.js

src/pages/settings/profile/index.jsx
```diff
@@ -0,0 +1,159 @@
...
+      const decimal = /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[^a-zA-Z0-9])(?!.*\s).{8,}$/;
+
+      if (confirmNewPassword !== newPassword) {
+        setNewPasswordError("Passwords don't match.");
+      } else if (!newPassword.match(decimal)) {
+        setNewPasswordError('Password should include at least 8 characters, a number, special character, uppercase letter, lowercase letter.');
+      } else {
+        await Auth.changePassword(user, oldPassword, newPassword)
+          .then((response) => pushNotificationMessage('success', endUserMessages.DETAILS_UPDATED, 3))
+          .catch((error) => { setOldPasswordError("Doesn't match old password"); });
```
xverges: Are we sure that an error always means that, or could we be misleading users?

src/utils/endUserMessages.js
```diff
@@ -21,4 +21,6 @@ export default {
   NEW_CLUSTER_CREATED: 'Created a new cluster. Check the `Data Management` window.',
   SAVING_SAMPLE: 'Saving sample...',
   SAVING_PROJECT: 'Saving project...',
+  DETAILS_UPDATED: 'Account details updated successfully.',
```
xverges: If any other change is requested, would you mind moving this where it belongs alphabetically? BTW, I would change that id to `ACCOUNT_DETAILS_UPDATED`.

kafkasl (changes requested):
Hello, I noticed that if the user changes the email to something invalid that will effectively lock out the user from his account. If the new email inserted is invalid the user won't be able to request a password to the new email. I think that the email needs to be validated like during the sign up or not be able to change that at all. Probably we should discuss this in the standup.

src/pages/settings/profile/index.jsx
StefanBabukov: The 2 errors you would get are - `InvalidParameterException` when password format is invalid - less than 6 characters etc. basically wrong password as well. The other one is `InvalidPasswordException`. The rest of the errors thrown by this are here https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ChangePassword.html. To what I see we won't see the other errors as the user is already logged in, but still should I filter the errors?

xverges: I think that it better to show an ugly error than mislead the user. If AWS is broken and throws a 500, I think that we cannot tell the user that he mistyped the password.

StefanBabukov: Ok, I'm printing the error message then - sometimes it could be ugly indeed
image
image
Maybe it's too ugly though and should print something custom instead (saying invalid **username** or password when only password is typed in - not correct error message)

src/pages/settings/profile/index.jsx
```diff
@@ -0,0 +1,170 @@
...
+                  />
+                </Form.Item>
+                <Form.Item
+                  label='Email address:'
+                  validateStatus={emailError ? 'error' : 'success'}
+                  help={emailError ? 'Invalid email address format' : ''}
+                >
+                  <Input
+                    type='email'
+                    disabled
```
robioson, line 112: Add a comment explaining why this is disabled currently

src/pages/settings/profile/index.jsx
```diff
@@ -0,0 +1,170 @@
...
+    if (name || email) {
+      setEmailError(false);
+      await Auth.updateUserAttributes(user, changedUserAttributes)
+        .then((response) => pushNotificationMessage('success', endUserMessages.ACCOUNT_DETAILS_UPDATED, 3))
+        .catch((e) => setEmailError(true));
+    }
+    if (oldPassword || newPassword || confirmNewPassword) {
+      setOldPasswordError(false);
+      setNewPasswordError(false);
+      const decimal = /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[^a-zA-Z0-9])(?!.*\s).{8,}$/;
```
robioson: This regex should be renamed. And probably also add a comment reminding people that it should be kept up to date with the Cognito validation policy.

robioson (commented):
Looks great, minor documentation suggestions

robioson (approved):
Thanks for the changes!

kafkasl (approved):

PR #296, opened 2021-05-31
[BIOMAGE-963] - Add species metadata input

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-963

https://ui-agi-ui296.scp-staging.biomage.net/

N.A.

- Add dropdown as input to species metadata

# Changes
### Code changes

- Remove `DATA_MISSING` status as it is not used anymore

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/components/data-management/ProjectDetails.jsx
+++ b/src/components/data-management/ProjectDetails.jsx
@@ -348,7 +348,7 @@ const ProjectDetails = ({ width, height }) => {
       const isSpeciesEmpty = (uuid) => metadataKey === 'species' && !samples[uuid].species;
       const isMetadataEmpty = (uuid) => metadataKey !== 'species'
         && (!samples[uuid].metadata[metadataKey]
-          || samples[uuid].metadata[metadataKey] === defaultNA);
+          || samples[uuid].metadata[metadataKey] === DEFAULT_NA);
 
       return isMetadataEmpty(sampleUuid) || isSpeciesEmpty(sampleUuid);
     };
@@ -380,7 +380,7 @@ const ProjectDetails = ({ width, height }) => {
           <MetadataEditor
             onReplaceEmpty={(value) => setCells(value, key, 'REPLACE_EMPTY')}
             onReplaceAll={(value) => setCells(value, key, 'REPLACE_ALL')}
-            onClearAll={() => setCells(defaultNA, key, 'CLEAR_ALL')}
+            onClearAll={() => setCells(DEFAULT_NA, key, 'CLEAR_ALL')}
             massEdit
           >
             <Input />
--- a/src/components/data-management/ProjectDetails.jsx
+++ b/src/components/data-management/ProjectDetails.jsx
@@ -1,13 +1,12 @@
 import React, { useState, useEffect, useRef } from 'react';
 import {
-  Table, Typography, Space, Tooltip, PageHeader, Button, Input, Progress, Row, Col,
+  Table, Typography, Space, Tooltip, Button, Input, Progress, Row, Col,
 } from 'antd';
 import { useRouter } from 'next/router';
 import { useSelector, useDispatch } from 'react-redux';
 import {
-  UploadOutlined, EditOutlined,
+  UploadOutlined,
 } from '@ant-design/icons';
-import _ from 'lodash';
 import PropTypes from 'prop-types';
 import useSWR from 'swr';
 import moment from 'moment';
@@ -256,19 +255,6 @@ const ProjectDetails = ({ width, height }) => {
         </div>
       );
     }
-
-    if (status === UploadStatus.DATA_MISSING) {
-      return (
-        <div style={{ whiteSpace: 'nowrap', height: '35px', minWidth: '90px' }}>
-          <Space>
-            <Text type='danger'>Data missing</Text>
-            <MetadataEditor size='small' shape='link' icon={<EditOutlined />}>
-              {_.find(tableColumns, { dataIndex: columnId }).fillInBy}
-            </MetadataEditor>
-          </Space>
-        </div>
-      );
-    }
   };
 
   const renderEditableFieldCell = (
@@ -295,7 +281,7 @@ const ProjectDetails = ({ width, height }) => {
       <EditableField
         deleteEnabled
         value={text}
-        onAfterSubmit={() => dispatch(updateSample(record.uuid, { name }))}
+        onAfterSubmit={(name) => dispatch(updateSample(record.uuid, { name }))}
         onDelete={() => dispatch(deleteSamples(record.uuid))}
         validationFunc={(name) => validateInputs(name, validationChecks, validationParams).isValid}
       />
@@ -329,7 +315,6 @@ const ProjectDetails = ({ width, height }) => {
           </Space>
         </MetadataPopover>
       ),
-      fillInBy: <Input />,
       width: 200,
     };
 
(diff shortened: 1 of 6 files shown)


src/components/data-management/SpeciesSelector.jsx
```diff
@@ -3,20 +3,38 @@ import PropTypes from 'prop-types';
...
+    return {
+      value: organismId,
+      label: !organismId ? <></> : createLabel(organism.scientific_name, organism.display_name),
+    };
+  };
+
   return (
     <Select
-      onChange={(value, option) => onChange(option.displayName)}
+      value={labeledValue(value)}
```
ivababukova: in here, if `value` is `null`, as it is set by default in default props, won't this line of code fail?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -256,19 +255,6 @@ const ProjectDetails = ({ width, height }) => {
         </div>
       );
     }
-
-    if (status === UploadStatus.DATA_MISSING) {
```
ivababukova, line 260: why do we remove the `DATA_MISSING` status?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -341,6 +326,46 @@ const ProjectDetails = ({ width, height }) => {
     dispatch(deleteMetadataTrack(name, activeProjectUuid));
   };
 
+  const updateSpeciesOrMetadata = (value, metadataKey) => {
+    const updateObject = metadataKey === 'species' ? { species: value } : { metadata: { [metadataKey]: value } };
```
ivababukova, line 337: what if `value` is `null`? Can `value` be `null`?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -341,6 +326,46 @@ const ProjectDetails = ({ width, height }) => {
...
+    return updateObject;
+  };
+
+  const replaceEmptyCells = (value, metadataKey) => {
+    const update = updateSpeciesOrMetadata(value, metadataKey);
```
ivababukova: the name `update` is confusing -- I would change it to a name that shows that the variable holds the name of the thing that will be updated. For example, I think `typeToUpdate` or `updateType` or something similar will be better

ivababukova: or `sampleType`? Because of line 345

src/components/data-management/ProjectDetails.jsx
```diff
@@ -341,6 +326,46 @@ const ProjectDetails = ({ width, height }) => {
...
+
+    const updatingSpecies = (sampleUuid) => metadataKey === 'species' && !samples[sampleUuid].species;
+    const updatingMetadata = (sampleUuid) => metadataKey !== 'species'
+      && (!samples[sampleUuid].metadata[metadataKey]
+        || samples[sampleUuid].metadata[metadataKey] === defaultNA);
+
```
ivababukova: why do we have that here? I thought that function `updateSpeciesOrMetadata` is taking care of this

src/components/data-management/ProjectDetails.jsx
```diff
@@ -357,37 +382,15 @@ const ProjectDetails = ({ width, height }) => {
...
-            onClearAll={() => {
-              activeProject.samples.forEach(
-                (sampleUuid) => dispatch(
-                  updateSample(sampleUuid, { metadata: { [key]: defaultNA } }),
-                ),
-              );
-            }}
+            onReplaceEmpty={(value) => replaceEmptyCells(value, key)}
+            onReplaceAll={(value) => replaceAllCells(value, key)}
+            onClearAll={() => clearAllCells(defaultNA, key)}
```
ivababukova: is there an easy way for us instead of calling `updateSpeciesOrMetadata` from within each of those functions, to call it once and then pass the result to them directly?

src/components/data-management/SpeciesSelector.jsx
aerlaut: No it doesn't. `organism` will be set to `undefined`, while `label` is set to an empty node. In the UI, this shows up as empty in the dropdown.

src/components/data-management/ProjectDetails.jsx
ivababukova: or, is there an easy, non leaky-abstraction way for us to use the same function for `onReplaceEmpty `, `onReplaceAll ` and for `onClearAll ` because they look almost the same?

src/components/data-management/SpeciesSelector.jsx
ivababukova: I see. It seems to me a better practice to avoid using `null` as a default value, even if the code works with null, and the reason is that if we use null, then for situations when the value is null without us setting it to null by default, it will be harder to capture and debug. It also feels more messy. As a default value, I would use an empty string for example, or a string `"empty"` or something similar. What do you think?

src/components/data-management/SpeciesSelector.jsx
```diff
@@ -3,20 +3,38 @@ import PropTypes from 'prop-types';
...
 const SpeciesSelector = (props) => {
-  const { data, onChange } = props;
+  const { data, value, onChange } = props;
 
-  if (!data) {
+  if (!data || data.length === 0) {
     return <Skeleton.Input style={{ width: 300 }} size='small' />;
   }
 
+  const createLabel = (scientificName, displayName) => (
```
ivababukova: This is a bit messy and unnecessarily coupled.  `createLabel` returns a React component, `labeledValue` is coupled with `createLabel` and calls it if there is no `organismId`, `organismId` is actually equal to that `value` that you have in default props. I would suggest:

1. Separation of concerns: Either make `labeledValue` to be only about returning organismId and rename it to `getOrganismId`. Or combine functions `createLabel` and `labeledValue` into 1 function if these 2 things have to be called together each time and rename that function to show better what it is actually doing.
2. if `value` in props is actually organismId, then why don't you rename it to `organismId`? `value` is super ambigious. Also, if it is in props,  why do you need to pass it in the `labeledValue` function?

src/components/data-management/SpeciesSelector.jsx
ivababukova: I can definitely see that the need for this usage of `createLabel` and `labeledValue` is coming from this line and also from line 75 below. I would definitely consider splitting these things up and making them less coupled. Maybe the original way of how label was defined on lines 57-64 that are deleted was better and all you need is a small function called "getLabelName" for line 36, without the return statement of `label: !organismId ? <></> : createLabel(organism.scientific_name, organism.display_name),` is all you need

ivababukova (changes requested):
There is a bit of easy to do and small cleanup we can do for SpeciesSelector and ProjectDetails, see my comments

src/components/data-management/SpeciesSelector.jsx
aerlaut: Using an empty string is cleaner. It's also easier to check with `!defaultValue`. I'll change the default value to ''.

src/components/data-management/ProjectDetails.jsx
aerlaut: It wasn't used any more with the data. However, it might be related to this https://biomage.atlassian.net/browse/BIOMAGE-1049

src/components/data-management/ProjectDetails.jsx
aerlaut: no, the value can't be `null`. It'll be the value of the item that's selected in the dropdown.

aerlaut: Inputting values into the dropdown to search does not update the `species` value in redux, so it doesn't corrupt the store.

src/components/data-management/ProjectDetails.jsx
aerlaut: I change it to `updateObject`

src/components/data-management/ProjectDetails.jsx
ivababukova: by the fact that it might be related to 1049, do you mean that we might want to keep this status?

aerlaut: After further investigation, keeping this piece of code does not show any controls for uploading individual files as required in BIOMAGE-1049, so I'll remove this code.

src/components/data-management/SpeciesSelector.jsx
aerlaut: Cleaned up by removing `labelInValue`, which requires the value to be of object `{ value: string, label : ReactNode}`, so now we can just pass `value` instead of having to construct the entire React node.

src/components/data-management/SpeciesSelector.jsx
aerlaut: Cleaned up by removing `labelInValue`

src/components/data-management/ProjectDetails.jsx
aerlaut: Refactored into `massUpdateCells`

src/components/data-management/ProjectDetails.jsx
```diff
@@ -342,6 +333,35 @@ const ProjectDetails = ({ width, height }) => {
...
+    if (!MASS_EDIT_ACTIONS.includes(actionType)) return;
+    const updateObject = createUpdateObject(value, metadataKey);
+
+    const canUpdateCell = (sampleUuid, action) => {
+      if (action !== 'REPLACE_EMPTY') return true;
+
+      const isSpeciesEmpty = (uuid) => metadataKey === 'species' && !samples[uuid].species;
+      const isMetadataEmpty = (uuid) => metadataKey !== 'species'
+        && (!samples[uuid].metadata[metadataKey]
+          || samples[uuid].metadata[metadataKey] === defaultNA);
```
ivababukova: my editor says that this `defaultNA ` is not defined

src/components/data-management/ProjectDetails.jsx
```diff
@@ -358,37 +378,15 @@ const ProjectDetails = ({ width, height }) => {
...
-            onClearAll={() => {
-              activeProject.samples.forEach(
-                (sampleUuid) => dispatch(
-                  updateSample(sampleUuid, { metadata: { [key]: DEFAULT_NA } }),
-                ),
-              );
-            }}
+            onReplaceEmpty={(value) => setCells(value, key, 'REPLACE_EMPTY')}
+            onReplaceAll={(value) => setCells(value, key, 'REPLACE_ALL')}
+            onClearAll={() => setCells(defaultNA, key, 'CLEAR_ALL')}
```
ivababukova: here too -- `defaultNA` not defined

ivababukova (approved):

ivababukova:
Miles better than last time I looked at it, happy to get it merged

src/components/data-management/ProjectDetails.jsx
aerlaut: Changed to `DEFAULT_NA`. Updated accordingly

src/components/data-management/ProjectDetails.jsx
aerlaut: updated

PR #325, opened 2021-06-08
[BIOMAGE-1049]Implement functionality to upload files for incomplete samples

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&modal=detail&selectedIssue=BIOMAGE-1049
https://ui-stefan-ui325.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -99,7 +99,7 @@ const UploadDetailsModal = (props) => {
         }}
         style={{ width: '140px', marginBottom: '10px' }}
       >
-        Replace file
+        {bundleName ? 'Replace file' : 'Upload'}
```
robioson: You could add a comment explaining that `bundleName` will be undefined if the file hasn't been uploaded

src/utils/checkIfFileValid.js
```diff
@@ -0,0 +1,25 @@
...
+    techOptions[selectedTech].validMimeTypes
+      .includes(
+        mime.lookup(fileName),
+      )
+    || techOptions[selectedTech].validExtensionTypes
+      .includes(
+        path.extname(fileName),
+      )
+  );
+  const acceptedFilesRegexp = `(${techOptions[selectedTech].acceptedFiles.join('|')})$`;
```
robioson: I know that naively writing unions using `join('|')` can cause issues with escaping. Might not be an issue here but Ruby has `Regexp.union` for this, I wonder if we could do something similar.

src/components/data-management/ProjectDetails.jsx
```diff
@@ -551,15 +534,14 @@ const ProjectDetails = ({ width, height }) => {
     if (!uploadDetailsModalDataRef.current) {
       return;
     }
-
     const { sampleUuid, file } = uploadDetailsModalDataRef.current;
-
-    const bucketKey = `${activeProjectUuid}/${sampleUuid}/${file.name}`;
+    const name = file.name || bundleToUpload.name;
```
robioson, line 537: Similar to below review comment, a comment here would be good to explain when each path is taken.

robioson (commented):
Looks good, few minor comments

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -139,7 +140,7 @@ const UploadDetailsModal = (props) => {
         {!isSuccessModal
           && (
             <Row style={{ marginTop: '5px', marginBottom: '5px' }}>
-              The following file has failed to upload
+              The following file has failed to upload or was not uploaded
```
cosa65: I think we should change this message to depend on the status that the modal is filled with currently.
"The following file has failed to upload" for UPLOAD_ERROR
"The following file was not uploaded" for FILE_NOT_FOUND

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -127,7 +127,8 @@ const UploadDetailsModal = (props) => {
       footer={(
         <Row style={{ width: '100%', justifyContent: 'center' }}>
           <Col>
-            {isSuccessModal ? downloadButton() : retryButton()}
+            {/* render retry button only if file was tried to be uploaded */}
+            {bundleName && (isSuccessModal ? downloadButton() : retryButton())}
```
cosa65: A more strict way to check for this (as in it fails if we reach an invalid state) could be checking that the status is not FILE_NOT_FOUND instead of checking for bundleName to exist

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -150,10 +151,12 @@ const UploadDetailsModal = (props) => {
           <Col span={5}>Category</Col>
           <Col span={10}>{fileCategory}</Col>
         </Row>
-        <Row style={{ marginTop: '5px', marginBottom: '5px' }}>
-          <Col span={5}>Filename</Col>
-          <Col span={10}>{bundleName}</Col>
-        </Row>
+        {bundleName && (
```
cosa65: perhaps use FILE_NOT_FOUND instead as in previous comment?

src/redux/actions/samples/updateSampleFile.js
```diff
@@ -16,18 +17,22 @@ const updateSampleFile = (
 ) => async (dispatch, getState) => {
   const updatedAt = moment().toISOString();
   const sample = getState().samples[sampleUuid];
+  // we'll need to remove the hard-coded 10x tech type once we start
+  // supporting other types and save the chosen tech type in redux
+  const valid = checkIfFileValid(fileName, '10X Chromium');
```
cosa65: we are using `updateSampleFile` only in `processUpload` and usually not to create a new sample file but to update an existing one so it is usually not necessary to do this check, I assume this validity check was added here so that it is run when we are using `UploadDetailsModal` to update a file, could it be moved into that modal instead then?

src/utils/checkIfFileValid.js
StefanBabukov: what do you mean i didnt get it - .join just adds the character between every element in the array. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join Or you mean something else?

src/redux/actions/samples/updateSampleFile.js
```diff
@@ -16,18 +17,22 @@ const updateSampleFile = (
...
 
   try {
     // Save sample only if upload is successful or error
     if (fileDiff.upload.status === UploadStatus.UPLOADED
```
cosa65: this is not part fo your PR, but could you perhaps change this if for `[UPLOADED, UPLOAD_ERROR].includes(fileDiff.upload.status)`

src/redux/actions/samples/updateSampleFile.js
```diff
@@ -16,18 +17,22 @@ const updateSampleFile = (
...
       || fileDiff.upload.status === UploadStatus.UPLOAD_ERROR) {
       const diffObject = {
-        fileNames: sample.fileNames.add(fileName),
+        fileNames: sample.fileNames,
```
cosa65: why was this line changed? If filenames is still a set then it should not be a problem, right?

src/redux/reducers/samples/samplesFileUpdate.js
```diff
@@ -20,8 +20,7 @@ const samplesFileUpdate = (state, action) => {
 
   newFile.lastModified = lastModified;
 
-  const newFileNames = new Set([...state[sampleUuid].fileNames, fileName]);
-
+  const newFileNames = Array.from(new Set([...state[sampleUuid].fileNames, fileName]));
```
cosa65, line 23: In the initial state fileNames is a set, why are we doing this change?

src/utils/checkIfFileValid.js
robioson: What I mean is that the things you are joining are strings being turned into a regex. However the strings (file names) contain dots, which have a special meaning in regex, so technically need to be escaped. Sorry, I know it's weird to use another language as an example but you will see that `Regexp.union` safely escapes special characters:

```irb
irb(main):012:0> Regexp.union(*acceptedFiles)
=> /barcodes\.tsv|barcodes\.tsv\.gz|features\.tsv|features\.tsv\.gz|genes\.tsv|genes\.tsv\.gz|matrix\.mtx|matrix\.mtx\.gz/
irb(main):013:0> acceptedFiles.join('|')
=> "barcodes.tsv|barcodes.tsv.gz|features.tsv|features.tsv.gz|genes.tsv|genes.tsv.gz|matrix.mtx|matrix.mtx.gz"
```
Because dots mean "match any one character", my prediction is that the current logic will allow any characters in place of the dots. This is easy to test.

Also, I know you did not write this and just moved it, but I just noticed it here!

robioson: OK, I think the mime type check prevents this from ever actually being an issue but it is technically wrong.

robioson: We could use https://stackoverflow.com/a/29700268/7799676.

```js
const acceptedFileRegexp =  `(${techOptions[selectedTech].acceptedFiles.map(_.escapeRegExp).join('|')})$`
```

Seems to be correct

robioson: image
Yeah, it is actually possible to witness this on the platform (the upload fails but beforehand it is marked as valid). My fix should work I think.

src/redux/actions/samples/updateSampleFile.js
StefanBabukov: Relates to the below comment and also sample.fileNames was already updated to include the new file

src/redux/reducers/samples/samplesFileUpdate.js
StefanBabukov: The API expects an array not a set and I was getting errors.

src/redux/actions/samples/updateSampleFile.js
StefanBabukov: But we still need to provide a `valid` field here so that the analysis can be launched. Do you think it should be just valid: true here, while the check is done in the `UploadDetailsModal` and if valid is true, then call `updateSampleFile`?

src/utils/checkIfFileValid.js
StefanBabukov: Thanks for the explanation! I changed it.

src/redux/actions/samples/updateSampleFile.js
cosa65: Ohh right, so it is already added in the reducer which means this is unnecessary.

Looks good and I definitely think it is better to perform all the setup for updates in the reducer while the action just sends what it received (and/or performs api calls), however I think I would go further and remove the `fileNames` field from this diff.

cosa65: Also something else that we will need to do eventually is separate sampleFile updates from sampleFile posts to the api a little bit more, but that's for a different task.

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -37,6 +37,7 @@ const UploadDetailsModal = (props) => {
   }, [replacementFileBundle]);
 
   const isSuccessModal = status === UploadStatus.UPLOADED;
+  const isNotUploaded = status === UploadStatus.FILE_NOT_FOUND;
```
cosa65: could it be isNotUploadedModal instead? to be more coherent with the uploaded status check in line 39

src/redux/actions/samples/updateSampleFile.js
cosa65: Yes, I believe that's the pattern we are already following in FileUploadModal, and apart from this we should use the same method for both modals (like the utils function you are using, perhaps compare it with your implementation and check if they can be integrated).

cosa65 (approved):

PR #346, opened 2021-06-21
[BIOMAGE-1105] Tweak Data Processing step dropdown

robioson:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1105
#### Link to staging deployment URL 
https://ui-james-ui346.scp-staging.biomage.net/
#### Anything else the reviewers should know about the changes here
I tried for a long time to find a way to fix the `Select` to the width of the longest option, but couldn't, so settled on hardcoding it. If there is reasonable way to do this just let me know!

image

# Changes
### Code changes
- Set `Select` width to width of longest option
- Added numbers to options
- Added comment describing when each option variant is used

# Definition of DONE
Your changes will be ready for merging after each of the steps below have been completed:

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging

To set up easy local testing with inframock, follow the instructions here: https://github.com/biomage-ltd/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/biomage-ltd/biomage-utils

### Documentation updates
Is all relevant documentation updated to reflect the proposed changes in this PR?

- [ ] Relevant Github READMEs updated
- [ ] Relevant wiki pages created/updated

### Approvers
- [ ] Approved by a member of the core engineering team
- [ ] (UX changes) Approved by vickymorrison (this is her username, tag her if you need approval)

### Just before merging:
- [ ] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional
- [ ] Photo of a cute animal attached to this PR

PR #348, opened 2021-06-22
Fix dropdown width

robioson:
This is a quick followup to #346 - I missed that the dropdown is not quite wide enough and shows an ellipsis when the longest option is selected.

Before:
image

After:
image

### DIFF
--- a/src/pages/experiments/[experimentId]/data-processing/index.jsx
+++ b/src/pages/experiments/[experimentId]/data-processing/index.jsx
@@ -381,12 +381,13 @@ const DataProcessingPage = ({ experimentId, experimentData, route }) => {
         <Col style={{ flex: 1, display: 'flex', justifyContent: 'flex-start', alignItems: 'center' }}>
           <Row style={{ display: 'flex' }} gutter={16}>
             <Col>
+              {/* Should be just wide enough that no ellipsis appears */}
               <Select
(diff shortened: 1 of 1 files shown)


src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -381,12 +381,13 @@ const DataProcessingPage = ({ experimentId, experimentData, route }) => {
...
           <Row style={{ display: 'flex' }} gutter={16}>
             <Col>
+              {/* Should be just wide enough that no ellipsis appears */}
               <Select
                 value={stepIdx}
                 onChange={(idx) => {
                   changeStepId(idx);
                 }}
-                style={{ fontWeight: 'bold', width: 270 }}
+                style={{ fontWeight: 'bold', width: 290 }}
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx:390]: is there a way to make the width fit with the current length of the text without having to hardcode a particular width?

Comment (robioson) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:390]: It's annoying but not that I could find - I mentioned this in #346.

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:390]: ok, fair enough

ivababukova (approved):

PR #350, opened 2021-06-23
[BIOMAGE-1089] Fix launch button disabled on analysis modal

robioson:
After editing both name and description, and then saving name.

# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1089

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/components/data-management/AnalysisModal.jsx
+++ b/src/components/data-management/AnalysisModal.jsx
@@ -28,7 +28,7 @@ const NewExperimentModal = (props) => {
   const dispatch = useDispatch();
 
   const [experimentsList, setExperimentsList] = useState([]);
-  const [numFieldsEditing, setNumFieldsEditing] = useState(1);
+  const [numFieldsEditing, setNumFieldsEditing] = useState(0);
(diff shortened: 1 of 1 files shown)


aerlaut (approved):
I see that the solution works. However, why is `numFieldsEditing`'s default value set to 1?

robioson:
Oh, right, this was really weird. When the modal is loaded for the first time, both of the `EditableField` components are initialised. This calls their `onEditing` handler. If it started at 0, the counter should therefore be at -2 at this point. Actually, it only gets reduced to -1, my guess is this is a race condition in redux. The proper fix would be to make `EditableField` not call its handler on initialisation. From what I gathered from Martin, this requires a lot of boilerplate to do. Also it might break other usages of the component although I doubt this was the intended behaviour.

A more intuitive fix is using `Math.max(0, ...)`. I will push this.

PR #355, opened 2021-06-23
Introduce immerjs to experiment settings 

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1075

https://ui-martinfosco-u355-a166.scp-staging.biomage.net/

https://github.com/biomage-ltd/api/pull/166

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-processing/CellSizeDistribution/CellSizeDistribution.jsx
```diff
@@ -93,8 +93,7 @@ const CellSizeDistribution = (props) => {
     (state) => state.componentConfig[plots[selectedPlot].plotUuid]?.config,
   );
   const expConfig = useSelector(
-    (state) => state.experimentSettings.processing[filterName][sampleId]?.filterSettings
-      || state.experimentSettings.processing[filterName].filterSettings,
+    (state) => state.experimentSettings.processing[filterName][sampleId].filterSettings,
```
StefanBabukov: What if it's not multisample, would we still have this path?

src/redux/actions/experimentSettings/updateSampleFilterSettings.js
```diff
@@ -0,0 +1,10 @@
+import { EXPERIMENT_SETTINGS_SAMPLE_FILTER_UPDATE } from '../../actionTypes/experimentSettings';
+
+const updateSampleFilterSettings = (step, sampleId, diff) => (dispatch) => {
```
StefanBabukov: Didnt updateProcessingSettings take care of this or?

cosa65:
In this PR I also removed the usage of default filter settings when there is not a setting for one sample in particular because it simplified the code considerably (and might help in debugging in the future).

src/components/data-processing/CellSizeDistribution/CellSizeDistribution.jsx
cosa65: In unisample we still use the sample's id, for example you can check and see that (even before this PR) we never modified the default settings, not even for unisample.

This `||` was added because for some of the steps, sample's filter settings weren't defined in the beginning and we only had the default values (stored in `state.experimentSettings.processing[filterName][sampleId].filterSettings`), not because of unisample.

Still using these default values hasn't been useful for a while now and makes code unnecessarily complex so I decided to remove it.

src/redux/actions/experimentSettings/updateSampleFilterSettings.js
cosa65: Yes, I guess actually some thing I could do is change `updateProcessingSettings` to be blocked to only handling the steps that aren't separated samples and see if I have to add more separation there.

cosa65: And Ill also add comments.

src/redux/reducers/experimentSettings/copyFilterSettingsToAllSamples.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-param-reassign */
+import produce from 'immer';
+
+const copyFilterSettingsToAllSamples = produce((draft, action) => {
+  const { step, sampleId, allSampleIds } = action.payload;
+
+  const settingsToCopy = draft.processing[step][sampleId];
```
StefanBabukov: I think you need to use `original` here ?

StefanBabukov:
There is a bug introduced by your changes to reproduce it:
1. Create a project and launch an analysis wait until pipeline is finished
2. Create another one and do the same
3. Go back to data-management and launch the first experiment
You'll get this 
image
Switching back and forth between analyses breaks the platform

To what I see when you load an analysis initially its all fine, when you go back and launch another one, the processing config gets populated with the previous samples. So if you have classifier: { randomSample:{...}} changing the project loads the samples from the previous one so its still  classifier: { randomSample:{...}}

src/redux/actions/experimentSettings/updateSampleFilterSettings.js
cosa65: Ended up just changing the name, if you can think of a better one please let me know. I want it to say that it is only for configure embedding, data integration or meta, basically all the updates that are for something that doesn't work on a per-sample basis.

src/redux/actions/experiments/updateExperiment.js
```diff
@@ -10,8 +9,6 @@ const updateExperiment = (
   experiment,
 ) => async (dispatch) => {
   try {
-    dispatch(saveExperiment(experimentId));
```
cosa65, line 13: Moved to a separate action to be able to update redux without calling the api

cosa65:
@StefanBabukov reply to https://github.com/biomage-ltd/ui/pull/355#issuecomment-868502696

The problem was that the processingConfig was never reloaded when the experimentId changed so we never actually got the newly selected experiment's processingConfig.

As to why this didn't cause a crash before and it does with my changes, I would say it is probably because before we had default values to avoid crashes and now we no longer do.
 
One good way to test this in develop is following your steps for my crash, on the last experiment (the one that crashed in my branch), check cellSizeDistribution minCellSize, its going to be 1080, which is the default value and not the real value that is stored in dynamodb, the analogous thing happens with any other setting.

To fix this I changed the updateExperimentInfo reducer, this action sets the current experimentId and that seems to me like  a good moment to determine when the experiment being referenced was changed and when the processingConfig needs to be reloaded (I set meta.loading = true in that place).

src/redux/actions/experimentSettings/processingConfig/copyFilterSettingsToAllSamples.js
```diff
@@ -0,0 +1,18 @@
...
+
+  const allSampleIds = cellSets.hierarchy?.find(
+    (rootNode) => (rootNode.key === 'sample'),
+  )?.children.map(
+    (child) => child.key,
+  );
+
+  dispatch({
+    type: EXPERIMENT_SETTINGS_COPY_SETTINGS_TO_ALL_SAMPLES,
+    payload: { step, sampleId, allSampleIds },
```
cosa65: pass samplSettings instead of sampleId

cosa65: By receiving it in the action creator directly

cosa65: same with allSampleIds

src/redux/actions/experimentSettings/processingConfig/updateNonSampleFilterSettings.js
```diff
@@ -0,0 +1,15 @@
+import { EXPERIMENT_SETTINGS_NON_SAMPLE_FILTER_UPDATE } from '../../../actionTypes/experimentSettings';
+
+const updateNonSampleFilterSettings = (configName, configChange) => (dispatch) => {
```
cosa65: Merge with updateSampleFilterSettings

src/redux/reducers/experimentSettings/processingConfig/updateNonSampleFilterSettings.js
```diff
@@ -0,0 +1,25 @@
...
+const updateNonSampleFilterSettings = produce((draft, action) => {
+  const { step, configChange } = action.payload;
+
+  if (!step) throw new Error(`Invalid step parameter received: ${step}`);
+
+  const originalProcessingConfig = original(draft.processing)[step] ?? {};
+
+  const newConfig = _.cloneDeep(originalProcessingConfig);
+
+  mergeObjectWithArrays(
```
cosa65, line 17: try just assigning configchange to draft

cosa65:
Use recent( instead of original(

src/redux/reducers/experimentSettings/processingConfig/loadedProcessingConfig.js
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable no-param-reassign */
+import produce, { current } from 'immer';
+import _ from 'lodash';
+
+import initialState from '../initialState';
+
+// Will be removed once the default values are stored in the pipeline
+const withAutomaticSettings = (processingConfig) => {
```
cosa65, line 8: This was added originally in a different PR than this one because we add `defaultFilterSettings` property in the ui on receiving processingConfig, it should probably be moved to the api or (even better) to gem2s when generating the processingConfig, would like to hear opinions on this and if we agree we can create a ticket to move it

src/redux/actions/experimentSettings/backendStatus/loadBackendStatus.js
```diff
@@ -63,11 +63,12 @@ const loadBackendStatus = (experimentId) => async (dispatch) => {
     // having to run the pipelines
     if (process.env.NODE_ENV === 'development'
       && (status.gem2s.status === pipelineStatusValues.NOT_CREATED
-        || status.qc.status === pipelineStatusValues.NOT_CREATED)) {
+        || status.pipeline.status === pipelineStatusValues.NOT_CREATED)) {
```
marcellp: did we change this back from `qc` to `pipeline`?

src/redux/actions/experimentSettings/processingConfig/loadedProcessingConfig.js
```diff
@@ -0,0 +1,25 @@
...
+      data: processingConfig,
+      fromGem2s,
+    },
+  });
+
+  if (fromGem2s) {
+    Object.keys(processingConfig).forEach((step) => {
+      dispatch(saveProcessingSettings(experimentId, step));
+    });
+  }
```
marcellp, line 22: why is this needed?

src/redux/actions/experimentSettings/processingConfig/updateFilterSettings.js
```diff
@@ -0,0 +1,24 @@
...
+    });
+  } else if (settingsWithoutSampleId.includes(step)) {
+    dispatch({
+      type: EXPERIMENT_SETTINGS_NON_SAMPLE_FILTER_UPDATE,
+      payload: { step, configChange: diff },
+    });
+  }
+};
+
+export default updateFilterSettings;
```
marcellp: Very happy with this interface!

src/redux/reducers/experimentSettings/processingConfig/loadedProcessingConfig.js
```diff
@@ -0,0 +1,49 @@
...
+const loadedProcessingConfig = produce((draft, action) => {
+  const { data, fromGem2s } = action.payload;
+
+  const originalProcessing = current(draft.processing);
+
+  let dataToSet = _.cloneDeep(data);
+
+  dataToSet.meta = {
+    ...originalProcessing.meta,
+    ...data.meta,
```
marcellp, line 36: This shouldn't be correct. It assumes that `originalProcessing.meta` is a simple object with no nested object properties that we'd need to merge. What if we end up with something like `stepsDone` here that needs special processing? Might be worth refactoring these two as a `_.merge` or something similar.

src/utils/experimentUpdatesHandler.js
```diff
@@ -94,24 +81,12 @@ const uglyTemporalFixedProcessingConfig = (processingConfig, filterName) => {
...
-      const currentObject = fixedProcessingConfig[key];
-      const settingsKey = Object.keys(currentObject).find((current) => currentObject[current].auto);
-      if (settingsKey) {
-        dispatch(updateProcessingSettings(
-          key, { defaultFilterSettings: fixedProcessingConfig[key][settingsKey] },
-        ));
-        dispatch(saveProcessingSettings(experimentId, key));
-      }
-    });
+    dispatch(loadedProcessingConfig(experimentId, fixedProcessingConfig, true));
```
marcellp, line 89: What happened here? I see `loadProcessingConfig` is much smaller, how was the actual improvement made?

marcellp (commented):
No major issues as far as I can see, I have a few comments however.

src/redux/actions/experimentSettings/backendStatus/loadBackendStatus.js
cosa65: It was never changed initially I think, there is a ticket to move the last parts of qc name to pipeline in the api, but right now this was causing the platform to crash in localhost, so I changed it back to `pipeline`

src/redux/actions/experimentSettings/processingConfig/loadedProcessingConfig.js
cosa65: This was added originally in a different PR than this one (different code but same thing, it sent updates to the api) when adding the `defaultFilterSettings` property in the ui on receiving processingConfig (it was in the `experimentUpdatesHandler`).

I think it should probably be moved to the api or (even better) to gem2s when generating the processingConfig, but I didn't want to delay the ticket more by doing something unrelated to it, so I just refactored it into the changes of my ticket.

If you agree with me on mving it somewhere else we could creat ea ticket for that.

src/utils/experimentUpdatesHandler.js
cosa65: what improvement?

marcellp (approved):
lgtm

PR #357, opened 2021-06-24
[BIOMAGE-687] - Update tooltips

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-687

https://ui-agi-ui357.scp-staging.biomage.net/

N.A.

- Add more information to `Hide` button tooltip.
- Added cellsets information to embedding
- Change `Tools` to `Genes` in the exploration tab
- Modify `Tooltip`s in `Genes`

# Changes
### Code changes

- `antd` has to be updated to enable `showSorterTooltip` which is used to change the tooltip of the header columns of the genes table in Genes List and Differential Expression. Upgrading to the newer version causes `Menu` on the sidebar to render twice, creating visible menu items and a duplicate non-visible components for ARIA support. That is why in the tests, the number of menu items rendered should be halved, because each item is rendered twice. 

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/__test__/components/ContentWrapper.test.jsx
```diff
@@ -71,7 +71,12 @@ describe('ContentWrapper', () => {
     expect(sider.length).toEqual(1);
 
     const menus = wrapper.find(Menu).children().find(Item);
-    expect(menus.length).toEqual(4);
+
+    // Menu item renders twice to support HOC usage (?)
+    // https://ant.design/components/menu/#Why-Menu-children-node-will-render-twice
+    const visibleMenuLength = menus.length / 2;
```
aerlaut, line 77: After upgrading to new antd, Menu.Item is rendered twice : 1 visible, and 1 invisible for ARIA support. Therefore, to get the real number of visible items, this value is divided by 2. [more info](https://ant.design/components/menu/#Why-Menu-children-node-will-render-twice)

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -68,7 +68,7 @@ const ExplorationViewPage = ({
         <HeatmapPlot experimentId={experimentId} width={width} height={height} />
       ),
     },
-    Tools: {
+    Genes: {
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx:71]: These changes are done to replace `Tools` with `Genes` in the data exploration display.

src/redux/actions/cellSets/createCellSet.js
```diff
@@ -18,6 +18,7 @@ const createCellSet = (experimentId, name, color, cellIds) => async (dispatch, g
     name,
     color,
     cellIds: new Set([...cellIds].map((id) => parseInt(id, 10))),
+    type: 'cellSets',
```
aerlaut, line 21: This field can be used to differntiate `cellSets` properties from `metadataCategorical` properties. All cell sets created from QC have this property. Newly created `cellSets` did not have this, so I added it to facilitate easier filtering of `cellSets` to be used in the embedding.

.github/workflows/ci-develop.yaml
```diff
@@ -30,6 +30,7 @@ jobs:
     - id: install
       name: Install dependencies
       run: |-
+        git config --global url."https://".insteadOf ssh://
```
aerlaut, line 33: Added this to change `npm` to use  `https` instead of `git` which requires an SSH key, which is not set up in the CI machine. This solves error 128 during npm ci builds.

aerlaut: Reference: https://www.programmersought.com/article/94636987109/

Dockerfile
```diff
@@ -12,7 +12,7 @@ COPY package.json package-lock.json /app/
 
 # install dependencies
 RUN apk add --no-cache git
-RUN npm ci --only=production
+RUN git config --global url."https://".insteadOf ssh:// && npm ci --only=production
```
aerlaut, line 15: Also add `ssh` to `https` in Dockerfile so the same process is used while building the container. Otherwise, the build fails because `ssh` is not available in the image.

aerlaut:
@vickymorrison can you take a look at the UI changes? The staging link is up

vickymorrison:
A couple of comments:

1. The addition of the cell set to the embedding crosshair hover is excellent but it takes up a lot of space, especially when the cell set names are long (see screenshot). Two suggestions to deal with this would be: 
- make the text size smaller - the text size currently looks like it is bigger than the gene list text size so I think it can be reduced without negatively impacting UX.
- shorten long cell set names e.g. in this example rather than "some really interesting cells" reduce to "some really int..." for cell sets greater than a specified length.
Screenshot 2021-06-25 at 09 44 48

2. With the update of the tooltips on the column titles (Dispersion, Pct1 and Pct2) there is now some inconsistency in that all the other column titles in the Gene/DE tables (Gene, Avg Log2FC and adj p-value) say "click to sort ascending". For consistency should these other tooltips be removed? Or will users then not know that they can sort the tables? My feeling is that the up and down arrows are self-explanatory and show that the tables can be sorted. And I have observed that often the hover on the 'Gene' title gets in the way of the gene search bar. Therefore I am leaning towards removing these "click to sort ascending" tooltips - what do you think?

aerlaut:
1. I've made the text smaller.
2. I agree with you, the tooltips are not necessary, since the arrows are there and it is quite common to see this functionality. I've removed them.

Are the changes ok @vickymorrison ?

vickymorrison:
Can you make the Cell id text the same size as the rest of the text in the hover? It looks weird that it is bigger. And from the user's perspective the Cell set info is more useful than Cell id so it doesn't make sense that the Cell id info is bigger. Otherwise, looks good.

src/components/data-exploration/CellInfo.jsx
```diff
@@ -35,23 +36,26 @@ const CellInfo = (props) => {
       }}
     >
       {cellInfo.cellName ? (
-        <div>
-          Cell id:&nbsp;
-          {cellInfo.cellName}
+        <div style={{ fontSize: '0.75rem' }}>
```
StefanBabukov: Suggestion:  maybe add these in a .css file and put those divs in a class. Or maybe that's not that important if its not going to be used in many places?

aerlaut: I dont' think it'll be used in any other place, but I'll refactor it and put it in a variable.

src/components/data-exploration/CellInfo.jsx
```diff
@@ -35,23 +36,26 @@ const CellInfo = (props) => {
...
       {cellInfo.expression !== undefined ? (
-        <div>
+        <div style={{ fontSize: '0.75rem' }}>
           Expression Level:&nbsp;
           {parseFloat(cellInfo.expression.toFixed(3))}
         </div>
       ) : <></>}
+      {cellInfo.cellSets?.length > 0 ? cellInfo.cellSets.map((cellSetName) => (
+        <div style={{ fontSize: '0.75rem' }}>
+          {`${_.truncate(cellSetName)}`}
```
StefanBabukov: do you need the `${}` here, why not just {_.truncate(cellSetName)}?

aerlaut: oh yea, I don't need that. Removed.

vickymorrison (approved):

StefanBabukov (approved):
LGTM just small comments

PR #385, opened 2021-07-07
[BIOMAGE-1189] added matomo tracking

kafkasl:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1189

https://ui-matomo-tracking.scp-staging.biomage.net/

You can run this PR locally or staging and see the events in biomage.matomo.cloud 
Ask Pol / Agi for the log in account.

# Changes
### Code changes

Added a nextjs tracker which handles all links implicitly. In addition I added a custom event to track when a user clicks the laun experiment button.

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

image

### DIFF
--- a/src/pages/_app.jsx
+++ b/src/pages/_app.jsx
@@ -9,6 +9,7 @@ import Amplify, { Storage, withSSRContext } from 'aws-amplify';
 import _ from 'lodash';
 import AWS from 'aws-sdk';
 import { Credentials } from '@aws-amplify/core';
+import { initTracking } from '../utils/tracking';
 import ContentWrapper from '../components/ContentWrapper';
 import NotFoundPage from './404';
 import UnauthorizedPage from './401';
@@ -16,6 +17,8 @@ import Error from './_error';
 import { wrapper } from '../redux/store';
 import '../../assets/self-styles.less';
 import '../../assets/nprogress.css';
+import { ssrGetCurrentEnvironment } from '../utils/environment';
+
 import CustomError from '../utils/customError';
 
 const mockCredentialsForInframock = () => {
@@ -61,6 +64,10 @@ const WrappedApp = ({ Component, pageProps }) => {
   const [amplifyConfigured, setAmplifyConfigured] = useState(!amplifyConfig);
 
   const environment = useSelector((state) => state.networkResources.environment);
+
+  useEffect(() => {
+    initTracking(environment);
+  }, []);
 
   useEffect(() => {
     if (amplifyConfig) {
@@ -183,6 +190,7 @@ WrappedApp.getInitialProps = async ({ Component, ctx }) => {
   try {
     const { Auth } = withSSRContext(ctx);
     Auth.configure(results.amplifyConfig.Auth);
+
     if (query?.experimentId) {
       const { default: getExperimentInfo } = require('../utils/ssr/getExperimentInfo');
       const experimentInfo = await getExperimentInfo(ctx, store, Auth);
--- a/src/utils/tracking.js
+++ b/src/utils/tracking.js
@@ -13,7 +13,7 @@ const trackingInfo = {
     siteId: 1,
   },
   [Env.STAGING]: {
-    enabled: true,
+    enabled: true, // TODO disable for staging before merging
     siteId: 2,
   },
   [Env.DEVELOPMENT]: {
@@ -31,7 +31,6 @@ const initTracking = async (environment) => {
   env = environment;
   const { siteId, enabled } = getTrackingDetails(env);
   if (enabled === false) {
-    console.log(`tracking init disabled in env ${env}`);
     return;
   }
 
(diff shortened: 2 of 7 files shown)


src/utils/tracking.js
```diff
@@ -0,0 +1,59 @@
+import { init, push } from '@socialgouv/matomo-next';
+import { ssrGetCurrentEnvironment } from './environment';
+
+const env = ssrGetCurrentEnvironment();
+const MATOMO_URL = 'https://biomage.matomo.cloud/';
+// Site IDs:
+const SITE_IDS = {
+  production: 1,
+  staging: 2,
+  development: 3,
```
ivababukova: what are these site ids?

src/utils/tracking.js
```diff
@@ -0,0 +1,59 @@
...
+  if (trackingEnabled === true) {
+    push(['resetUserId']);
+    // we also force a new visit to be created for the pageviews after logout
+    push(['appendToTrackingUrl', 'new_visit=1']);
+  } else {
+    console.log(`setting tracking ID disabled in env ${env}`);
+  }
+};
+// the push method takes as parameter t
+const trackAnalysisLaunched = () => {
```
ivababukova, line 56: I see that we have a special method to track when a user launches analysis. Why do we need a method for this, but we don't need methods for other user activities?

src/utils/tracking.js
```diff
@@ -0,0 +1,70 @@
...
+    siteId: 3,
+  },
+};
+
+const hostName = window.location.hostname;
+
+const trackingEnabled = trackingInfo[hostName]?.enabled;
+const siteId = trackingInfo[hostName]?.siteId;
+
+console.log(window.location.hostname);
```
aerlaut: Don't forget to remove this

src/utils/tracking.js
```diff
@@ -0,0 +1,72 @@
...
+  },
+};
+let hostName = false;
+if (typeof window !== 'undefined') {
+  hostName = window.location.hostname;
+  console.log(window.location.hostname);
+}
+
+const trackingEnabled = trackingInfo[hostName]?.enabled && hostName !== false;
+const siteId = trackingInfo[hostName]?.siteId;
```
marcellp: I would make these into a function that returns the two values and takes `hostName`, call it `getTrackingDetails` or something.

src/utils/tracking.js
```diff
@@ -0,0 +1,72 @@
...
+  'localhost:5000': {
+    enabled: true,
+    siteId: 3,
+  },
+};
+let hostName = false;
+if (typeof window !== 'undefined') {
+  hostName = window.location.hostname;
+  console.log(window.location.hostname);
+}
```
marcellp: How about we pass the `req` object into `initTracking` and extract the hostname from there.

src/utils/tracking.js
```diff
@@ -0,0 +1,72 @@
...
+
+const trackingEnabled = trackingInfo[hostName]?.enabled && hostName !== false;
+const siteId = trackingInfo[hostName]?.siteId;
+
+const initTracking = () => {
+  if (trackingEnabled === true) {
```
marcellp: Essentially, the changes here would boil down to:

```
const initTracking = (req) => {
  const hostName = req.get('host');

  getTrackingDetails(hostName);
  
  [...]
}
```

src/pages/_app.jsx
```diff
@@ -62,6 +63,10 @@ const WrappedApp = ({ Component, pageProps }) => {
 
   const environment = useSelector((state) => state.networkResources.environment);
 
+  useEffect(() => {
+    initTracking();
+  }, []);
```
marcellp: This shouldn't be necessary here. You can call this from within `getInitialProps` and that will guarantee that it will only be called once, when the user first loads the site. You could do something like:

```
WrappedApp.getInitialProps = async ({ Component, ctx }) => {
  const {
    store, req, query, res,
  } = ctx;
  // Do nothing if not server-side
  if (!req) { return { pageProps: {} }; }
  
  initTracking(req);

  [...]
}
```

That should solve your issues along with my other comments.

marcellp (changes requested):
I heard of these bugs and I would make the changes below to fix them. There may be other solutions available, but this one seems like the quickest and best one to me.

src/utils/tracking.js
robioson: Also, can't you rather do `Object.is(window, undefined)`?

src/utils/tracking.js
kafkasl: those are defined on biomage.matomo.cloud. I added a comment

kafkasl (commented):
> I heard of these bugs and I would make the changes below to fix them. There may be other solutions available, but this one seems like the quickest and best one to me.

@marcellp changes have been addressed

src/utils/tracking.js
kafkasl: we aren't tracking any specific interactions so far, only visits to the different parts of the website. I added this as an example of how to track specific events (and we agreed we do want to track when the user clicks the launch analysis). In the future we'll have to add more events (like downloading plots & tables) or other stuff but there's another ticket for that and currently our primary goal is to get tracking data and returning users.

marcellp (approved):

PR #387, opened 2021-07-08
[BIOMAGE-1155] [BIOMAGE-1164] Fix volcano plot issues

robioson:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1155
https://biomage.atlassian.net/browse/BIOMAGE-1164
https://ui-stealthy-bronze-angelfish.scp-staging.biomage.net/
The y-axis spec referred to invalid config properties which prevented it from showing up. The reviewer needs to check that y-axis title and values match up (in other words that the values that now display on the y-axis look correct).

While testing the fix Oliver and I noticed some more problems with the volcano plot so I fixed them here too.

When fixing the second part of the ticket ("All" not being selectable) I also fixed a validation issue in the differential expression form.
# Changes
### Code changes
- Got y-axis to show up
- Removed insignificant and unknown gene statuses
- Renamed statuses to be presentable
- Fixed range sliders (set min and max to default values)
- Implemented option for auto range

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/__test__/redux/reducers/componentConfig/__snapshots__/updateConfig.test.js.snap
```diff
@@ -102,7 +102,7 @@ Object {
         "enabled": true,
         "position": "top-right",
       },
-      "logFoldChangeDomain": 20,
+      "logFoldChangeDomain": 0,
```
cosa65, line 105: why was this changed?

cosa65:
Currently the log2 fold change value can be set as high as we want, to the point that it breaks the plot, as you can see here
Screen Shot 2021-07-09 at 13 01 38

I am thinking that perhaps we could add some kind of maximum value to it based on the values that we receive?
There are some examples of implementations of this in the doublet filter's probability threshold (although in this case I guess the maximum value should be calculated on receiving the result from the worker).

Also, do we want to allow negative values? Because right now we are allowing them, for example I just wrote -100 and got this:
Screen Shot 2021-07-09 at 13 04 15

In doublet filter the component `SliderWithInput` is handling these updates in a way that helps avoid having these kind of problems (a value being set when it is over/below the allowed range). In my opinion I would look into either using this component (in case having a slider is useful here) or refactoring it so that you can use only the input with the max/min values working correctly.

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -76,8 +75,8 @@ const VolcanoPlot = ({ experimentId }) => {
       controls: [{
         name: 'volcanoDimensions',
         props: {
-          xMax: Math.round(spec.xMax),
-          yMax: Math.round(spec.maxNegativeLogpValue) + 2,
+          xMax: 20,
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:78]: Why is this value now hardcoded at 20?

src/__test__/redux/reducers/componentConfig/__snapshots__/updateConfig.test.js.snap
robioson: 0 means "auto". Also, 20 was _huge_ as a default and seemed random.

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (robioson) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:78]: I explained this in the commit description; basically, `xMax` is almost the same as what auto sets, which meant the slider offered basically no adjustment with it as a maximum value. Maybe it would be better to set it here to `xMax * 10`.

cosa65 (approved):

PR #393, opened 2021-07-13
[BIOMAGE-1102] Add code coverage

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1102

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


.babelrc
```diff
@@ -15,13 +15,6 @@
     ]
   ],
   "plugins": [
-    [
```
cosa65, line 18: This would cause jest with coverage flag to conflict in a very weird way with some antd components.

It seems to be a well know issue and this change seemed to fix it. Based on https://github.com/facebook/jest/issues/6938#issuecomment-545451157

ivababukova: doesn't removing this piece of code break some other stuff as well though?

cosa65: I tried running tests without problems and going through the application locally it seems to be ok.

I'll test the staging once it is up

cosa65: Also I messages @marcellp in case he has more context on possible problems this removal might cause.

If it does cause many problems there is another "solution" which is to extract components from antd objects without object destructuring (which is what fails in the tests).

As an example, this breaks the tests:

```
import b from 'antd';
const {a} = b;
```


This doesn't 
```
import b from 'antd';
const a = b.a;
```

ivababukova: interesting, we could just as well be fine without it. Let's try it in staging and see if it is ok

cosa65: Turns out it did break staging, just had a call with marcell and he figured out a way to have this work correctly (the latest commit).

codecov.yml
```diff
@@ -0,0 +1,15 @@
+codecov:
```
kafkasl: most of the file have `.yaml` extension not `.yml` would be nice to be consistent

codecov.yml
```diff
@@ -0,0 +1,15 @@
...
+  require_ci_to_pass: yes
+
+coverage:
+  range: 80..100
```
kafkasl: I would maybe use 60-100 so that "acceptable values" start looking nice at 80

codecov.yml
```diff
@@ -0,0 +1,15 @@
...
+  round: down
+  precision: 2
+
+  status:
+    project:                   # measuring the overall project coverage
+      default:                 
+        enabled: no
```
kafkasl: I think enabling it already might be a good choice. Maybe with some leniency until people gets used to it like in example "Ease Target Coverage" https://docs.codecov.com/docs/common-recipe-list

codecov.yaml
```diff
@@ -0,0 +1,19 @@
...
+  range: 60..100
+  round: down
+  precision: 2
+
+  status:
+    project:                   # measuring the overall project coverage
+      default:                 
+        target: auto
+        # adjust accordingly based on how flaky your tests are
+        # this allows a 10% drop from the previous base commit coverage
```
kafkasl: 5%

marcellp (approved):

PR #408, opened 2021-07-23
[BIOMAGE-1137] - Fix sample order

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1137

https://ui-agi-uipr408.scp-staging.biomage.net/

UI - https://github.com/biomage-ltd/ui/pull/408
API - https://github.com/biomage-ltd/api/pull/188

- Sample can be ordered using drag and drop
- Reordering sample does not cause re-run of GEM2S
- Renaming experiment name does not cause re-run of GEM2S
- Renaming samples and modifying metadata still cause re-run.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


vickymorrison (approved):

src/components/data-management/ProjectDetails.jsx
```diff
@@ -390,33 +395,47 @@ const ProjectDetails = ({ width, height }) => {
...
     },
     {
+      index: 4,
       key: 'matrix',
       title: 'Matrix.mtx',
       dataIndex: 'matrix',
       render: (tableCellData) => renderUploadCell('matrix', tableCellData),
     },
     {
+      index: 0,
```
cosa65: should this be `index: 5`?

aerlaut: oh right! Can't believe I missed it 🤦

aerlaut: fixed

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -30,6 +32,7 @@ const loadCellSets = (experimentId) => async (dispatch, getState) => {
       payload: {
         experimentId,
         data: json.cellSets,
+        order: sampleIds,
```
cosa65, line 35: We could access this value inside the CELL_SETS_LOADED reducer from the current state, right? Instead of passing it as parameter.

src/redux/actions/pipeline/runGem2s.js
```diff
@@ -46,7 +46,8 @@ const runGem2s = (experimentId) => async (dispatch, getState) => {
       payload: {
         experimentId,
         experimentName: experiments[experimentId].name,
-        projectUuid: experiments[experimentId].projectUuid,
+        projectId,
```
cosa65, line 51: Why not keep the previou way of passing the projectId payload? We already getting name from that anyways, right?

src/redux/actions/samples/loadSamples.js
```diff
@@ -18,14 +18,18 @@ const loadSamples = (
     const response = await fetchAPI(url);
     const data = await response.json();
 
+    let samples;
+    if (!experimentId) samples = data[0].samples;
```
cosa65: The api should follow the same schema for the same request, why does it vary?

src/redux/reducers/cellSets/helpers.js
```diff
@@ -1,10 +1,12 @@
-const createHierarchyFromTree = (data) => data && data.map((rootNode) => {
+const createHierarchyFromTree = (data, order = null) => data && data.map((rootNode) => {
```
cosa65, line 2: Why is order null? Is there a case in which we will not want to preserve order?

src/utils/composeTree.js
```diff
@@ -25,4 +25,31 @@ const composeTree = (hierarchy, properties, filterType = null) => {
...
 };
 
+/**
+ * Reorders tree to conform to a specific ordering.
+ *
+ * `order` is a list containing the property keys that we want to reorder.
+ *
+ */
+const setHierarchyOrder = (inputHierarchy, order = null) => {
+  if (!order) {
```
cosa65: I would just avoid calling this function if we don't have an order instead of allowing it to be null but not doing anything in that case.

src/utils/composeTree.js
```diff
@@ -25,4 +25,31 @@ const composeTree = (hierarchy, properties, filterType = null) => {
...
+    return inputHierarchy;
+  }
+
+  return inputHierarchy
+    .filter((root) => root.children !== undefined)
```
cosa65: If you do `filter((root) => root.children)` does this achieve the same?

cosa65 (changes requested):

src/redux/actions/cellSets/loadCellSets.js
aerlaut: `state` in the reducer only gets the state of the current store i.e. `cellSets`. `sampleIds` is stored in a different store: `experimentSettings`, so it can't be accessed through the `state` in the reducer.

src/redux/actions/pipeline/runGem2s.js
aerlaut: `projetUuid` is stored as `projectId` in `experiments` table in dynamoDB. That is also the key used in the `experimentInfo` reducer. unfortunately, the way it is stored in `info` is under teh key `projectUuid`. I think this is a major problem. We have several items that means the same things but under different naming conventions : 
- samples : sampleId(s), sampleUuid(s), sample(s)
- project : projectId, projectUuid

I vaguely remember there is a ticket about this? I think it's quite a refactor involving all parts of the platform.

```
const updateExperimentInfo = produce((draft, action) => {
  const {
    experimentId,
    experimentName,
    projectId,
    sampleIds,
  } = action.payload;

  draft.info.experimentId = experimentId;
  draft.info.experimentName = experimentName;
  draft.info.projectUuid = projectId;
  draft.info.sampleIds = sampleIds;

  // Experiment id was updated so processing config requires reloading
  draft.processing.meta.loading = true;
}, initialState);
```

src/utils/composeTree.js
aerlaut: This function was left over when I was trying out different approaches to for cellSets ordering. It's not used anymore so I'll remove this,

src/utils/composeTree.js
aerlaut: Deleted together with the function

src/redux/reducers/cellSets/helpers.js
aerlaut: There are other calls to `createHierarchyFromTree` in which we might not want to specify an order. In these cases, `order` is not passed to the function. We can not set a default value, in which case `order` becomes `undefined` in those cases and will still be evaluated the same, in regards to the use in this function.

src/redux/actions/samples/loadSamples.js
aerlaut: API schema call for `projects/{projectId}/samples` : 

```
application/json:
  schema:
    type: array
    items:
      type: object
      properties:
        experimentId:
          type: string
        projectUuid:
          type: string
        samples:
          $ref: ./models/api-body-schemas/Samples.v1.yaml
```

API schema for `/experiments/{experimentId}/samples` :

```
application/json:
  schema:
    $ref: ./models/api-body-schemas/Samples.v1.yaml
```

cosa65 (approved):
Approved so it can be merged tomorrow, just let me know if another review is necessary

PR #390, opened 2021-07-12
[BIOMAGE-1183] Modify the expression embedding colours

robioson:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1183
#### Link to staging deployment URL 
https://ui-marcellp-jamesr-test.scp-staging.biomage.net/
#### Anything else the reviewers should know about the changes here

# Changes
### Code changes
- For the Vega plot in Plots and Tables, added new colour scheme option "Default" which is white-purple on light background and grey-yellow on inverted background
- For the Vitessce plot in Data Exploration, changed colour scheme to white-purple. Replaced static image with CSS gradient :rocket: 
# Definition of DONE
Your changes will be ready for merging after each of the steps below have been completed:

### Testing
- [ ] Unit tests written
- [ ] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging

To set up easy local testing with inframock, follow the instructions here: https://github.com/biomage-ltd/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/biomage-ltd/biomage-utils

### Documentation updates
Is all relevant documentation updated to reflect the proposed changes in this PR?

- [ ] Relevant Github READMEs updated
- [ ] Relevant wiki pages created/updated

### Approvers
- [ ] Approved by a member of the core engineering team
- [ ] (UX changes) Approved by vickymorrison (this is her username, tag her if you need approval)

### Just before merging:
- [ ] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional
- [ ] Photo of a cute animal attached to this PR

PR #410, opened 2021-07-23
Implement default colour scheme for heatmap and embedding

robioson:
# Background
https://ui-james-ui410.scp-staging.biomage.net/data-management
#390
#405
A new "default" colour scheme was added in the first PR listed. This was implemented for the continuous embedding plot but we forgot to implement it for the heatmap, which broke it on default settings.

# Changes
### Code changes
This PR adds back the default colour scheme, and implements it for heatmap (alongside embedding) so it doesn't break.

I also fixed the elbow plot ignoring the colour inversion setting.
# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/plots/styling/ColourbarDesign.jsx
```diff
@@ -18,6 +18,7 @@ const ColourbarDesign = (props) => {
         <p><strong>Colour Schemes</strong></p>
         <Form.Item>
           <Radio.Group onChange={(e) => onUpdate({ colour: { gradient: e.target.value } })} value={config.colour.gradient}>
+            <Radio value='default'>Default</Radio>
```
cosa65, line 21: Im thinking that this might cause problems/confusion in the future and we might run into this problem again since all the other configuration options for gradient are actual strings accepted by vega.

What do you think about perhaps having the value of this radio item be defined as a constant (`const DEFAULT_GRADIENT_SCHEME = purplered;`) or something?

Leaving it up to each spec to deal with default specifically is a little bit dangerous and it also is a little bit of repetition of code I think.

Judging by what I say it shouldn't be that big of a problem (you can even have the `config.colour.toggleInvert` value checked in here and it wouldn't be a problem, if `toggleInvert` isn't available in a plot styling config then it will just be a check against a falsy value, undefined)

robioson: The main reason I did it like this is that if a user has a configuration saved, and we change the `value` of a `Radio` so that the value saved matches none of the options, none of the radio buttons get selected, which is not exactly a problem but ugly. My second reason was that we could change what the `default` value represented in future or even per plot without adding too many options (especially if the best option is not the same for all plots, which I can see happening).

I agree that this is a potential issue since it would cause the default settings to create a plot that is invalid and doesn't show up (as you found for heatmap).

I tried what you suggested, if I understood it correctly:

```jsx
<Radio
  value={config.colour.toggleInvert
    ? 'darkgreen'
    : 'purplered'
}>Default</Radio>
```

To make this work we need to reload the component on configuration changes, which I'm not sure how to do, or if it's a good idea.

One middle of the road could be using `<Radio value='purplered'>Default</Radio>`, then changing the code in the continuous spec to switch to `'darkgreen'` when `toggleInvert` is true (I believe it is currently the only spec that currently supports both colour scheme and background invert).

Lastly, we could just abandon the 'Default' option and simply say 'Purple-Red', and drop the inverted colour scheme for inverted background. I initially only added this because we used `'purples'` as the colour scheme, which was unreadable on the black background, but `'purplered'` doesn't have this problem'

cosa65: I think it's ok to leave it with the default then. The only thing we actually need to do is make sure that all the plots that are currently using this styling configuration have the piece of code that turns `default` into a valid vega option (two of the plots in configure embedding and one in data integration).

cosa65 (approved):

PR #414, opened 2021-07-27
[BIOMAGE-1165] Make gene selection case insensitive

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1165
https://ui-grasp6-ui414.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -105,7 +106,7 @@ describe('ViolinIndex', () => {
   it('loads by default the gene with the highest dispersion, allows another to be selected, and updates the plot\'s title', async () => {
     await renderViolinIndex();
 
-    const geneSelection = rtl.screen.getAllByRole('tab')[0];
+    const geneSelection = rtl.screen.getAllByText('Gene Selection')[0];
     expect(geneSelection).toHaveTextContent('Gene Selection');
```
aerlaut: This test doesn't do anything. It queries for an element containing the text, and expects the retrieved element to have the text, so if the element is there, of course expects will be true.

If you're only looking for 1 element, you can just use `getByText('Gene Selection')`.

The replaced test checks for the order of the element. If this is not what we're testing for, instead of `toHaveTextContent`, you can just check for the existence of the element using something like `expect(geneSelection.length).toEqual(1)`

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -116,17 +117,20 @@ describe('ViolinIndex', () => {
...
+    await rtl.waitFor(() => expect(generateSpecSpy).toHaveBeenCalledTimes(3));
     await expectStringInVegaCanvas(newGeneShown, 1);
   });
 
   it('allows selection of the grouping/points, defaulting to louvain and All', async () => {
     await renderViolinIndex();
 
-    const dataSelection = rtl.screen.getAllByRole('tab')[1];
+    const dataSelection = rtl.screen.getAllByText('Select Data')[0];
     expect(dataSelection).toHaveTextContent('Select Data');
```
aerlaut: Same like the comment above about `getAllByText('Select Data')`. If you're only looking for 1 element, you can just use `getByText('Select Data')`.

The replaced test checks for the order of the element. If this is not what we're testing for, instead of `toHaveTextContent`, you can just check for the existence of the element using something like `expect(dataSelection.length).toEqual(1)`

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -166,7 +170,7 @@ describe('ViolinIndex', () => {
   it('has a Data Tranformation panel', async () => {
     await renderViolinIndex();
 
-    const tabs = rtl.screen.getAllByRole('tab');
+    const tabs = rtl.screen.getAllByText('Data Transformation');
```
aerlaut: Same here, you can try using `getByText` instead of `getAllByText`

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -256,24 +260,26 @@ describe('ViolinIndex', () => {
...
     const titleOverride = '€'; // Single character so that useUpdateThrottled does not mess with the tests
     generateSpecSpy.mockClear();
     userEvent.type(titleInput, titleOverride);
     await rtl.waitFor(() => expect(generateSpecSpy).toHaveBeenCalled());
-    await expectStringInVegaCanvas('MockGeneWithHighestDispersion', 0);
+    await expectStringInVegaCanvas('MockGeneWithHighestDispersion', 1);
     await expectStringInVegaCanvas(titleOverride, 1);
 
-    const geneSelection = tabs.find((tab) => tab.textContent === 'Gene Selection');
+    const geneSelection = rtl.screen.getAllByText('Gene Selection')[0];
```
aerlaut: Same here, try use `getByText` to get only one element?

src/components/plots/ViolinPlot.jsx
```diff
@@ -37,29 +37,46 @@ const ViolinPlot = (props) => {
     geneNamesFilter: null,
     sorter: { field: PROPERTIES[0], columnKey: PROPERTIES[0], order: 'descend' },
   };
+  const updatePlotWithChanges = (changes) => {
+    dispatch(updatePlotConfig(plotUuid, changes));
+  };
+
+  useEffect(() => {
+    // initially if there is a saved gene in the config
```
aerlaut: I think the code is clear enough. You can delete this or give a better comment

src/redux/actions/genes/loadGeneExpression.js
```diff
@@ -36,17 +36,20 @@ const loadGeneExpression = (
...
-  const displayedGenes = genes.filter((gene) => upperGenesAlreadyLoaded.has(gene.toUpperCase()));
-
+  const displayedGenes = genesAlreadyLoaded.filter(
+    (gene) => upperCaseArray(genes).includes(gene.toUpperCase()),
+  );
   if (genesToFetch.length === 0) {
+    // All genes are already loaded.
+    if (onUpdate) {
+      onUpdate({ shownGene: displayedGenes[0] });
+    }
```
aerlaut: Please move this out of this action creator.  This creates a tight coupling between action creator and updating the plot. You can return the displayed gene from the action and use trigger the update outside of this action creator.

src/redux/actions/genes/loadGeneExpression.js
```diff
@@ -66,7 +69,9 @@ const loadGeneExpression = (
     const data = await fetchCachedWork(
       experimentId, body, backendStatus.status, { timeout: 30 },
     );
-
+    if (onUpdate) {
+      onUpdate({ shownGene: Object.keys(data)[0] });
+    }
```
aerlaut: As above, please move this out of the action creator.

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-unresolved */
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:1]: What is this import for? Is there an import that can't be unresolved?

src/components/plots/styling/violin/ViolinControls.jsx
```diff
@@ -0,0 +1,68 @@
...
+} from 'antd';
+import { useSelector } from 'react-redux';
+import SelectData from 'components/plots/styling/violin/SelectData';
+import SingleGeneSelection from '../SingleGeneSelection';
+
+const { Panel } = Collapse;
+
+const ViolinControls = (props) => {
+  const { config, onUpdate, setSearchedGene } = props;
+  const cellSets = useSelector((state) => state.cellSets);
```
aerlaut: Can you call this under `index` and pass in `cellSets` as a prop? That would reduce coupling between components,

jsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+    "compilerOptions": {
+      "baseUrl": "src",
```
aerlaut, line 3: nice 👍

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -7,15 +7,16 @@ import { createStore, applyMiddleware } from 'redux';
 import thunk from 'redux-thunk';
 import _ from 'lodash';
 import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
-
+import '@testing-library/jest-dom';
```
aerlaut: I think you can remove this. Are you using any of the [custom matchers](https://github.com/testing-library/jest-dom#usage) provided by the library?

aerlaut (changes requested):
Can you please address the comments?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:1]: Yeah, as I added the `jsonconfig.js` file we can now import things from a defined relative path, but the eslint still marks this as unresolved path so thats why I do this disable.

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:1]: I see.. ok. Does that mean we have to have this `eslint-disable` in most other files too?

aerlaut:
Error showed up when accessing Violin plot because user skipped configure embedding. I asked @StefanBabukov to add a check and ask the user to configure embedding before coming back to the violin plot.

aerlaut:
@vickymorrison Stefan added a check in the violin plot if user's. This prevents the violin plot from throwing an error if the clusters have not been configured (i.e. users did skipped configure embedding). Is the wording for the error message good enough?

I was thinking :
No clustering available -> Your data have not been clustered.

Set up your clustering in the configure embedding step in Data Processing to view this plot, or select a different data. ->
Select a different grouping to use for this plot under "Select Data", or configure your clusters in the Configure Embedding step of Data Processing.

image

vickymorrison:
In reply to the comment above: My understanding is that the clustering is calculated automatically during Data Processing. So even if the user skips the Configure Embedding step (or indeed the whole Data Processing module) the clustering is still calculated automatically. Certainly, this is the ideal behaviour and as far as I know this is what is currently implemented. Is that correct? If so, then there is no use case for "no clustering available" unless something fails on our side. In fact, if there is no clustering then many parts of Cellscope will be useless including the default view in Data Exploration (which shows clustering on the embedding). All in all, I don't think there is need to account for a "no clustering" state.

aerlaut:
> In reply to the comment above: My understanding is that the clustering is calculated automatically during Data Processing. So even if the user skips the Configure Embedding step (or indeed the whole Data Processing module) the clustering is still calculated automatically. Certainly, this is the ideal behaviour and as far as I know this is what is currently implemented. Is that correct? If so, then there is no use case for "no clustering available" unless something fails on our side. In fact, if there is no clustering then many parts of Cellscope will be useless including the default view in Data Exploration (which shows clustering on the embedding). All in all, I don't think there is need to account for a "no clustering" state.

There were some changes that breaks this behaviour and caused Configure Embedding to not automatically be run - we have to have a look at this again. Even with the embedding running automatically, I think it's still good to have this check to prevent the platform from throwing error in case configure embedding failed.

StefanBabukov:
I guess this is a fix for the violin plot if this happens, maybe we should think how to fix it so it doesn't happen anyway, without cellSets most of the platform will be useless like Vicky said.

ivababukova:
@StefanBabukov I see this has been opened for a while. How about you call someone today and go through the code together? Let's get it reviewed so we can merge it soon.

aerlaut (approved):
LGTM

PR #416, opened 2021-07-28
[BIOMAGE-1236] - Download Processed Seurat Object

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1236

https://ui-agi-download-seurat.scp-staging.biomage.net/

UI - https://github.com/biomage-ltd/ui/pull/416
API - https://github.com/biomage-ltd/api/pull/194

- Download is done by preparing a signed download link that allows direct download from AWS S3.

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-management/ProjectDetails.jsx
```diff
@@ -577,22 +622,30 @@ const ProjectDetails = ({ width, height }) => {
...
+            Processed Seurat object (.rds)
+          </a>
         </Tooltip>
       </Menu.Item>
-      <Menu.Item disabled>
-        <Tooltip title='Coming soon!' placement='left'>
-          Data Processing settings
-        </Tooltip>
+      <Menu.Item
+        disabled
```
kafkasl: add coming soon tooltip here too

src/components/data-management/ProjectDetails.jsx
```diff
@@ -577,22 +622,30 @@ const ProjectDetails = ({ width, height }) => {
...
       </Menu.Item>
-      <Menu.Item disabled>
-        <Tooltip title='Coming soon!' placement='left'>
-          {/* <Tooltip title='Data Processing filters have been applied' placement='left'> */}
-          Processed Seurat object (.rds)
+      <Menu.Item
+        key='download-processed-seurat'
+        disabled={!pipelineHasRun}
+      >
+        <Tooltip title='Data Processing filters have been applied' placement='left'>
```
kafkasl: I think this tooltip is not very clear. I would probably write something like "Seurat object produced by applying data processing filters" (not very happy with that  phrasing either. Do you have any suggestion @vickymorrison ?

src/pages/data-management/index.jsx
```diff
@@ -54,6 +55,13 @@ const DataManagementPage = ({ route }) => {
   }, []);
 
   useEffect(() => {
+    if (activeProject?.experiments.length > 0) {
+      // Right now we have one experiment per project, so we can just load the experiment
+      // This has to be changed when we have more than one experiment
+      const activeExperimentId = activeProject.experiments[0];
+      dispatch(loadBackendStatus(activeExperimentId));
```
kafkasl: why is this needed?

src/redux/reducers/experiments/experimentsDataDownloaded.js
```diff
@@ -0,0 +1,20 @@
+const experimentsDataDownload = (state, action) => {
```
kafkasl: shouldn't this reducer use the new method that @cosa65 introduced to avoid the state unpacking?

kafkasl (changes requested):

src/components/data-management/ProjectDetails.jsx
aerlaut: added

src/components/data-management/ProjectDetails.jsx
aerlaut: "Data with filters applied"?

src/pages/data-management/index.jsx
aerlaut: To check whether pipeline has been run for this experiment. To do this, the status of the `pipeline` run is checked. Backend status was not loaded for this module.

src/redux/reducers/experiments/experimentsDataDownloaded.js
aerlaut: Whoops, these reducers should be removed

src/components/data-management/ProjectDetails.jsx
vickymorrison: How about "With Data Processing filters and settings applied"?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -104,6 +115,47 @@ const ProjectDetails = ({ width, height }) => {
     }
   }, [samples, activeProject]);
 
+  const downloadData = async (type) => {
+    const experimentId = activeProject?.experiments[0];
+
+    try {
+      if (!environment) throw new Error('Environment is not set');
```
kafkasl: why is this needed? the variable is not being used in the method

src/components/data-management/ProjectDetails.jsx
```diff
@@ -64,15 +71,19 @@ const ProjectDetails = ({ width, height }) => {
...
   const activeProject = useSelector((state) => state.projects[activeProjectUuid]) || false;
+  const { backendStatus } = useSelector((state) => state.experimentSettings);
+
   const [tableData, setTableData] = useState([]);
   const [tableColumns, setTableColumns] = useState([]);
   const [sortedSpeciesData, setSortedSpeciesData] = useState([]);
   const [sampleNames, setSampleNames] = useState(new Set());
   const [canLaunchAnalysis, setCanLaunchAnalysis] = useState(false);
   const [analysisModalVisible, setAnalysisModalVisible] = useState(false);
+  const [pipelineHasRun, setPipelineHasRun] = useState(true);
```
kafkasl: why is the default `true`? shouldn't we assume that the pipeline has not run?

src/components/data-management/ProjectDetails.jsx
```diff
@@ -104,6 +115,47 @@ const ProjectDetails = ({ width, height }) => {
...
+    if (
+      !loading
+      && !error
+      && status?.pipeline
+      && status.pipeline?.status === pipelineStatusValues.SUCCEEDED
+    ) {
+      setPipelineHasRun(true);
+    }
+
+    setPipelineHasRun(false);
```
kafkasl: this is being defined on a project-level. This will break once there is more than a single-experiment per project. We are already having issues where different projects interact among them so I'd be very careful to limit the visibility of state to the relevant component.

src/components/data-management/ProjectDetails.jsx
```diff
@@ -104,6 +115,47 @@ const ProjectDetails = ({ width, height }) => {
     }
   }, [samples, activeProject]);
 
+  const downloadData = async (type) => {
+    const experimentId = activeProject?.experiments[0];
```
kafkasl: I think the logic of choosing the experiment to download should not be inside this function it just does too many things inducing errors. This function should receive a type and the experimentID (or whatever it's needed) and then just download. The logic about which is the active experiment shouldn't be here. I suspect that this kind of accesses are the ones causing projects to interfere with one another. You click one experiment but the activeProject is not the correct one. If every project passes explicitly it's own ID this will not happen.

src/pages/data-management/index.jsx
```diff
@@ -54,6 +55,13 @@ const DataManagementPage = ({ route }) => {
   }, []);
 
   useEffect(() => {
+    if (activeProject?.experiments.length > 0) {
```
kafkasl: what happens if experimens is undefined? can it happen?

kafkasl (changes requested):

src/pages/data-management/index.jsx
aerlaut: No, `experiments` default to an empty array

src/components/data-management/ProjectDetails.jsx
aerlaut: oh yea, this is supposed to be `false`. Changed accordingly,

src/pages/data-management/index.jsx
kafkasl: actually this is not good because of at least 2 reasons.
1. this isn't awaited so you can be effectively acting on the backend status of the previous experiment
2. this is called every time you change an experiment which is a waste. 

What I suggest (but might be out of the scope of the PR) is to store the backend statuses indexed by experimentID so you just load them once and avoid the potential issue of using the wrong status. If you don't do that then at least `await` the backend status load if possible

src/components/data-management/ProjectDetails.jsx
aerlaut: Oh yea, it's not used anymore. I'll remove it,

src/components/data-management/ProjectDetails.jsx
aerlaut: Refactored to another function. It now takes experimentId as a parameter

vickymorrison:
Suggestion (but not critical): when the rds object is downloaded it appears in my downloads as "r.rds". It would be much better if the file is named with the project name and "processed" e.g. 'ProjectID_processed_object' or similar. The user might download both the raw and processed objects from multiple projects and it would be much better UX if it is clear which is which.

aerlaut:
> Suggestion (but not critical): when the rds object is downloaded it appears in my downloads as "r.rds". It would be much better if the file is named with the project name and "processed" e.g. 'ProjectID_processed_object' or similar. The user might download both the raw and processed objects from multiple projects and it would be much better UX if it is clear which is which.

Sure. The projectId is quite long, so I'll only use the first part of the projectId (e.g. if projectId = AA-BB-CC-DD-EE, I'll use AA for the file name).

Added in https://github.com/biomage-ltd/api/pull/194/commits/fe3570d351163713de2e94e1f4cb1e13f7cb78f5

src/components/data-management/ProjectDetails.jsx
aerlaut: Refactored in 08d66f3

src/components/data-management/ProjectDetails.jsx
```diff
@@ -575,25 +582,50 @@ const ProjectDetails = ({ width, height }) => {
...
+          title={
+            activeProject?.experiments?.length > 0
+            && pipelineHasRun(activeProject.experiments[0])
+              ? 'With Data Processing filters and settings applied'
+              : 'Launch analysis to process data'
+          }
+          placement='left'
+          onClick={() => {
+            // Change if we have more than one experiment per project
+            const experimentId = activeProject.experiments[0];
```
kafkasl: shouldn't this be `activeProject?.experiments[0]`

kafkasl (approved):

src/components/data-management/ProjectDetails.jsx
aerlaut: Existence of experiments[0] is checked on the first clause, so it should be available.

aerlaut: The menu item is now disabled if there is no experimentId, so the click does not work if there is no experimentId.

StefanBabukov (approved):

src/components/data-management/ProjectDetails.jsx
aerlaut: change accordingly

PR #427, opened 2021-08-06
fix classifier legend

alexvpickering:
The legend labels are mixed up:

image

green should be "below" and orange should be "above"


aerlaut (approved):
LGTM 👍

PR #429, opened 2021-08-09
[BIOMAGE-1222] Implement the marker heatmap page in plots and tables

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1222
https://ui-stefan-ui429-api198.scp-staging.biomage.net/
https://github.com/biomage-ltd/api/pull/198

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+    dispatch(loadCellSets(experimentId));
+  }, []);
+
+  useEffect(() => {
+    if (louvainClustersResolution && config && !config.selectedGenes.length) {
+      dispatch(loadMarkerGenes(experimentId, louvainClustersResolution, plotUuid, config.numGenes));
+    }
+  }, [louvainClustersResolution, hierarchy, config?.numGenes]);
+
+  const sortGenes = (newGenes) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:65]: can you add unit tests to check for whether this sorting works, also to verify that we are not sorting when we are removing a gene and any other case you can come up with?

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+const { Panel } = Collapse;
+
+const plotUuid = 'markerHeatmapPlotMain';
+const plotType = 'markerHeatmap';
+
+const route = {
+  path: 'marker-heatmap',
+  breadcrumbName: 'Marker-Heatmap',
+};
+const HeatmapPlot = ({ experimentId }) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I would call this component `MarkerHeatmap`, just to distinguish it from the heatmap component easier

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+  const dispatch = useDispatch();
+  const config = useSelector((state) => state.componentConfig[plotUuid]?.config);
+  const { expression: expressionData } = useSelector((state) => state.genes);
+  const { error, loading } = expressionData;
+  const cellSets = useSelector((state) => state.cellSets);
+  const { hierarchy, properties } = cellSets;
+  const selectedGenes = useSelector((state) => state.genes.expression.views[plotUuid]?.data) || [];
+  const [currentSearchedGenes, setCurrentSearchedGenes] = useState([]);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I can see that the heatmap component in plots and tables doesn't need to have a state variable for the selected genes, but uses `selectedGenes` directly. Why in the case of maker heatmap we need to keep track of this and in the case of the heatmap plot we don't?

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+              onReset={onReset}
+            />
+            <PlotStyling formConfig={plotStylingControlsConfig} config={config} onUpdate={updatePlotWithChanges} defaultActiveKey={['5']} />
+          </Space>
+        </Col>
+      </Row>
+    </div>
+  );
+};
+
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: lots of this and other parts of this file are identical to what we have for the Heatmap component in Plots and Tables -- have you considered reusing the same heatmap component for them and building this MarkerHeatmap component based on it?

src/redux/actionTypes/genes.js
```diff
@@ -60,9 +60,11 @@ const MARKER_GENES_LOADED = `${GENES}/markerGenesLoaded`;
  */
 const MARKER_GENES_ERROR = `${GENES}/markerGenesError`;
 
+const GENE_ORDER_SET = `${GENES}/geneOrderSet`;
```
ivababukova: as I suggested during our call, will be best to keep this elsewhere, since it is something specifically applicable to the marker heatmap in plots and tables, it is not used anywhere else.

src/redux/reducers/componentConfig/initialState.js
```diff
@@ -99,6 +99,41 @@ const heatmapInitialConfig = {
   truncatedValues: false,
 };
 
+// PLOTS & TABLES - Marker heatmap
+const markerHeatmapInitialConfig = {
```
ivababukova, line 106: most of this config is the same as the config of `heatmapInitialConfig`, I would recommend that you modify the code here to copy whatever is in `heatmapInitialConfig` and then just build on top of it. so have something like:

```
const markerHeatmapInitialConfig = {
    ...heatmapInitialConfig,
    groupedTracks: ['louvain'],
    and so on
}
```

src/redux/reducers/genes/index.js
```diff
@@ -3,7 +3,7 @@ import {
   GENES_PROPERTIES_LOADING, GENES_PROPERTIES_LOADED_PAGINATED, GENES_PROPERTIES_ERROR,
   GENES_SELECT, GENES_DESELECT,
   GENES_EXPRESSION_LOADING, GENES_EXPRESSION_LOADED, GENES_EXPRESSION_ERROR,
-  MARKER_GENES_LOADING, MARKER_GENES_LOADED, MARKER_GENES_ERROR,
+  MARKER_GENES_LOADING, MARKER_GENES_LOADED, MARKER_GENES_ERROR, GENE_ORDER_SET,
```
ivababukova: also, this sounds like it is better to be named to something `MARKER_GENE_ORDER_SET` and it definitely should be made clear in the code that it is for plots and tables. This doesn't look like the right place in the store for it

vickymorrison:
Sorry Stefan - I can't review this from the user's perspective because every time I change something it becomes unresponsive.
https://user-images.githubusercontent.com/74908153/129191107-a1dd8e4a-3c65-49e7-8f9c-bbdb30bb7081.mov

vickymorrison:
Does it re-run the 'findmarkers' for the custom cell sets when this option is selected? Or is it still showing markers for the Louvain clusters? Somehow I don't believe the marker genes are being calculated for the custom cells sets as this doesn't look like a marker heatmap for custom cell sets A and B:
Screenshot 2021-08-12 at 14 19 23

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,121 @@
...
+  });
+
+  // marker heatmap renders
+  it('loads initially', async () => {
+    await renderHeatmapPage();
+
+    expect(loadMarkersSpy).toHaveBeenCalled();
+  });
+
+  it('loads marker genes on selecting', async () => {
```
ivababukova, line 209: we need more tests apart from loading marker genes on selecting. 
Tests like "the plot changes on change of a given schema parameter", checking that the schema is updated correctly, checking that the relevant update calls were triggered.
Tests about how would the marker heatmap behave if a given part of the redux store is not defined: what if genes were missing? What if experimentSettings were undefined, etc. 
Try to think about what can possibly go wrong and add tests to check about it.

src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -4,18 +4,19 @@ import {
 
 import { fetchCachedWork } from '../../../utils/cacheRequest';
 
-const loadMarkerGenes = (experimentId, resolution) => async (dispatch, getState) => {
+const loadMarkerGenes = (
+  experimentId, resolution, plotUuid, numGenes = 5,
```
ivababukova, line 8: I think in here you want to be passing `plotType`, not `plotUuid`, because eventually we will have multiple plots of the same type, and then you would want to pass the type, not the uuid which will be unique.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+    if (!louvainClustersResolution) {
+      dispatch(loadProcessingSettings(experimentId));
+    }
+    dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
+    dispatch(loadCellSets(experimentId));
+  }, []);
+
+  useEffect(() => {
+    if (louvainClustersResolution && config && !config.selectedGenes.length) {
+      dispatch(loadMarkerGenes(experimentId, louvainClustersResolution, plotUuid, config.numGenes));
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:62]: better pass plotType, not plotUuid

src/__test__/redux/actions/genes/__snapshots__/loadMarkerGenes.test.js.snap
```diff
@@ -38,6 +38,7 @@ Object {
     "genes": Array [
       "geneA",
     ],
+    "plotUuid": "interactiveHeatmap",
```
ivababukova, line 41: this is plotType, not plotUuid

src/redux/actions/genes/index.js
```diff
@@ -2,10 +2,12 @@ import loadGeneExpression from './loadGeneExpression';
...
 import loadPaginatedGeneProperties from './loadPaginatedGeneProperties';
 import loadMarkerGenes from './loadMarkerGenes';
+import setGeneOrder from './setGeneOrder';
 
 export {
   loadGeneExpression,
   changeGeneSelection,
   loadPaginatedGeneProperties,
   loadMarkerGenes,
+  setGeneOrder,
```
ivababukova: as I said elsewhere, genes store doesn't sound like the correct place to store the order of the genes in the Plots and Tables marker heatmap, consider moving it out.

src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -37,6 +37,7 @@ const loadMarkerGenes = (experimentId, resolution) => async (dispatch, getState)
         experimentId,
         genes: order,
         data: markerGeneExpressions,
+        plotUuid,
```
ivababukova, line 40: again, put plotType here

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I use `currentSelectedGenes` for when the loadexpression fails and we want to rerun it with the old genes which are not in redux yet.

src/redux/actions/genes/loadMarkerGenes.js
StefanBabukov: Yeah but if you look at `markerGenesLoaded.js` you can see that the genes are put under expression.views[plotUuid] thats where selectedGenes are stored. If we have another markerheatmap plot it will have different selectedGenes depending on what the user chose.

StefanBabukov: or should this still be plotType?

vickymorrison:
Metadata track bug - menu says Louvain track is shown but it does not appear on the plot.
Screenshot 2021-08-13 at 14 58 33

vickymorrison:
The re-loading has a lag: I make a change and the heatmap disappears but then the OLD version reappears for a few seconds before the NEW version appears. Video example below of changing the number of genes per custer. This lag is 10 seconds.
https://user-images.githubusercontent.com/74908153/129368438-ccb657cd-acac-415c-bc7d-34f4165c6c6a.mov

vickymorrison:
When scratchpad is selected in the menu I would expect the marker heatmap to update to show me markers for the custom cell sets. This does not happen. 
Screenshot 2021-08-13 at 15 02 01

Following a Slack chat this likely needs bioinformatics input - happy for you to merge this ticket and create a new one for this issue.

StefanBabukov:
The re-loading has a lag: I make a change and the heatmap disappears but then the OLD version reappears for a few seconds before the NEW version appears. Video example below of changing the number of genes per custer. This lag is 10 seconds.
https://user-images.githubusercontent.com/74908153/129368438-ccb657cd-acac-415c-bc7d-34f4165c6c6a.mov
------------
I'm pretty sure this is how rendering vega plots works - shows the previous state of the plot and when its rendered shows the new one - we should work on improving the rendering time in a future ticket (downsampling possibly)

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: the heatmap in plots and tables seems to be able to do that without the need for an extra state variable. Why can't we do this in here too?

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,235 @@
...
+  it('sorts genes properly', async () => {
+    await renderHeatmapPage();
+    await rtl.waitFor(() => expect(configUpdatedSpy).toHaveBeenCalled());
+
+    store.dispatch(loadGeneExpression(experimentId, ['gene0', 'gene1', 'gene3', 'gene2'], plotUuid));
+    await rtl.waitFor(() => expect(configUpdatedSpy).toHaveBeenCalledTimes(2));
+    expect(store.getState().componentConfig[plotUuid].config.selectedGenes).toEqual(['gene0', 'gene1', 'gene2', 'gene3']);
+  });
+
+  it('removing a gene keeps the sorted order', async () => {
```
ivababukova: also check that it is not re-sorting them after removing a gene

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,235 @@
...
+    const nGenesInput = rtl.getByTestId(geneSelection.parentElement, 'num-genes');
+    userEvent.type(nGenesInput, 10);
+
+    const runButton = rtl.getByText(geneSelection.parentElement, 'Run');
+    userEvent.click(runButton);
+
+    expect(loadMarkersSpy).toHaveBeenCalled();
+  });
+
+  it('sorts genes properly', async () => {
```
ivababukova: add tests for cases when you are adding or removing genes, but sort is not necessary.

StefanBabukov: This test is for that when you add a gene - checks if the genes are sorted properly with the newly added gene

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
StefanBabukov: Also added that test too look at the changes

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,235 @@
...
+    <Provider store={store}>
+      <MarkerHeatmap
+        experimentId={experimentId}
+      />
+    </Provider>,
+  );
+  await rtl.waitFor(() => expect(loadConfigSpy).toHaveBeenCalledTimes(1));
+};
+
+describe('Marker heatmap plot', () => {
```
ivababukova, line 187: that's some good addition of tests, but it is not sufficient. Add tests to check for what happens if the redux store is empty, if parts of the redux store are empty, if after requesting expression an error is returned and so on. Try to think what can possibly go wrong in this file and add a test for it.

vickymorrison:
> ## The re-loading has a lag: I make a change and the heatmap disappears but then the OLD version reappears for a few seconds before the NEW version appears. Video example below of changing the number of genes per custer. This lag is 10 seconds.
> https://user-images.githubusercontent.com/74908153/129368438-ccb657cd-acac-415c-bc7d-34f4165c6c6a.mov
> 
> I'm pretty sure this is how rendering vega plots works - shows the previous state of the plot and when its rendered shows the new one - we should work on improving the rendering time in a future ticket (downsampling possibly)

But the issue is that the heatmap disappears while it is being rendered (with "we're getting your data" message), but then the OLD one appears for 5-10secs before the NEW one eventually loads. It would be better if there was simply a longer time of "we're getting your data" as it is confusing to the user that the OLD version reappears.

src/__test__/redux/reducers/genes/markerGenesLoaded.test.js
```diff
@@ -25,6 +25,7 @@ describe('markerGenesLoaded', () => {
       payload: {
         genes,
         data,
+        plotUuid: 'interactiveHeatmap',
```
ivababukova, line 28: change to `plotType`

ivababukova: or actually, judging from file `src/components/data-exploration/heatmap/HeatmapPlot.jsx` it should be `componentType`

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
ivababukova: There are also lots of small functions that are not covered like `getCellClusterFromCellId` -- we need them tested too. Looking at the function `getCellClusterFromCellId`, looks like it will crash if `properties` is undefined or if there is no `clusters` or if `clusters` are empty. 
We have to have guards against all these things and make sure that the code works not only if the user uses it exactly as they are intended to, but when they try to do something else. 
Throughout all this code, try to apply a mindset towards programming for failure -- go through each line and think: "how can possibly this line break?" and then add handling for that and a unit test. Right now, it feels like this is programmed for the golden path use case. I am concerned that if user attempts to do something slightly different, the heatmap will crash.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,235 @@
...
+    const nGenesInput = rtl.getByTestId(geneSelection.parentElement, 'num-genes');
+    userEvent.type(nGenesInput, 10);
+
+    const runButton = rtl.getByText(geneSelection.parentElement, 'Run');
+    userEvent.click(runButton);
+
+    expect(loadMarkersSpy).toHaveBeenCalled();
+  });
+
+  it('sorts genes properly when adding a gene', async () => {
```
ivababukova, line 223: also missing tests for the heatmap states: that when we are loading a gene, the heatmap is in loading state. When there's been an error in loading a gene, the heatmap is in error state.

StefanBabukov: Do we  not have tests for this - loadGeneExpression do i  need to test for that here again?

src/components/plots/helpers/populateHeatmapData.js
```diff
@@ -121,9 +142,11 @@ const populateHeatmapData = (
     // we want to avoid displaying elements which are not in a louvain cluster
     // so initially consider as enabled only cells in louvain clusters
     // See: https://biomage.atlassian.net/browse/BIOMAGE-809
-    const louvainClusters = hierarchy.find((clusters) => clusters.key === 'louvain');
-
-    const cellIsInLouvainCluster = getCellsSetInGroup(louvainClusters);
+    const selectedCellSet = heatmapSettings?.selectedCellSet ? heatmapSettings.selectedCellSet : 'louvain';
```
ivababukova, line 145: I don't understand the addition of line 145, can you explain? Also, is it possible to remove the hardcoding of `louvain` in here? In the very hear future, we will want to support other types of clustering, not only louvain, and this line will be a problem

src/components/plots/styling/PlotStyling.jsx
```diff
@@ -19,8 +19,6 @@ import ColourInversion from './ColourInversion';
 import ExpressionValuesType from './ExpressionValuesType';
 import ExpressionValuesCapping from './ExpressionValuesCapping';
 
-import MetadataTracksDisplayEditor from './MetadataTracksDisplayEditor';
```
ivababukova, line 22: why is this no longer used?

src/components/plots/helpers/populateHeatmapData.js
StefanBabukov: This is for the 'Select data' menu, when we don't have that option in the config - use 'louvain', don't know how can that be not hardcoded at the moment.

src/components/plots/styling/PlotStyling.jsx
StefanBabukov: Duplicate component - look at `HeatmapMetadataTrackSettings`

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+    if (louvainClustersResolution && config) {
+      dispatch(loadMarkerGenes(experimentId, louvainClustersResolution, plotUuid, config.numGenes));
+    }
+  }, [louvainClustersResolution, hierarchy, config?.numGenes]);
+  const sortGenes = (newGenes) => {
+    const clusters = hierarchy.find((cluster) => cluster.key === config.selectedCellSet).children;
+
+    const getCellIdsForCluster = (clusterId) => properties[clusterId].cellIds;
+
+    const getAverageExpressionForGene = (gene, currentCellIds) => {
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:70]: this function is not unit tested either

Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:70]: It is tested - its part of the sorting and sorting is tested and working.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
StefanBabukov: These functions are tested - they are used when sorting, which is tested. What can crash it, if there are no cellsets  - we load them - there wont be an option to sort anything if they are not there.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -0,0 +1,307 @@
...
+      const expressionValues = expressionData.data[gene].rawExpression.expression;
+      let totalValue = 0;
+      currentCellIds.forEach((cellId) => {
+        totalValue += expressionValues[cellId];
+      });
+      return totalValue / currentCellIds.size;
+    };
+
+    const getClusterForGene = (gene) => {
+      let maxAverageExpression = [0];
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I think that the default value of `maxAverageExpression` as a list with 0 is confusing -- what you technically have is a list with 2 items, where the first index shows the value of the max expression and the second shows the index of the cluster where this value occurs. 
The code will be cleaner if you either:
- use 2 variables for this: `maxAvgExpression`, just a number denoting max expression; and `maxExpressionClusterIndex` denoting the cluster index when it occured. Then in line 90 you just return `maxExpressionClusterIndex`.
- use 1 object for this, called `maxAvgExpression` that looks like something like this:
```
let maxAvgExpression = {"expression": 0, "clusterIndex": -1}
```
then on line 90 you return maxAvgExpression["clusterIndex"]

Comment (ivababukova) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: another thing: are you sure that it is not possible to have negative expression? If negative expression is possible, then setting default maxExpression to 0 is incorrect.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
ivababukova: in loadGeneExpression we test the store, in here you would test if the component correctly responds to the updates of the store. So you would check that `Skeleton` gets rendered when the relevant parts of the redux store are in loading state. What you can do is to mock the store, settings some of the contents to be loading: true (like cellSets or geneExpression for this plotUuid) and then assert that instead of MarkerHeatmap, we render `Skeleton`.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,268 @@
...
+
+    await rtl.waitFor(() => expect(loadMarkersSpy).toHaveBeenCalled());
+  });
+
+  it('sorts genes properly when adding a gene', async () => {
+    await renderHeatmapPage(defaultStore);
+    await rtl.waitFor(() => expect(configUpdatedSpy).toHaveBeenCalled());
+    store.dispatch(loadGeneExpression(experimentId, ['gene0', 'gene1', 'gene3', 'gene2'], plotUuid));
+    await rtl.waitFor(() => expect(configUpdatedSpy).toHaveBeenCalledTimes(2));
+    expect(store.getState().componentConfig[plotUuid].config.selectedGenes).toEqual(['gene0', 'gene1', 'gene2', 'gene3']);
```
ivababukova, line 228: much, much better now

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -0,0 +1,268 @@
...
+    expect(loadMarkersSpy).toHaveBeenCalledTimes(1);
+  });
+
+  it('loads cellsets if not available', async () => {
+    const newStore = { ...defaultStore, cellSets: { loading: true } };
+    await renderHeatmapPage(newStore);
+    await rtl.waitFor(() => expect(cellSetsLoadedSpy).toHaveBeenCalled());
+  });
+
+  it('loads processing settings if louvainresolution is not available', async () => {
```
ivababukova, line 256: nice, this is what I meant on the call. Also add a test with `newStore` equal to what is its initial value, then try to load the marker heatmap and assert that the heatmap goes into loading state (Skeleton is displayed) and that dispatches for the relevant actions have been triggered.

src/components/plots/helpers/populateHeatmapData.js
ivababukova: fair enough, let's worry about it later then

src/components/plots/helpers/populateHeatmapData.js
```diff
@@ -238,7 +260,7 @@ const populateHeatmapData = (
 
   data.trackColorData = trackData.map((datum) => datum.trackColorData).flat();
   data.trackGroupData = trackData.map((datum) => datum.groupData).flat();
-
+  data.clusterSeparationLines = trackData[0].clusterSeparationLines;
```
ivababukova, line 263: is it possible that `trackData` will be empty? Because then `trackData[0]` will error out

src/components/plots/styling/heatmap/HeatmapControls.jsx
```diff
@@ -26,18 +60,78 @@ const HeatmapControls = (props) => {
...
         </Space>
       </Panel>
-      <Panel header='Metadata tracks' key='11'>
+      {markerHeatmap && (
+        <>
+          <Panel header='Select Data' key='selectData'>
+            <p>Select the cell sets to show markers for:</p>
+            <Select
+              value={{
+                key: firstLetterUppercase(config.selectedCellSet),
```
ivababukova, line 111: why aren't you using the key of the selectedCellSet? That should be unique, whereas we can have two or more cell sets with the same name

src/utils/plots/getSelectOptions.js
```diff
@@ -0,0 +1,18 @@
+// this returns a formatted array for the select element
+
+const getSelectOptions = (options) => {
```
ivababukova, line 3: look at whether we also use this in other files and also use it there

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -60,7 +60,7 @@ const ExplorationViewPage = ({
     Heatmap: {
       toolbarControls: (
         <>
-          <HeatmapSettings componentType={COMPONENT_TYPE} key='heatmap-settings' />
+          <HeatmapSettings componentType={COMPONENT_UUID} key='heatmap-settings' />
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: I know I asked you to make everything the same type and this is why you changed it to component_uuid, but I didn't realise that now this is going to be confusing here: `componentType={COMPONENT_UUID}`. Sounds like it was maybe better to use component type for data exploration and component uuid for plots and tables. Can you change that back? Sorry!

ivababukova (commented):
some very small things left here in terms of code. Did you look at the loading displyaing previous state briefly issue reported by Vicky?

src/components/plots/helpers/populateHeatmapData.js
StefanBabukov: no, it will return something from generateTrackData

ivababukova (approved):

PR #436, opened 2021-08-12
[BIOMAGE-1112] Validate uploaded files

robioson:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1112
https://ui-james-ui436.scp-staging.biomage.net/

To test you should upload as many data sets as you can to ensure they are not rejected (false positives).

Check that uploading non-compressed files causes them to be compressed prior to upload.

Also test that the pipeline is able to run after uploading compressed files and non-compressed files.

Check that you can replace files.

I will squash before merging.

# Changes
### Code changes

In a nutshell: Validate files by checking signatures instead of extensions.

In detail:
- When a file is selected for upload, read its signature
   - Determine whether it is compressed
   - Determine if it looks like what we expect for the file type (features/matrix/barcodes)
   - Store this "verdict" with the file
- Read this verdict to determine whether to compress the file before upload
- Quite a bit of refactoring of the file upload paths to reduce code duplication

To do:

- [x] remove remaining usage of old validation function
- [x] refactor `loadAndCompressIfNecessary`
- [x] check `uploadFileBundle` in `ProjectDetails` - why is this totally separate to the code in `FileUploadModal`? Right now I think it can leak an uncompressed file
- [x] validate contents of zipped files
- [x] fix tests

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/fileInspector.js
```diff
@@ -0,0 +1,68 @@
...
+      data = Buffer.from(chunk.slice(0, 16));
+    });
+    gunzip.push(await readFileToBuffer(file.slice(0, 128)));
+  }
+
+  const valid = isGzipped ? Verdict.VALID_ZIPPED : Verdict.VALID_UNZIPPED;
+
+  // check matrix file starts with matrix signature
+  if (file.name.startsWith('matrix')
+    && !data.slice(0, MATRIX_SIGNATURE.length).compare(MATRIX_SIGNATURE)) {
```
robioson: Not that important but I HATE the formatting of these conditions - it was the only thing I could get eslint to be happy with

src/utils/fileInspector.js
```diff
@@ -0,0 +1,68 @@
...
+
+  const isGzipped = !data.slice(0, 2).compare(GZIP_SIGNATURE);
+
+  if (isGzipped) {
+    // if gzipped, decompress a small chunk to further validate contents
+    // eslint-disable-next-line no-unused-vars
+    const gunzip = new Gunzip((chunk, _) => {
+      data = Buffer.from(chunk.slice(0, 16));
+    });
+    gunzip.push(await readFileToBuffer(file.slice(0, 128)));
```
robioson: This is more convoluted than it ought to be but it was the only way I could get fflate to decompress only a portion of the file - using `gunzipSync` for example with a slice would just hang

src/utils/fileInspector.js
```diff
@@ -0,0 +1,68 @@
...
+  if (file.name.startsWith('matrix')
+    && !data.slice(0, MATRIX_SIGNATURE.length).compare(MATRIX_SIGNATURE)) {
+    return valid;
+  }
+
+  // check genes file starts with Ensembl Stable ID - ENS or "ENS
+  if ((file.name.startsWith('features') || file.name.startsWith('genes'))
+      && (!data.slice(0, 3).compare(FEATURES_SIGNATURE)
+      || !data.slice(1, 4).compare(FEATURES_SIGNATURE))) {
+    return valid;
```
ivababukova: Is this going to detect an invalid features file because of this reason here: https://this-is-biomage.slack.com/archives/C0250TDU1PF/p1628172894026400?thread_ts=1628171840.026200&cid=C0250TDU1PF ? @ogibson-biomage will be able to provide you with more details on what exactly was invalid there.

src/utils/fileInspector.js
ivababukova: yes, it is quite confusing ... that's because the line can't be longer than 80 chars, right? You can disable a certain eslint condition for a line or for a file btw.

src/utils/fileInspector.js
```diff
@@ -0,0 +1,68 @@
...
+      && strFromU8(data).match(/^[ACGT]+$/)) {
+    return valid;
+  }
+
+  return Verdict.INVALID_FORMAT;
+};
+
+export {
+  inspectFile,
+  Verdict,
```
ivababukova: you can avoid having to export `Verdict` and then requiring other files like in here for example: https://github.com/biomage-ltd/ui/pull/436/files#diff-4ed70651f95f5b1b4906927dd5f27e7eb831492e15906cfe63be1c7ac4a35558R204 by directly returning the code + the error message if validation failed or success message if it didn't.

src/utils/loadAndCompressIfNecessary.js
```diff
@@ -1,37 +1,25 @@
...
-
-          resolve(uintArrayToBuffer(compressedFile));
-        });
-      }
-    };
-
-    reader.readAsArrayBuffer(bundle);
+import readFileToBuffer from './readFileToBuffer';
+
+const compress = (buffer) => new Promise((resolve, reject) => {
```
ivababukova: I see there are a bunch of helper methods in utils for data upload, like `src/utils/readFileToBuffer.js` and this one for example. Why don't you create a new folder in `utils` for `data_upload` and move these files there, so that they are easier to find?

src/__test__/utils/processUpload.test.js
```diff
@@ -23,32 +23,35 @@ const validFilesList = [
...
     },
     upload: { status: UploadStatus.UPLOADING },
     errors: '',
+    compressed: true,
+    valid: true,
   },
   {
     name: 'WT13/matrix.mtx.gz',
     bundle: {
-      name: 'features.tsv.gz',
```
ogibson, line 45: what was this?

src/__test__/utils/processUpload.test.js
```diff
@@ -23,32 +23,35 @@ const validFilesList = [
...
   {
     name: 'WT13/barcodes.tsv.gz',
     bundle: {
-      name: 'features.tsv.gz',
+      name: 'barcodes.tsv.gz',
       path: '/WT13/barcodes.tsv.gz',
       type: 'application/gzip',
-      valid: true,
     },
     upload: { status: UploadStatus.UPLOADING },
```
ivababukova, line 39: do you have any unit tests that test the behaviour when the user is trying to upload some invalid files?

src/utils/fileInspector.js
robioson: Yes, it would catch this! That data seems to have gene names instead of Ensembl IDs.

src/__test__/utils/processUpload.test.js
robioson: The filename in the `bundle.name` property didn't match the outer `name` property. It was simply a copy-paste typo when the tests were written.

src/utils/fileInspector.js
robioson: It was mainly because eslint wants the logical operators to be at the start of the line rather than the end, imo it makes it more confusing.

src/__test__/utils/processUpload.test.js
robioson: There are tests for what happens when a file it marked as invalid but not for the validation itself. I will try to add them by mocking `readFileToBuffer`.

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -29,12 +28,14 @@ const UploadDetailsModal = (props) => {
     if (replacementFileBundle) {
       // we'll need to remove the hard-coded 10x tech type once we start
       // supporting other types and save the chosen tech type in redux
```
ivababukova: in order to make it more visible that this is hardcoded and it is hardcoded here, why don't you create a const "TECH_TYPE", make it equal to `10X Chromium` and put it at the top of this file?

src/utils/upload/fileInspector.js
```diff
@@ -0,0 +1,67 @@
...
+    return valid;
+  }
+
+  return Verdict.INVALID_FORMAT;
+};
+
+export {
+  inspectFile,
+  Verdict,
+};
```
ivababukova, line 74: Somehow it got lost, but I had a comment here proposing that instead of exporting `Verdict` and having `src/utils/upload/processUpload.js` to have to understand what each error code in `Verdict` means, you can make function `inspectFile` return the code + the mesage directly. Then print to the user the message returned by the function if there is an error directly

src/components/data-management/FileUploadModal.jsx
```diff
@@ -57,31 +55,11 @@ const FileUploadModal = (props) => {
...
-        bundle: file,
-        upload: {
-          status: UploadStatus.UPLOADING,
-          progress: 0,
-        },
-        valid: valid.isValidType && valid.isValidFilename,
-        errors: error.join(', '),
-      });
-    });
+    const newFiles = await Promise.all(filteredFiles.map(async (file) => (
```
StefanBabukov: dont think the function passed in map needs to be async

src/components/data-management/FileUploadModal.jsx
```diff
@@ -38,12 +37,11 @@ const FileUploadModal = (props) => {
...
 
   // Handle on Drop
-  const onDrop = (acceptedFiles) => {
-    const newList = [];
+  const onDrop = async (acceptedFiles) => {
     let filesNotInFolder = false;
     const filteredFiles = acceptedFiles
       // Remove all hidden files
-      .filter((file) => !file.name.startsWith('.'))
+      .filter((file) => !file.name.startsWith('.') && !file.name.startsWith('__MACOSX'))
```
StefanBabukov, line 44: Can you explain what is this check for starting with __MACOSX? I dont know

src/utils/upload/fileUploadSpecifications.js
```diff
@@ -1,6 +1,6 @@
 const fileUploadSpecifications = {
   '10X Chromium': {
-    acceptedFiles: [
+    acceptedFiles: new Set([
```
StefanBabukov, line 3: Why do you make this a set if they are hardcoded and we dont add values here?

robioson: To check for membership efficiently

src/components/data-management/FileUploadModal.jsx
robioson: Creating zip files on macOS creates this irritating hidden directory. I remember a user a while ago complaining about these files being selected for upload and having to manually remove them, so I thought I might as well add this to the filtering logic while I was here.

src/utils/upload/processUpload.js
```diff
@@ -196,5 +186,41 @@ const processUpload = async (filesList, sampleType, samples, activeProjectUuid,
...
+const bundleToFile = async (bundle, technology) => {
+  // This is the first stage in uploading a file.
+
+  let filename;
+  if (bundle.path) {
+    // trim path to just the file and its folder
+    filename = _.takeRight(bundle.path.split('/'), 2).join('/');
+  } else {
+    filename = bundle.name;
+  }
```
StefanBabukov: suggestion - make this into a ternary operator?

src/components/data-management/FileUploadModal.jsx
robioson: Correct, I also noticed this but forgot! Thanks

src/components/data-management/FileUploadModal.jsx
StefanBabukov: Interesting, thanks!

src/utils/upload/fileUploadSpecifications.js
StefanBabukov: Oh so you mean `has()` is faster than `includes()`?

src/utils/upload/processUpload.js
robioson: Yeah, this is probably neater

src/utils/upload/fileUploadSpecifications.js
robioson: Yes, `Set#has` is `O(1)` whereas `Array#includes` is `O(n)`

robioson: Before we created a regex from the array but this was unnecessarily complicated

src/utils/upload/fileInspector.js
robioson: Do you mean like `return [code, message]`?

src/utils/upload/fileInspector.js
```diff
@@ -0,0 +1,66 @@
...
+
+const Verdict = {
+  INVALID_NAME: -2,
+  INVALID_FORMAT: -1,
+  VALID_ZIPPED: 0,
+  VALID_UNZIPPED: 1,
+};
+
+const MATRIX_SIGNATURE = Buffer.from('%%MatrixMarket');
+const FEATURES_SIGNATURE = Buffer.from('ENS');
```
robioson: Feature files can also start with `lnc` - https://this-is-biomage.slack.com/archives/C01CP46TJ0Z/p1627654709006400

src/utils/upload/fileInspector.js
```diff
@@ -0,0 +1,66 @@
...
+  // check genes file starts with Ensembl Stable ID - ENS or "ENS
+  if ((file.name.startsWith('features') || file.name.startsWith('genes'))
+      && (!data.slice(0, 3).compare(FEATURES_SIGNATURE)
+      || !data.slice(1, 4).compare(FEATURES_SIGNATURE))) {
+    return valid;
+  }
+
+  // check barcodes file starts with a 16 digit DNA sequence
+  if (file.name.startsWith('barcodes')
+      && data.toString().match(/^[ACGT]+$/)) {
```
robioson, line 64: This might not always be valid

src/utils/upload/fileInspector.js
ivababukova: yes

ivababukova:
I tried using it right now and uploaded some videos. As you can see on the image, they show a cross for being invalid, but there is no message that they are invalid or what is invalid:
Screenshot 2021-08-20 at 09 55 32
Another little thing is that I tried to hover over the red cross and that didn't make the tooltip show up -- it will be better if the tooltip shows up both on hover over the filename and hover over the red cross.

robioson:
This is because the filenames are so long. This was a problem prior to this PR. The tooltips would help with this.

src/utils/upload/fileInspector.js
robioson: One problem with this is that we check the verdict again here https://github.com/biomage-ltd/ui/pull/436/files#diff-cf8293615aa71eb30ec9e5199c47f4cc10f13be1ffa44f2319797cd448c85bcbR216 You need to export it anyway for this

ivababukova: fair enough, then maybe leave it as is

ivababukova (approved):

robioson:
I gave up trying to squash this sadly, it looks complicated when you have merge commits in between. Lesson for the future: don't merge in master repeatedly if you want to squash.

PR #439, opened 2021-08-13
Only showing up new project modal when the user logs in

kafkasl:
The modal pops in and out because of the logic making it appear on project delete. It's a mess because experiments / projects are updated all over the place so we can't reliably know when to display it. The compromise solution is to just show it up the first time the user logs in. 

**Note** this is needed to have reliable integration tests.


aerlaut (approved):
LGTM

PR #458, opened 2021-08-23
[BIOMAGE-1273] Refactor ProjectDetails component

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1273
https://ui-stefan-ui458.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/pages/data-management/index.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable import/no-unresolved */
```
ivababukova, line 1: why are you adding this?

StefanBabukov: because we can import stuff without a relative path but like 'utils/something.js' eslint still marks it as unresolved though.

src/components/data-management/ProjectDetails.jsx
```diff
@@ -723,117 +530,33 @@ const ProjectDetails = ({ width, height }) => {
         visible={analysisModalVisible}
         onLaunch={async (experimentId) => {
           const lastViewed = moment().toISOString();
-          await dispatch(updateExperiment(experimentId, { lastViewed }));
-          await dispatch(updateProject(activeProjectUuid, { lastAnalyzed: lastViewed }));
+          dispatch(updateExperiment(experimentId, { lastViewed }));
+          dispatch(updateProject(activeProjectUuid, { lastAnalyzed: lastViewed }));
```
ivababukova, line 347: why removing the await?

src/components/data-management/Header.jsx
```diff
@@ -0,0 +1,106 @@
...
+            placement='bottomRight'
+            disabled={
+              projects.ids.length === 0
+            || activeProject?.samples?.length === 0
+            }
+          >
+            <Button>
+              Download
+            </Button>
+          </Dropdown>
```
ivababukova: Why don't you move all that stuff in `DownloadData` component and in here just call it, passing all the props needed?

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,112 @@
...
+  const experimentSettings = useSelector((state) => state.experimentSettings);
+  const experiments = useSelector((state) => state.experiments);
+  const samples = useSelector((state) => state.samples);
+
+  const pipelineHasRun = (experimentId) => (
+    experiments[experimentId]?.meta?.backendStatus?.pipeline?.status === pipelineStatus.SUCCEEDED
+  );
+  const gem2sHasRun = (experimentId) => (
+    experiments[experimentId]?.meta?.backendStatus?.gem2s?.status === pipelineStatus.SUCCEEDED
+  );
```
ivababukova: since every time you also have a check for `activeProject?.experiments?.length > 0` when you use `pipelineHasRun` and `gem2sHasRun`, I would recommend that you move this check: `activeProject?.experiments?.length > 0` to be part of these statements

ivababukova: Why do you actually have to check that too? Wouldn't it be sufficient if `experimentId` is in `experiments`?

src/pages/data-management/index.jsx
aerlaut: This is because we added `baseUrl` in `jsconfig.json` : https://nextjs.org/docs/advanced-features/module-path-aliases, but it's not yet widely used it seems

src/components/data-management/ProjectDetails.jsx
StefanBabukov: because we don't need it for dispatching actions

src/components/data-management/DownloadData.jsx
StefanBabukov: We need these checks to determine whether we can download :
raw seurat object - needs gem2s to have finished
processed seurat object - needs pipeline to have finished

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }, [activeProject]);
+
+  const pipelineHasRun = (experimentId) => (
+    (backendStatus[experimentId]?.status.pipeline?.status === pipelineStatus.SUCCEEDED)
+    && activeProject?.experiments?.length
+  );
+  const gem2sHasRun = (experimentId) => (
+    (backendStatus[experimentId]?.status?.gem2s?.status === pipelineStatus.SUCCEEDED)
+   && activeProject?.experiments?.length
```
ivababukova: add `> 0` after length to make it more clean

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const allSamplesAnalysed = () => {
+    // Returns true only if there is at least one sample in the currently active
+    // project AND all samples in the project have been analysed.
+    const steps = Object.values(_.omit(experimentSettings?.processing, ['meta']));
+
+    return steps.length > 0
+      && activeProject?.samples?.length > 0
+      && activeProject?.samples?.every((s) => steps[0].hasOwnProperty(s));
+  };
```
ivababukova: in this code, there is no point computing `steps` on line 44 if `activeProject?.samples?.length` is 0 and if `activeProject?.samples?.every((s) => steps[0].hasOwnProperty(s));`. Since checking for `activeProject?.samples?.length` is the "cheapest" in terms of complexity (it is basically O(1)), I would modify this `allSamplesAnalysed` function to first check for that, then return False if this is not the case, or if it is true, carry on with the other 2 checks.

src/components/data-management/Header.jsx
```diff
@@ -0,0 +1,92 @@
...
+  const dispatch = useDispatch();
+  const activeProject = useSelector((state) => state.projects[activeProjectUuid]) || false;
+  const projects = useSelector((state) => state.projects);
+  const changeDescription = (description) => {
+    dispatch(updateProject(activeProjectUuid, { description }));
+  };
+  return (
+    <>
+      <Row style={{ display: 'flex', justifyContent: 'space-between' }}>
+        <Title level={3}>{activeProject.name}</Title>
```
ivababukova: this will error out of `activeProject` is set to false, which seems to be the default value set for it in line 19.

src/components/data-management/Header.jsx
```diff
@@ -0,0 +1,92 @@
...
+    activeProjectUuid, createMetadataColumn, isAddingMetadata, canLaunchAnalysis,
+    setUploadModalVisible, openAnalysisModal,
+  } = props;
+  const dispatch = useDispatch();
+  const activeProject = useSelector((state) => state.projects[activeProjectUuid]) || false;
+  const projects = useSelector((state) => state.projects);
+  const changeDescription = (description) => {
+    dispatch(updateProject(activeProjectUuid, { description }));
+  };
+  return (
```
ivababukova: instead of checking in every bit in here if `projects.ids.length === 0`, I would add an if statement for it before this return, then if the if is true, we can just return only the thing we can, which is:

```
<Row>
        <Col>
          {
            activeProjectUuid && (
              <Space direction='vertical' size='small'>
                <Text type='secondary'>{`ID : ${activeProjectUuid}`}</Text>
                <Text strong>Description:</Text>
                <Paragraph
                  editable={{ onChange: changeDescription }}
                >
                  {activeProject.description}

                </Paragraph>
              </Space>
            )
          }
        </Col>
      </Row>
```
if it is false, then do the logic you have in here, but remove all these `activeProject?.samples?.length === 0` checks

src/components/data-management/Header.jsx
```diff
@@ -0,0 +1,92 @@
...
+            disabled={projects.ids.length === 0}
+            onClick={() => setUploadModalVisible(true)}
+          >
+            Add samples
+          </Button>
+          <Button
+            disabled={
+              projects.ids.length === 0
+            || activeProject?.samples?.length === 0
+            || isAddingMetadata
```
ivababukova: this `isAddingMetadata` doesn't read very well, or at least it is not obvious to me what it is about. Can you explain?

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,136 @@
...
+          </Menu.Item>
+          <Menu.Item
+            key='download-processed-seurat'
+            disabled={
+              !pipelineHasRun(activeProject?.experiments[0])
+            }
+            onClick={() => {
+              // Change if we have more than one experiment per project
+              const experimentId = activeProject?.experiments[0];
+              downloadData(experimentId, downloadTypes.PROCESSED_SEURAT_OBJECT);
```
ivababukova: this `downloadData` function is defined in `utils`, but not used anywhere in the code apart from in here. In this case, having it in a separate file only complicates the code, because now you have to be calculating `experimentId` for it, also have to search for it and so on. I suggest you move the function `downloadData` to be defined in this file, remove the `experimentId` from its argument and have `const experimentId = activeProject?.experiments[0];` as the first line of the function. I would also rename `downloadData` to `downloadExperimentData`, because it looks like you are downloading data for the experiment id

src/components/data-management/Header.jsx
```diff
@@ -0,0 +1,92 @@
...
+import {
+  Row, Typography, Space, Button, Dropdown, Col,
+} from 'antd';
+import {
+  updateProject,
+} from '../../redux/actions/projects'; import DownloadData from './DownloadData';
+
+const { Title, Text, Paragraph } = Typography;
+
+const Header = (props) => {
```
ivababukova: I would rename this component: I wouldn't call it `Header`, because we already have `Header` in Plots and Tables and Data Exploration that means something else and it is confusing. This seems to be about the Project buttons, it is more of a project menu kind of thing. Definitely not a Header.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,136 @@
...
+import {
+  Menu, Tooltip, Dropdown, Button,
+} from 'antd';
+import { useSelector, useDispatch } from 'react-redux';
+import PropTypes from 'prop-types';
+import { saveAs } from 'file-saver';
+import pipelineStatus from '../../utils/pipelineStatusValues';
+import downloadData from '../../utils/data-management/downloadExperimentData';
+import downloadTypes from '../../utils/data-management/downloadTypes';
+import { exportQCParameters, filterQCParameters } from '../../utils/exportQCParameters';
```
ivababukova, line 12: Functions in `exportQCParameters` are only used in this file, so I suggest removing them from utils and moving them in here

ivababukova: I would move this `exportQCParameters` file to the data management folder in utils

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+  useEffect(() => {
+    if (activeProject?.experiments?.length) {
+      dispatch(loadBackendStatus(activeProject?.experiments[0]));
+    }
+  }, [activeProject]);
+
+  const pipelineHasRun = (experimentId) => (
+    (backendStatus[experimentId]?.status.pipeline?.status === pipelineStatus.SUCCEEDED)
+    && activeProject?.experiments?.length
```
aerlaut: Are the `&& activeProject?.experiments?.length` still necessary? Both this and and `gem2hasRun` is passed experimentId from `activeProject.experiments.length[0]`, so this `activeProject?.experiments?.length` must be > 0?

aerlaut: Maybe what you can do is first check if `experimentId !== undefined`, as it may return undefined from `activeProject?.experiments[0]`

src/components/data-management/Header.jsx
StefanBabukov: We use it when the user is in the middle of adding metadata - disable the Add metadata button so they can't click it again while creating metadata

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,163 @@
...
+  const activeProject = useSelector((state) => state.projects[activeProjectUuid]);
+  const experimentSettings = useSelector((state) => state.experimentSettings);
+  const backendStatus = useSelector((state) => state.backendStatus);
+  const samples = useSelector((state) => state.samples);
+  const projects = useSelector((state) => state.projects);
+  const experimentId = activeProject?.experiments[0];
+
+  useEffect(() => {
+    if (activeProject?.experiments?.length) {
+      dispatch(loadBackendStatus(activeProject?.experiments[0]));
```
ivababukova: use experimentId

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,122 @@
...
+} from '../../redux/actions/projects'; import DownloadData from './DownloadData';
+import fileUploadSpecifications from '../../utils/upload/fileUploadSpecifications';
+import UploadStatus from '../../utils/upload/UploadStatus';
+
+const { Title, Text, Paragraph } = Typography;
+
+const ProjectMenu = (props) => {
+  const {
+    activeProjectUuid, createMetadataColumn, isAddingMetadata,
+    setUploadModalVisible, openAnalysisModal,
```
ivababukova, line 18: take a look if any of these functions aren't bettter to be moved to this file and not imported as props, dimilarly to what you did with `checkLaunchAnalysis`

src/pages/data-management/index.jsx
ivababukova: what is it about? If it is not used, should we remove it?

src/components/data-management/SamplesTable.jsx
```diff
@@ -0,0 +1,120 @@
...
+      helperClass='row-dragging'
+      onSortEnd={onSortEnd}
+      {...otherProps}
+    />
+  );
+
+  const DraggableRow = (otherProps) => {
+    const index = tableData.findIndex((x) => x.key === otherProps['data-row-key']);
+    return <SortableRow index={index} {...otherProps} />;
+  };
```
ivababukova, line 86: can you move these components to not be nested inside `SamplesTable`, but rather be defined on the same level as `SamplesTable`? Better to put them on top

ivababukova: it is quite confusing to reason about them and quickly realise which one depends on which and so on

src/components/data-management/SamplesTable.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+  return (
+    <Row>
+      <Col>
+        <Table
+          size='small'
+          scroll={{
+            x: 'max-content',
+            y: height - 250,
```
ivababukova, line 95: this `y: height - 250` : does it actually work for various screen sizes?

StefanBabukov: this defines, how long the table should be to start scrolling it, what do you mean work for various screen sizes?

ivababukova: sounds like it is ok then

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,122 @@
...
+  } = props;
+  const dispatch = useDispatch();
+  const activeProject = useSelector((state) => state.projects[activeProjectUuid]) || false;
+  const projects = useSelector((state) => state.projects);
+  const samples = useSelector((state) => state.samples);
+  const changeDescription = (description) => {
+    dispatch(updateProject(activeProjectUuid, { description }));
+  };
+
+  const checkLaunchAnalysis = () => {
```
ivababukova: I would rename this to `canLaunchAnalysis`, to make it obvious that this function returns a boolean that states whether it is ok for the user to launch analysis or not.

src/pages/data-management/index.jsx
StefanBabukov: I'm using it in some places i find it convenient to import things without gettting lost in the path

aerlaut:
- Would it be better to also refactor `renderUploadCell`, `renderEditableFieldCell`, `renderSampleCells` into separate components?
- Is it also possible to take out metadata-related functions out of `ProjectDetails`?

ivababukova:
This PR reduces code coverage a bit. @StefanBabukov do you mind adding a few tests for the uncovered parts? If you are unsure what you can test for, call me and we can take a look together.

src/components/data-management/MetadataColumns.jsx
```diff
@@ -0,0 +1,99 @@
...
+};
+InitializedMetadataColumn.propTypes = {
+  name: PropTypes.string.isRequired,
+  validateInputs: PropTypes.func.isRequired,
+  setCells: PropTypes.func.isRequired,
+  deleteMetadataColumn: PropTypes.func.isRequired,
+  key: PropTypes.string.isRequired,
+  activeProjectUuid: PropTypes.string.isRequired,
+};
+export { TemporalMetadataColumn, InitializedMetadataColumn };
```
ivababukova: I find it hard to understand a couple of things:
- what are these components doing? Maybe their names are a bit missleading
- why are they put in the same file? They don't seem to be inter-related

StefanBabukov: TemporalMetadataColumn - temporary column until you provide a name or cancel it,
InitialisedMetadataColumn- already created metadata column

Felt like they were similar so i included them in one file so its easier to understand

ivababukova: Ok, so I think that the names are confusing. `TemporalMetadataColumn`: it is actually not a column yet, it is just a popover for adding in the metadata name. And I just saw that there is an existing file: `src/components/data-management/MetadataPopover.jsx`, so I would remove this `TemporalMetadataColumn` entirely and move everything left to the `MetadataPopover` component.

ivababukova: Now this `InitializedMetadataColumn` -- it is actually the metadata column, so I would name it `MetadataColumn`.

src/components/data-management/SamplesTableCells.jsx
```diff
@@ -0,0 +1,166 @@
...
+        onAfterSubmit={(name) => dispatch(updateSample(record.uuid, { name }))}
+        onDelete={() => dispatch(deleteSamples([record.uuid]))}
+      />
+    </Text>
+  );
+};
+SampleCells.propTypes = {
+  cellInfo: PropTypes.object.isRequired,
+};
+export { UploadCell, EditableFieldCell, SampleCells };
```
ivababukova: what are these `UploadCell` and `EditableFieldCell` and `SampleCells` for?

src/components/data-management/SamplesTableCells.jsx
```diff
@@ -0,0 +1,166 @@
...
+import UploadStatus, { messageForStatus } from '../../utils/upload/UploadStatus';
+import EditableField from '../EditableField';
+
+const { Text } = Typography;
+const UploadCell = (props) => {
+  const { file, showDetails } = props;
+
+  const { progress = null, status = null } = file?.upload ?? {};
+
+  if (status === UploadStatus.UPLOADED) {
```
ivababukova, line 22: rename `UploadStatus` to `UploadStates`

src/components/data-management/SamplesTableCells.jsx
```diff
@@ -0,0 +1,166 @@
...
+import { useDispatch } from 'react-redux';
+import PropTypes from 'prop-types';
+import {
+  deleteSamples, updateSample,
+} from '../../redux/actions/samples';
+import UploadStatus, { messageForStatus } from '../../utils/upload/UploadStatus';
+import EditableField from '../EditableField';
+
+const { Text } = Typography;
+const UploadCell = (props) => {
```
ivababukova, line 17: rename `UploadCell` to `UploadStatus`

src/__test__/components/data-management/ProjectDetails.test.jsx
```diff
@@ -3,18 +3,22 @@ import '@testing-library/jest-dom';
...
 import thunk from 'redux-thunk';
-import { render, screen } from '@testing-library/react';
-
+import * as rtl from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { createStore, applyMiddleware } from 'redux';
+import _ from 'lodash';
+import { fireEvent } from '@testing-library/dom';
+import rootReducer from '../../../redux/reducers/index';
+import * as projectsMetadataCreate from '../../../redux/actions/projects/createMetadataTrack';
```
cosa65: since we are getting the default export I would use the same name instead of `projectsMetadataCreate ` as the file

cosa65: Why are using the `import * as x from ...` pattern instead of `import x from ...`?

src/__test__/components/data-management/ProjectDetails.test.jsx
```diff
@@ -3,18 +3,22 @@ import '@testing-library/jest-dom';
 import { Provider } from 'react-redux';
 import configureStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
-import { render, screen } from '@testing-library/react';
-
+import * as rtl from '@testing-library/react';
```
cosa65, line 6: You can do `import rtl, { render, screen } from '@testing-library/react';` instead which is the most common pattern we usually follow in the project I think

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+  const projects = useSelector((state) => state.projects);
+
+  // Change if we have more than one experiment per project
+  const experimentId = activeProject?.experiments[0];
+
+  useEffect(() => {
+    if (experimentId) {
+      dispatch(loadBackendStatus(experimentId));
+    }
+  }, [activeProject]);
```
cosa65: Could we use experimentId instead here?

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+  const backendStatus = useSelector((state) => state.backendStatus);
+  const samples = useSelector((state) => state.samples);
+  const projects = useSelector((state) => state.projects);
+
+  // Change if we have more than one experiment per project
+  const experimentId = activeProject?.experiments[0];
+
+  useEffect(() => {
+    if (experimentId) {
+      dispatch(loadBackendStatus(experimentId));
```
cosa65, line 37: In the case we already have a backend status for this experiment, do you think we would want to load it again? I am unsure of what would be best, so this is a pretty open question, not a suggestion.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+    }
+  }, [activeProject]);
+
+  const pipelineHasRun = () => (
+    experimentId
+    && (backendStatus[experimentId]?.status.pipeline?.status === pipelineStatus.SUCCEEDED)
+
```
cosa65, line 47: remove this line

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+    }
+  };
+
+  return (
+    <Dropdown
+      overlay={() => (
+        <Menu>
+          <Menu.Item
+            key='download-raw-seurat'
+            disabled={!gem2sHasRun()}
```
cosa65: Would this fit better in a `useState` `useEffect` pattern perhaps?

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+            onClick={() => {
+              downloadExperimentData(downloadTypes.RAW_SEURAT_OBJECT);
+            }}
+          >
+            <Tooltip
+              title={
+                gem2sHasRun()
```
cosa65: Following 
> Would this fit better in a useState useEffect pattern perhaps?
We are calling it here too, so perhaps that useState would be better.

Otherwise, we still only need to call it once per render, so if we don't use the hooks we could still have a 
`const gem2sHasRun = getGem2sHasRun()` or something and then just call that in these places.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+              }
+              placement='left'
+            >
+              Raw Seurat object (.rds)
+            </Tooltip>
+          </Menu.Item>
+          <Menu.Item
+            key='download-processed-seurat'
+            disabled={
+              !pipelineHasRun()
```
cosa65: Same thing about `useState` as with gem2s.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+            }
+            onClick={() => {
+              // Change if we have more than one experiment per project
+              downloadExperimentData(downloadTypes.PROCESSED_SEURAT_OBJECT);
+            }}
+          >
+            <Tooltip
+              title={
+                pipelineHasRun(activeProject?.experiments[0])
```
cosa65: This parameter is unused.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,167 @@
...
+                  ? 'With Data Processing filters and settings applied'
+                  : 'Launch analysis to process data'
+              }
+              placement='left'
+            >
+              Processed Seurat object (.rds)
+            </Tooltip>
+          </Menu.Item>
+          <Menu.Item
+            disabled={!allSamplesAnalysed()}
```
cosa65: Same comment as with `pipelinesHasRun` and `gem2sHasRun`

src/components/data-management/DownloadData.jsx
cosa65: Also I would call it `qcHasRun`, pipeline (except in places we have not updated it yet) is used to refer to either gem2s or qc now.

src/components/data-management/MetadataColumn.jsx
```diff
@@ -0,0 +1,57 @@
...
+    rules.MIN_1_CHAR,
+    rules.ALPHANUM_SPACE,
+    rules.START_WITH_ALPHABET,
+    rules.UNIQUE_NAME_CASE_INSENSITIVE,
+  ];
+  return (
+    <Space>
+      <EditableField
+        deleteEnabled
+        onDelete={(e, currentName) => deleteMetadataColumn(currentName)}
```
cosa65, line 29: I would keep the `onDelete` pattern in the prop we receive, so `onDeleteMetadataColumn`, if this pattern is used somewhere else I would also change it to `on...` which what is normally used.

src/components/data-management/MetadataColumn.jsx
```diff
@@ -0,0 +1,57 @@
...
+        onAfterSubmit={(newName) => dispatch(
+          updateMetadataTrack(name, newName, activeProjectUuid),
```
cosa65, line 31: This should be dispatched, right?

src/components/data-management/MetadataColumn.jsx
```diff
@@ -0,0 +1,57 @@
...
+        )}
+        value={name}
+        validationFunc={
+          (newName) => validateInputs(newName, metadataNameValidation)
```
cosa65: Why inputs instead of input?

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,121 @@
...
+
+const { Title, Text, Paragraph } = Typography;
+
+const ProjectMenu = (props) => {
+  const {
+    activeProjectUuid, createMetadataColumn, isAddingMetadata,
+    setUploadModalVisible, openAnalysisModal,
+  } = props;
+  const dispatch = useDispatch();
+  const activeProject = useSelector((state) => state.projects[activeProjectUuid]) || false;
```
cosa65: Why are we using a `|| false`? Isn't it better to just keep the `null` or `undefined` we get?

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,121 @@
...
+  const projects = useSelector((state) => state.projects);
+  const samples = useSelector((state) => state.samples);
+  const changeDescription = (description) => {
```
cosa65: I feel like we could just call `dispatch(updateProject(activeProjectUuid, { description }));` in the single place we are using this function instead of adding this function and calling it.

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,121 @@
...
+    dispatch(updateProject(activeProjectUuid, { description }));
+  };
+
+  const canLaunchAnalysis = () => {
+    if (activeProject?.samples?.length === 0 || projects.ids.length === 0) return false;
```
cosa65: we are using `projects.ids.length === 0` in a lot of places, we could instead just create a
`const anyProjectsAvaliable = projects.ids.length` and make it more readable.

Also I feel like we could apply this same change in other files too, could you check?

cosa65: As an example, we could do this same thing with `activeProject?.metadataKeys.length` here too I believe

src/components/data-management/ProjectMenu.jsx
```diff
@@ -0,0 +1,121 @@
...
+            Add samples
+          </Button>
+          <Button
+            disabled={
+              projects.ids.length === 0
+            || activeProject?.samples?.length === 0
+            || isAddingMetadata
+            }
+            onClick={() => {
+              createMetadataColumn();
```
cosa65, line 75: `onCreateMetadataColumn`

src/components/data-management/SamplesTableCells.jsx
```diff
@@ -0,0 +1,166 @@
...
+  const { file, showDetails } = props;
+
+  const { progress = null, status = null } = file?.upload ?? {};
+
+  if (status === UploadStatus.UPLOADED) {
+    return (
+      <div
```
cosa65, line 24: I feel like we could perhaps separate each of these into other files, one for each type of table cell (for each status)?

cosa65: It would make this file significantly shorter

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -57,13 +62,31 @@ const UploadDetailsModal = (props) => {
     return `${weekDayName}, ${fullDate} at ${fullTime}`;
   };
 
+  const downloadFile = async () => {
+    const bucketKey = `${activeProjectUuid}/${sampleUuid}/${file.name}`;
+
+    const downloadedS3Object = await Storage.get(bucketKey, { download: true });
```
cosa65: could we move this logic interacting with external services to a folder in utils? perhaps create a folder for s3 or something and move both this and the `upload` folder in utils into it?

src/pages/data-management/index.jsx
cosa65: I didn't know that this had been added, is there a ticket for making the existing relative paths absolute?

Also we should disable this eslint policy in the whole project if it is necessary to add it in every single file we use this.

cosa65 (dismissed):

src/__test__/components/data-management/ProjectDetails.test.jsx
StefanBabukov: If I import it like you requested it errors for some reason 
`import createMetadataTrack from '../../../redux/actions/projects/createMetadataTrack';`
image

src/__test__/components/data-management/ProjectDetails.test.jsx
StefanBabukov: render and screen are a subset of rtl, so that doesnt work
image

src/components/data-management/DownloadData.jsx
StefanBabukov: I'll add a check for !backendStatus[experimentId] so we load it if only there is no data for the experiment

src/components/data-management/MetadataColumn.jsx
StefanBabukov: Yeah this is when you make changes to the editable field we want to update it

src/components/data-management/MetadataColumn.jsx
StefanBabukov: EditableField is used in many places so passing onDeleteMetadataColumn won't be accurate for the other use cases

src/components/data-management/SamplesTableCells.jsx
StefanBabukov: Wouldn't this make it a bigger mess - like 4 other components which  are used in only one place? We already have a lot of stuff in components/data-management, plus managing the passing of props and etc. that file is not that big in my opinion, what do you think?

StefanBabukov:
"I didn't know that this had been added, is there a ticket for making the existing relative paths absolute?"

No there isn't. I don't think we should change every path to absolute - we already have them defined correctly so what's the point? I was thinking to use this in further imports you make so its easier

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,171 @@
...
+      setTimeout(() => {
+        URL.revokeObjectURL(link.href);
+        link.parentNode.removeChild(link);
+      }, 0);
+    } catch (e) {
+      pushNotificationMessage('error', endUserMessages.ERROR_DOWNLOADING_DATA);
+    }
+  };
+
+  return (
```
ivababukova, line 86: There is a bug in this code when I tested in staging: if I click the `DownloadData` button, the website crashes, showing this error:
Screenshot 2021-08-30 at 10 59 42

ivababukova: When you fix it, make sure you also include a unit test to test for this bug and prevent us from having it again.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const getQcHasRun = () => (
+    experimentId
+    && (backendStatus[experimentId]?.status.pipeline?.status === pipelineStatus.SUCCEEDED)
+  );
+
+  const getGem2sHasRun = () => (
+    experimentId
+    && (backendStatus[experimentId]?.status?.gem2s?.status === pipelineStatus.SUCCEEDED)
+  );
```
marcellp: These should be in a `useState` and `useEffect` should update the state on changes to `backendStatus` instead of the current operation -- this causes unnecessary rerenders as is.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,171 @@
...
+    // project AND all samples in the project have been analysed.
+    if (!activeProject?.samples?.length) {
+      return false;
+    }
+    const steps = Object.values(_.omit(experimentSettings?.processing, ['meta']));
+
+    return steps.length > 0
+      // eslint-disable-next-line no-prototype-builtins
+      && activeProject?.samples?.every((s) => steps[0].hasOwnProperty(s));
+  };
```
marcellp: This should be an effect of `activeProject` and `allSamplesAnalyzed` should be a state variable.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const gem2sHasRun = getGem2sHasRun();
+  const qcHasRun = getQcHasRun();
+  const allSamplesAnalysed = getAllSamplesAnalysed();
```
marcellp: This would be no longer necessary as all of them are going to be state variables.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,171 @@
...
+
+  const downloadExperimentData = async (type) => {
+    try {
+      if (!experimentId) throw new Error('No experimentId specified');
+      if (!Object.values(downloadTypes).includes(type)) throw new Error('Invalid download type');
```
marcellp: `downloadTypes` should be a `Set` so you can run `Set.has` on it in O(1) instead of O(n) for `Array.includes`

src/components/data-management/SamplesTable.jsx
```diff
@@ -0,0 +1,120 @@
+/* eslint-disable import/no-unresolved */
+/* eslint-disable react/jsx-props-no-spreading */
+import React, { useEffect, useState } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import PropTypes from 'prop-types';
+import { Table, Row, Col } from 'antd';
+import { DEFAULT_NA } from 'redux/reducers/projects/initialState';
+import { updateExperiment } from 'redux/actions/experiments';
+import { updateProject } from 'redux/actions/projects';
+import arrayMove from 'array-move';
```
marcellp: Run

`npm update array-move`

then change import here to 

`import {arrayMoveImmutable} from 'array-move';`

src/components/data-management/SamplesTable.jsx
```diff
@@ -0,0 +1,120 @@
...
+    setTableData(newData);
+  }, [projects, samples, activeProjectUuid]);
+
+  const onSortEnd = ({ oldIndex, newIndex }) => {
+    if (oldIndex !== newIndex) {
+      // This can be done because there is only one experiment per project
+      // Has to be changed when we support multiple experiments per project
+      const experimentId = activeProject.experiments[0];
+
+      const newData = arrayMove([].concat(tableData), oldIndex, newIndex).filter((el) => !!el);
```
marcellp: after the new version of `array-move`, this can now call `arrayMoveImmutable(tableData, ...)` without cloning the object.

src/components/data-management/DownloadData.jsx
```diff
@@ -0,0 +1,170 @@
...
+import { saveAs } from 'file-saver';
+import downloadTypes from 'utils/data-management/downloadTypes';
+import { getFromApiExpectOK } from 'utils/getDataExpectOK';
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from 'utils/endUserMessages';
+import pipelineStatus from '../../utils/pipelineStatusValues';
+import { exportQCParameters, filterQCParameters } from '../../utils/data-management/exportQCParameters';
+import { loadBackendStatus } from '../../redux/actions/backendStatus/index';
+
+const DownloadData = (props) => {
```
ivababukova, line 19: I don't see any unit tests for this component

ivababukova: we can add them as part of your next PR

marcellp (approved):

src/__test__/components/data-management/ProjectDetails.test.jsx
cosa65: weird, I guess we can look at it later since merging this PR is more important now

src/components/data-management/SamplesTableCells.jsx
cosa65: I was thinking that moving these 4 new files into a new folder would deal with that problem but it is not that important, you so leaving as it is is fine too.

cosa65:
> we already have them defined correctly so what's the point? I was thinking to use this in further imports you make so its easier

Whenever we move a file we need to adapt all of the relative imports defined in it, which means extra work whenever we reorganize directories

PR #460, opened 2021-08-23
Refactor work loop

marcellp:
# Background
#### Link to issue
https://biomage.atlassian.net/browse/BIOMAGE-1369

https://ui-new-worker-loop.scp-staging.biomage.net/

https://github.com/biomage-ltd/worker/pull/185
https://github.com/biomage-ltd/api/pull/210

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/work/seekWorkResponse.js
```diff
@@ -0,0 +1,101 @@
...
+  if (!storageResp.ok) {
+    return null;
+  }
+
+  const response = await storageResp.json();
+
+  return response;
+};
+
+const seekFromAPI = async (
```
cosa65, line 28: I think I would call this `seekWorkResponse` and export it as default since it actually does all of the work here, including `seekFromS3`

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+    return response;
+  }
+
+  const responseData = JSON.parse(response.results[0].body);
+
+  if (response.response?.cacheable) {
+    await cache.set(ETag, responseData);
+  }
+
+  console.log('finally', responseData);
```
cosa65: Remove this

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
+/* eslint-disable no-underscore-dangle */
```
marcellp, line 1: important change

src/utils/work/seekWorkResponse.js
```diff
@@ -0,0 +1,101 @@
+import moment from 'moment';
```
marcellp, line 1: important change

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+    return cachedData;
+  }
+
+  // If new genes are needed, construct payload, try S3 for results,
+  // and send out to worker if there's a miss.
+  const { pipeline: { startDate: qcPipelineStartDate } } = backendStatus;
+
+  const missingGenesBody = { ...body, genes: missingGenes };
+  const ETag = createObjectHash({
+    experimentId, missingGenesBody, qcPipelineStartDate, extras, lol: 'tgv 23f23f23222222gv11111fsdfwegwegwe',
```
marcellp: remove this magic value

src/redux/reducers/cellSets/cellSetsError.js
```diff
@@ -6,7 +6,7 @@ const cellSetsError = (state, action) => {
   return {
     ...initialState,
     loading: false,
-    error,
+    error: error || true,
```
cosa65: Could we use `??` instead?

src/redux/actions/cellSets/runCellSetsClustering.js
```diff
@@ -36,7 +36,20 @@ const runCellSetsClustering = (experimentId, resolution) => async (dispatch, get
...
+        if (err) {
+          throw err;
+        }
+
+        const louvainSets = JSON.parse(res.results[0].body);
+        const newCellSets = [
+          louvainSets,
+        ];
+        dispatch(updateCellSetsClustering(experimentId, newCellSets));
+      },
```
cosa65: Why is this `eventCallback` in the ui that triggered this request? We want this logic to also be present in the other ui's that are currently browsing this experiment, right? (otherwise we could run into problems where one ui has an older louvain clusters version and it renames a cluster or something and so it overwrites the new louvain clusters with its old version).

.github/workflows/ci.yaml
```diff
@@ -30,15 +30,16 @@ jobs:
...
-    - id: test
-      name: Run unit tests
-      uses: mattallty/jest-github-action@v1
-      env:
-        AWS_DEFAULT_REGION: 'eu-west-1'
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      with:
-        test-command: "npm run coverage"
-        coverage-comment: false
+    # TODO: this needs to be re-enabled before merging and tests fixed
```
ivababukova: don't forget about this

package.json
```diff
@@ -87,6 +87,7 @@
     "null-loader": "^3.0.0",
     "object-hash": "^2.0.3",
     "optimize-css-assets-webpack-plugin": "^5.0.3",
+    "p-timeout": "^5.0.0",
```
ivababukova: did you actually end up using this package? Because I can't see it being imported anywhere

package.json
```diff
@@ -19,7 +19,7 @@
     "build": "next build",
     "prod": "NODE_ENV=production node server.js",
     "pretest": "eslint ./src",
-    "test": "jest",
+    "test": "jest --silent",
```
ivababukova, line 22: what does this flag do?

src/__test__/utils/cacheRequest.test.js
```diff
@@ -1,5 +1,5 @@
 import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
```
ivababukova, line 1: I would also rename the test file to be called `fetchWork.test.js`

ivababukova: and create a `work` folder in the test utils folder to replicate the non-tests structure

src/__test__/utils/sendWork.test.js
```diff
@@ -1,6 +1,6 @@
 import { v4 as uuidv4 } from 'uuid';
```
ivababukova, line 1: Similarly, I would rename this test file to be called `seekFromAPI.test.js` and I will put it in that `work` folder in `__test__/utils`

src/__test__/utils/sendWork.test.js
```diff
@@ -29,7 +29,20 @@ jest.mock('../../utils/socketConnection', () => {
   };
 });
 
-describe('sendWork unit tests', () => {
+jest.mock('aws-amplify', () => ({
+  configure: jest.fn().mockImplementation(() => ({
+    Storage: {
+      AWSS3: {
+        bucket: 'biomage-originals-test',
```
ivababukova, line 38: how about we put this bucket name in config? And in here, we get the config value for it instead of hardcoding it?

src/__test__/utils/sendWork.test.js
```diff
@@ -64,12 +76,12 @@ describe('sendWork unit tests', () => {
...
   it('Sends work to the backend when called and returns valid response.', async () => {
-    const response = await sendWork(
-      experimentId, timeout, body,
+    const response = await seekFromAPI(
+      experimentId, body, timeout, 'badbeef',
     );
 
     expect(socketConnectionMocks.mockEmit).toHaveBeenCalledWith('WorkRequest', {
-      uuid: 'my-random-uuid',
+      ETag: 'badbeef',
```
ivababukova: why `badbeef` when everything in this test is supposed to work without errors?

src/components/ContentWrapper.jsx
```diff
@@ -68,30 +68,31 @@ const ContentWrapper = (props) => {
...
     }
 
     dispatch(loadBackendStatus(experimentId));
 
-    setUpConnection();
+    (async () => {
+      const io = await connectionPromise;
+      const cb = experimentUpdatesHandler(dispatch);
+
+      io.off();
```
ivababukova, line 95: why first off and then on?

src/components/ContentWrapper.jsx
```diff
@@ -68,30 +68,31 @@ const ContentWrapper = (props) => {
...
-
   useEffect(() => {
     if (!experimentId) {
       return;
     }
 
     dispatch(loadBackendStatus(experimentId));
 
-    setUpConnection();
+    (async () => {
```
ivababukova, line 89: this is subjective, but it seemed neater to me when it was written as a `setUpConnection` function. Are there any particular reasons to write it as an anonymous function?

src/components/FocusButton.jsx
```diff
@@ -30,7 +30,7 @@ const FocusButton = (props) => {
   const focused = focusData.store === store && focusData.key === lookupKey;
 
   return (
-    <Tooltip placement='right' title={`${(focused) ? 'Hide color on' : 'Show color on'} embedding`}>
+    <Tooltip placement='right' title={`${(focused) ? 'Hide color on' : 'Show color on'} embedding`} destroyTooltipOnHide>
```
ivababukova, line 33: I assume that you added `destroyTooltipOnHide` to improve the performance? Is this something that we need to be adding for every tooltip?

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
```diff
@@ -94,45 +98,49 @@ const CellSetsTool = (props) => {
...
+    setNumSelected(numSelectedUnfiltered.size);
+  }, [activeTab, allSelected]);
+
+  const onNodeUpdate = useCallback((key, data) => {
     dispatch(updateCellSetProperty(experimentId, key, data));
-  };
+  }, []);
 
-  const onNodeDelete = (key) => {
+  const onNodeDelete = useCallback((key) => {
```
ivababukova, line 124: I didn't know about this `useCallback`. Do you know what was the impact on performance when you added it in? We might want to tell others in the team about it

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
```diff
@@ -231,6 +228,18 @@ const CellSetsTool = (props) => {
     );
   };
 
+  if (loading) return <Skeleton active />;
+  if (!cellSetTreeData || !metadataTreeData) return <Skeleton active />;
```
ivababukova: why not in the same if statement? Also, why putting a skeleton when there is no cell set or metadata data? Wouldn't that confuse the user that something is loading, when actually nothing is getting loaded?

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
```diff
@@ -278,16 +274,27 @@ const HierarchicalTree = (props) => {
...
+  const [renderedTreeData, setRenderedTreeData] = useState([]);
+  useEffect(() => {
+    if (!treeData) {
+      return;
+    }
+
+    setRenderedTreeData(renderTitlesRecursive(treeData));
+  }, [treeData]);
+
+  if (!treeData) return <Skeleton active />;
```
ivababukova, line 287: if there is no treeData, shouldn't this component be an error state, not loading state?

src/utils/work/seekWorkResponse.js
marcellp: That's not correct. `seekFromS3` is used in other places, notably in `fetchWork` when we check S3 before we even go to the API (if the result is in S3, we don't check the API at all).

src/redux/reducers/cellSets/cellSetsError.js
marcellp: I guess we could.

package.json
marcellp: no, we didnt. removing

package.json
marcellp: it suppresses `console` messages in the tests so it's easier to browse through in a terminal with limited buffer. there is also a `test:loud` that makes those messages appear. this is similar in behavior to how the `api` works now.

src/__test__/utils/sendWork.test.js
marcellp: It's a common hexadecimal value used for representing a garbage value. I understand the confusion, I'll just swap it with something else, like `facefeed`.

src/components/ContentWrapper.jsx
marcellp: I added a comment to explain. Basically, it's in order to prevent us from accidentally getting work response messages and experiment updates from a previous experiment that the user browsed to. `io.off` effectively deletes all previous hooks that you may have created, including `WorkRespone` hooks.

src/components/ContentWrapper.jsx
marcellp: Not necessarily, but I feel like it being refactored to a different place is misleading. We are essentially creating a promise by calling an `async` function, but we are not waiting for this promise to be resolved. This is a potentially dangerous operation, and putting the async function inline means people would be less likely to assume that `setUpConnection` is not an async function.

src/components/FocusButton.jsx
marcellp: No, this was in order to prevent the tooltip "sticking" to the button when we hide/show cell sets.

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
marcellp: We talked with Stefan about it when reviewing their PR.

It is a good idea to tell people about it, although it is something that has to be used with care. In this case, this resulted in some performance improvement as a change to the function called a cascading rerender, including a new cell set tree being generated each time. That took a long time, and hurt performance. However, in most cases, `useCallback` can actually cause a performance penalty.

I can explain more on Slack if you want.

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
marcellp: There is not really a reason for not putting them together, but I have found that creating one big `if` statement can make things harder to read. In addition, in the flow, `loading` will be first set to `false`, and **then** we'll have to wait for `cellSetTreeData` and `metadataTreeData` to be defined, so this better reflects the actual steps that go into loading the cell sets imho.

In addition, `cellSetTreeData` and `metadataTreeData` are never `null` in their normal state. They may be an empty array, but that's not a falsy value in Javascript. The only time these values are falsy is when the hook to create `cellSetTreeData` and `metadataTreeData` has not been run yet.

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
marcellp: Similarly to above, no. In that case `CellSetsTool` would pass an empty array, and not a falsy value.

src/redux/actions/cellSets/runCellSetsClustering.js
marcellp: Yes, in fact, we are not meant to have this here at all. I will comment it out and instead add an explanation for how to use it as an example in case someone were to create something that takes an `eventCallback` in the future.

For now, we won't update clustering information for a user if another user updated the clusters for the same experiment. This is because that would actually cause inconsistent behavior, as we do not do this for any other work type either (e.g. for embeddings).

package-lock.json
```diff
@@ -22330,14 +22330,16 @@
...
-      "requires": {
-        "p-finally": "^1.0.0"
+      },
+      "dependencies": {
+        "p-timeout": {
+          "version": "3.2.0",
+          "resolved": "https://registry.npmjs.org/p-timeout/-/p-timeout-3.2.0.tgz",
+          "integrity": "sha512-rhIwUycgwwKcP9yTOOFK/AKsAopjjCakVqLHePO3CC6Mir1Z99xT+R63jZxAT5lFZLa2inS5h+ZS2GvR99/FBg==",
+          "requires": {
+            "p-finally": "^1.0.0"
```
ivababukova, line 22340: is this change still needed?

src/__test__/utils/sendWork.test.js
ivababukova: fair enough

src/components/ContentWrapper.jsx
ivababukova: fair enough, happy to have it like this

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
ivababukova: In terms of the loading explanation: sounds like it is **very important** that loading states are always set to false initially by default, because if we set them to true, this will make a component never load. I am a bit concerned that it is very easy to not think about this and create a new loading state and set it to true by default (I actually saw it in a PR a few months ago). So what can we do to make sure that this never happens?

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+    throw new Error('Disabling network interaction on server');
+  }
+
+  const { pipeline: { startDate: qcPipelineStartDate } } = backendStatus;
+  if (body.name === 'GeneExpression') {
+    return fetchGeneExpressionWork(experimentId, timeout, body, backendStatus, extras);
+  }
+
+  const ETag = createObjectHash({
+    experimentId, body, qcPipelineStartDate, extras, lol: '4234r32f2f32322111112222f',
```
ivababukova: need to remove this amazing lol parameter

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+
+  // First, let's try to fetch this information from the local cache.
+  const data = await cache.get(ETag);
+  if (data) {
+    return data;
+  }
+
+  // Then, we may be able to find this in S3.
+  let response = await seekFromS3(ETag);
+  console.warn('The response from S3 is', response);
```
ivababukova: we don't need this console log anymore

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+
+  // If response cannot be fetched, go to the worker.
+  if (!response) {
+    console.warn('no response in s3 for', body, 'sending to worker');
```
ivababukova: no need for this console log

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+      eventCallback,
+      {
+        PipelineRunETag: qcPipelineStartDate,
+        ...extras,
+      },
+    );
+  }
+
+  if (!response) {
+    console.debug(`No response immediately resolved for ${body} (ETag: ${ETag}) -- this is probably an event subscription.`);
```
ivababukova, line 133: do we still need this console debug message?

src/utils/work/seekWorkResponse.js
```diff
@@ -0,0 +1,101 @@
...
+import WorkResponseError from '../WorkResponseError';
+import getAuthJWT from '../getAuthJWT';
+import WorkTimeoutError from '../WorkTimeoutError';
+
+const seekFromS3 = async (ETag) => {
+  const configuredBucket = Amplify.configure().Storage.AWSS3.bucket;
+  const storageUrl = await Storage.get(
+    ETag,
+    {
+      bucket: configuredBucket.replace('biomage-originals', 'worker-results'),
```
ivababukova, line 14: Can we have a variable in config for the work results bucket and then just use it here instead of this hardcoding?

src/utils/work/fetchWork.js
```diff
@@ -0,0 +1,152 @@
...
+    return response;
```
ivababukova, line 134: if there is no response, aren't we supposed to throw an error?

src/utils/work/seekWorkResponse.js
```diff
@@ -0,0 +1,101 @@
...
+
+  return response;
+};
+
+const seekFromAPI = async (
+  experimentId,
+  body,
+  timeout,
+  ETag,
+  eventCallback,
```
ivababukova, line 33: what is the purpose of the `eventCallback`?

src/utils/work/fetchWork.js
marcellp: fixed

src/utils/work/fetchWork.js
marcellp: fixed

src/utils/work/fetchWork.js
marcellp: fixed

src/utils/work/fetchWork.js
marcellp: I'd keep it there, maybe it will be useful in some cases.

src/utils/work/fetchWork.js
marcellp: No, in this case we do not expect a response, as e.g. in cases like updating clustering the response will be broadcast to all clients, and we automatically subscribe to all future reclustering updates, for example. In this case `fetchWork` behaves more like a callback function than a promise. See `eventCallback` in `seekWorkRespone` and its use in `updateCellSetClustering`.

src/utils/work/seekWorkResponse.js
marcellp: See above.

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
marcellp: Don't see why that would be the case. As it is right now, the component loads if `loading` is `false`, and `cellSetTreeData` is truthy, and `metadataTreeData` is truthy. You can set `loading` to be `true` by default, there's nothing wrong with it, the component will not load until `loading` is `false`.

.github/workflows/ci.yaml
marcellp: removed now that tests pass

src/utils/work/seekWorkResponse.js
marcellp: we can't, as the `configBucket` is derived from AWS so we can't access it from the browswer

package-lock.json
marcellp: no, but not sure why its still there. its probably a cross-dependency with something else. i cant remove it, but its not in `package.json`.

src/components/EditableField.jsx
```diff
@@ -134,7 +134,7 @@ const EditableField = (props) => {
             deleteEnabled
               ? (
                 <Tooltip placement='top' title='Delete' mouseLeaveDelay={0}>
-                  <Button size='small' shape='circle' icon={<DeleteOutlined />} onClick={deleteEditableField} />
+                  <Button aria-label='Delete' size='small' shape='circle' icon={<DeleteOutlined />} onClick={deleteEditableField} />
```
ivababukova, line 137: what are these labels for? Are they for the integration testing?

marcellp: they are for accessibility, so screenreaders can find a label for the button (since the text is just an SVG). they also help locate the item for unit testing, yes

src/__test__/utils/cacheRequest.test.js
marcellp: now fixed

src/__test__/utils/sendWork.test.js
marcellp: now fixed

cosa65 (approved):

ivababukova (approved):

PR #482, opened 2021-08-31
[BIOMAGE-1230] - Add back data test id and check for experiments

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1230

https://ui-agi-ui482.scp-staging.biomage.net/

N.A.

- PR #458 causes `data-test-id` that is attached to the Launch Analysis button to be removed. This PR adds back the `data-test-id` attribute
- This PR also adds a check to the UI that ensures that experiments data relevant to a project are loaded before being able to launch analysis. This prevents an error that comes up when the analysis modal is launched, but the data to render the list is not yet available (loading), causing the error to pop up. This doesn't happen in normal situations because the time it takes for the user user to choose a project and click "Launch Analysis" allows for the experiments data to be loaded. This error shows up in testing because the action is almost spontaneous.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/components/data-management/AnalysisModal.jsx
+++ b/src/components/data-management/AnalysisModal.jsx
@@ -33,7 +33,9 @@ const NewExperimentModal = (props) => {
   const [isWorking, setIsWorking] = useState(false);
 
   useEffect(() => {
-    const updatedList = activeProject?.experiments?.map((experimentId) => experiments[experimentId])
+    if (!activeProject?.experiments?.length > 0) return;
+
+    const updatedList = activeProject.experiments.map((experimentId) => experiments[experimentId])
       .filter((experiment) => experiment !== undefined);
 
     setExperimentsList(updatedList);
@@ -48,7 +50,7 @@ const NewExperimentModal = (props) => {
   ];
 
   const renderAnalysisList = () => {
-    if (experimentsList.length === 0) {
+    if (!experimentsList?.length > 0) {
       return (
         <Row justify='center'>
           <ClipLoader size={30} color='#8f0b10' />
@@ -62,7 +64,6 @@ const NewExperimentModal = (props) => {
         bordered
         dataSource={experimentsList}
         itemLayout='vertical'
-        loading={experimentsList?.length === 0}
         renderItem={(experiment) => (
           <List.Item
             key={`${experiment.id}`}
--- a/src/components/data-management/AnalysisModal.jsx
+++ b/src/components/data-management/AnalysisModal.jsx
@@ -35,7 +35,8 @@ const NewExperimentModal = (props) => {
   useEffect(() => {
     if (!activeProject?.experiments?.length > 0) return;
 
-    const updatedList = activeProject.experiments.map((experimentId) => experiments[experimentId])
+    const updatedList = activeProject.experiments
+      .map((experimentId) => experiments[experimentId])
       .filter((experiment) => experiment !== undefined);
 
     setExperimentsList(updatedList);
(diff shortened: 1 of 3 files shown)


src/components/data-management/AnalysisModal.jsx
```diff
@@ -45,6 +49,81 @@ const NewExperimentModal = (props) => {
...
+                    setNumFieldsEditing(Math.max(0, numFieldsEditing + (editing || -1)));
+                  }}
+                />
+              </strong>
+              <EditableField
+                onAfterSubmit={async (description) => {
+                  dispatch(
+                    updateExperiment(experiment.id, { description: description.trim() }),
+                  );
+                  await dispatch(saveExperiment(experiment.id));
```
robioson: Why await this? If you remove, remember to remove `async` too

src/components/data-management/AnalysisModal.jsx
```diff
@@ -45,6 +49,81 @@ const NewExperimentModal = (props) => {
     rules.ALPHANUM_DASH_SPACE,
   ];
 
+  const renderAnalysisList = () => {
+    if (!experimentsList?.length > 0) {
```
robioson, line 54: I would just check `=== 0` rather than negating a greater than

robioson: Although, if `experimentsList` can be undefined, and it is, this condition would become false instead of true.

I would either remove the optional chain (if it in fact can't be undefined) or do `(experimentsList === undefined || experimentsList.length === 0)`

src/components/data-management/AnalysisModal.jsx
```diff
@@ -32,9 +33,12 @@ const NewExperimentModal = (props) => {
   const [isWorking, setIsWorking] = useState(false);
 
   useEffect(() => {
-    setExperimentsList(
-      activeProject?.experiments?.map((experimentId) => experiments[experimentId]) || [],
-    );
+    if (!activeProject?.experiments?.length > 0) return;
+
+    const updatedList = activeProject.experiments.map((experimentId) => experiments[experimentId])
```
robioson: I would place a newline before `.map` so it and `.filter` are lined up

robioson (approved):
I like this, especially the test. Some small comments.

src/components/data-management/AnalysisModal.jsx
aerlaut: Yes, this test is basically the shorter form of `(experimentsList === undefined || experimentsList.length === 0)`

src/components/data-management/AnalysisModal.jsx
aerlaut: :+1:

src/components/data-management/AnalysisModal.jsx
aerlaut: I'm not sure - it was there before and I didn't add it. I'll check to see what the impact of removing `async/await` have.

aerlaut: There doesn't seem to be any effect, so I'm removing it.

src/components/data-management/AnalysisModal.jsx
robioson: Oh right, didn't realise this was a deliberate JS trick!

PR #488, opened 2021-09-01
[BIOMAGE-1341] Add test data for species test

robioson:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1341
https://ui-ready-heliotrope-woodpecker.scp-staging.biomage.net/
https://github.com/biomage-ltd/testing/pull/15
Also tried to fix the issue whereby the organisms list is requested many times while the platform is open.
I tried using [`useSWRImmutable`](https://swr.vercel.app/docs/revalidation#disable-automatic-revalidations) but this prevented the data being made at all (I don't understand why).
# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-management/SpeciesSelector.jsx
```diff
@@ -21,6 +21,8 @@ const SpeciesSelector = (props) => {
       dropdownMatchSelectWidth={400}
       showSearch
       placeholder='Search for common or scientific name...'
+      data-test-class='species-select'
+      data-test-id={testID}
```
aerlaut: I think it's better to directly use the value `fill-species-select` here instead of passing it via a props. Is the going to change?

robioson: Yes, I need the selector on the fill species box to be different from the ones beside samples. This is ugly so I wouldn't mind a better way to distinguish them

aerlaut: Do you need to distinguish between different species selectors? If you don't then you don't need to use `data-test-id`. Using data-test-class is fine. You can query for `.species-select` then select a specific element using `.eq()`

robioson: Yes I do. I suppose I could try to get the "Fill species" box and get the selector within that. Using `eq` seems super fragile imo

src/components/data-management/ProjectDetails.jsx
```diff
@@ -301,6 +302,7 @@ const ProjectDetails = ({ width, height }) => {
             massEdit
           >
             <SpeciesSelector
+              testID='fill-species-select'
```
aerlaut: We have refactored the test ids into a common file `src/utils/integrationTestConstants.js`. You can see how test id is implemented here : https://github.com/biomage-ltd/ui/blob/master/src/components/data-management/ProjectMenu.jsx#L73

src/components/data-management/SpeciesSelector.jsx
aerlaut: I think using `eq()` to select an element and modify it is not fragile, as long as you can ensure that the correct selecteors are applied and you control the data that is being tested. I use it in `randomizeSampleName()` to select the project name and change it. I see that you are using `Int Test - Medata Project` in your PR:  https://github.com/biomage-ltd/testing/pull/15/files, so I think you can always be confident that the data that is loaded in the project will be there.

aerlaut (changes requested):
Some comments on the UI implementation

src/components/data-management/ProjectDetails.jsx
cosa65: you'll also need to call this attribute `test-data-id`

kafkasl:
closing this PR as the species metadata will be removed from cellenics

PR #490, opened 2021-09-02
[BIOMAGE-1219] - Navigation checks

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1219

https://ui-agi-nav-check-490.scp-staging.biomage.net/

N.A.

- Change mechanism to use Redux because we need to fire events from components coming from different levels of nesting.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


vickymorrison (approved):

src/components/ChangesNotAppliedModal.jsx
```diff
@@ -1,37 +1,51 @@
...
       title='Changes not applied'
-      onCancel={onCancel}
+      onCancel={() => dispatch(navigateFromProcessingTo(''))}
       footer={(
         <Space size='large' style={{ display: 'flex', justifyContent: 'center' }}>
           <Button
             type='primary'
             key='run'
-            onClick={onRun}
+            onClick={() => {
```
cosa65, line 33: Why was this moved into the modal? FilesUploadModal doesn't trigger the upload for example

src/components/ChangesNotAppliedModal.jsx
```diff
@@ -1,37 +1,51 @@
...
+            }}
             style={{ width: '100px' }}
           >
             Run
           </Button>
           <Button
             type='primary'
             key='discard'
-            onClick={onDiscard}
+            onClick={() => {
```
cosa65, line 44: Same with this, was was it moved into the modal?

src/components/ChangesNotAppliedModal.jsx
```diff
@@ -1,37 +1,51 @@
...
+
+  const router = useRouter();
+  const dispatch = useDispatch();
+
+  const { changedQCFilters, navigationPath } = useSelector((state) => state.experimentSettings.processing.meta);
 
   return (
     <Modal
-      visible={visible}
+      visible={navigationPath}
```
cosa65, line 25: Why navigation path is being used in visible?

src/components/ChangesNotAppliedModal.jsx
```diff
@@ -1,37 +1,51 @@
...
 const ChangesNotAppliedModal = (props) => {
-  const {
-    steps, visible, onRun, onDiscard, onCancel,
-  } = props;
+  const { experimentId } = props;
+
+  const router = useRouter();
+  const dispatch = useDispatch();
+
+  const { changedQCFilters, navigationPath } = useSelector((state) => state.experimentSettings.processing.meta);
```
cosa65, line 21: Why do we store navigationPath in redux?

src/redux/actions/experimentSettings/navigateFromProcessingTo.js
```diff
@@ -0,0 +1,10 @@
+import {
+  EXPERIMENT_SETTINGS_NAVIGATE_TO,
```
cosa65, line 2: I don't feel like this fits in experiment settings

src/components/ChangesNotAppliedModal.jsx
aerlaut: Because it has to be shared between multiple components on different hierarchy levels. The link in ContentWrapper needs it, and the "finish QC" button inside the Data Processing component also needs to know about it.

src/components/ChangesNotAppliedModal.jsx
aerlaut: It's a shortcut basically, if it's an empty string the modal doesn't show.

src/components/ChangesNotAppliedModal.jsx
aerlaut: This modal directly listens to the `navigationPath` to get rendered or not. This is so because this modal needs to be called from both the navigation links and also the finish qc button. Because it listens directly to the store, I thought it would be ok for its actions to also be directly related to the store instead of taking functions from its parent component.

src/components/ChangesNotAppliedModal.jsx
aerlaut: Same as above

src/redux/actions/experimentSettings/navigateFromProcessingTo.js
aerlaut: I put it in `experimentSettings` because it's the closest to `changedQCFilters`. I'm not sure where else to put it,

src/components/ChangesNotAppliedModal.jsx
cosa65: The thing is that no other modal is doing this, all of them work through callbacks

src/components/ChangesNotAppliedModal.jsx
cosa65: We should probably use null instead of an empty string, and perhaps to make it more explicit use `navigationPath === null`

src/components/ChangesNotAppliedModal.jsx
cosa65: Perhaps could we wrap this `ChangesNotAppliedModal` in a `NavigationHandler` or something with a better name that handles the navigation requests.

When something dispatches to navigate somewhere it should check, if we currently are in data processing and are trying to go somewhere else with unsaved changes then show modal and handle the callbacks.

cosa65: and move the callbacks back outside, with this navigation component handling those callbacks via dispatches.

src/redux/actions/experimentSettings/navigateFromProcessingTo.js
cosa65: I think it fits at the root level of the store, with a new reducer and all that

cosa65: it's about navigation, so not really linked with any of the already existing aprts of the store.

src/components/ChangesNotAppliedModal.jsx
aerlaut: Hmm... are you suggesting something like a middleware?

aerlaut: Does that mean the other navigation will also route throught this middleware component?

cosa65: something like our notifications manager

PR #491, opened 2021-09-02
[HACKATHON] - Implement project search filter

aerlaut:
# Background
#### Link to issue 

https://trello.com/c/9IJg1qR8/40-add-a-search-bar-to-search-for-project-name-id-in-the-projects-list

N.A.

- Search by project name, project id or experiment id.

# Changes
### Code changes

# Definition of DONE

### Testing
- [x] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


vickymorrison:
The full text in the search bar is not visibile - I suggest shortening the instruction and/or adding a tooltip. Other than that, this new feature is great!

Screenshot 2021-09-03 at 14 04 48

src/__test__/components/data-management/ProjectCard.test.jsx
```diff
@@ -0,0 +1,37 @@
...
+      name: projectName,
+      uuid: projectUuid,
+    },
+  },
+};
+
+const mockStore = configureMockStore([thunk]);
+
+describe('ProjectCard', () => {
+  it('Displays correctly', () => {
```
cosa65, line 38: While we're at it we could add some more tests based on variation of the props

src/components/data-management/ProjectCard.jsx
```diff
@@ -0,0 +1,96 @@
...
+      type='primary'
+      style={isActive ? activeProjectStyle : { cursor: 'pointer' }}
+      onClick={onClick}
+    >
+      <Descriptions
+        layout='horizontal'
+        size='small'
+        column={1}
+      >
+        <Descriptions.Item contentStyle={{ fontWeight: 700, fontSize: 16 }}>
```
cosa65: You could extract the const {Item} = Descriptions, right? Might make this a little bit more readable

src/components/data-management/ProjectsListContainer.jsx
```diff
@@ -1,146 +1,166 @@
...
+    // If UUID is prefixed with sandbox_id, remove prefix
+    const projectUuid = substrings.length > 5 ? substrings.slice(-5).join('-') : uuid;
+
+    return validate(projectUuid);
   };
 
-  const deleteProject = () => {
-    dispatch(deleteProjectAction(deleteProjectUuid));
-    setDeleteModalVisible(false);
+  // const experimentsAreLoaded = (project, experiments) => {}
```
cosa65: Remove this

src/components/data-management/ProjectsListContainer.jsx
```diff
@@ -1,146 +1,166 @@
...
+  const [newProjectModalVisible, setNewProjectModalVisible] = useState(false);
+  const [justLoggedIn, setJustLoggedIn] = useState(true);
+  const activeProject = projectsList[activeProjectUuid];
+
+  const existingExperiments = activeProject?.experiments
+    .map((experimentId) => experiments[experimentId]);
+
+  const experimentIds = new Set(experiments.ids);
+  const experimentsAreLoaded = activeProject?.experiments
+    .every((experimentId) => experimentIds.has(experimentId));
```
cosa65: I feel like these shouldn't be constants that are recalculated on every render, I would move them to functions or even better, if they are only used in one useEffect then inside those

src/components/data-management/ProjectsListContainer.jsx
```diff
@@ -1,146 +1,166 @@
...
-    backgroundColor: blue[0],
-    cursor: 'pointer',
-    border: `2px solid ${blue.primary}`,
+  const experiments = useSelector((state) => state.experiments);
+  const [newProjectModalVisible, setNewProjectModalVisible] = useState(false);
+  const [justLoggedIn, setJustLoggedIn] = useState(true);
+  const activeProject = projectsList[activeProjectUuid];
+
+  const existingExperiments = activeProject?.experiments
+    .map((experimentId) => experiments[experimentId]);
```
cosa65: Move this inside `createNewProject`

src/components/data-management/ProjectsListContainer.jsx
```diff
@@ -1,146 +1,166 @@
...
+  } = useSelector((state) => state.samples.meta);
 
-  const activeProjectStyle = {
-    backgroundColor: blue[0],
-    cursor: 'pointer',
-    border: `2px solid ${blue.primary}`,
+  const experiments = useSelector((state) => state.experiments);
+  const [newProjectModalVisible, setNewProjectModalVisible] = useState(false);
+  const [justLoggedIn, setJustLoggedIn] = useState(true);
+  const activeProject = projectsList[activeProjectUuid];
```
cosa65: After moving the other stuff you can check if this can be moved too

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,114 @@
...
+
+const validationChecks = [
+  rules.MIN_8_CHARS,
+  rules.MIN_2_SEQUENTIAL_CHARS,
+  rules.ALPHANUM_DASH_SPACE,
+  rules.UNIQUE_NAME_CASE_INSENSITIVE,
+];
+
+const ProjectsListContainer = (props) => {
+  const { height, filter } = props;
```
cosa65, line 13: Is there some way we can find out the `height` without receiving it from outside?

src/components/data-management/ProjectsListContainer.jsx
marcellp: I believe @ivababukova has fixes that clears these up.

src/components/data-management/ProjectsListContainer.jsx
marcellp: I believe @ivababukova has fixes that clears these up.

src/components/data-management/ProjectsListContainer.jsx
marcellp: I believe @ivababukova has fixes that clears these up.

marcellp (approved):

marcellp (dismissed):

marcellp (dismissed):
I think this should wait until https://github.com/biomage-ltd/ui/pull/494 by @ivababukova is approved and merged. You will potentially waste a lot of time on feedback on code that is actually being fixed up in that other PR and you probably will need to do some conflict resolution after that's merged in.

Other than that, the code looks good.

cosa65:
> I think this should wait until #494 by @ivababukova is approved and merged. You will potentially waste a lot of time on feedback on code that is actually being fixed up in that other PR and you probably will need to do some conflict resolution after that's merged in.
> 
> Other than that, the code looks good.

Good point, @aerlaut let me kno when you need a new code review for this PR

src/__test__/components/data-management/ProjectCard.test.jsx
aerlaut: Hmm.. I don't think there's much else to check. I added checks for the remaining attributes to be shown.

src/components/data-management/ProjectCard.jsx
aerlaut: Done 👍

src/components/data-management/ProjectsListContainer.jsx
aerlaut: no more

src/components/data-management/ProjectsList.jsx
aerlaut: I had a quick search on this matter. It seems that we may be able to use a `ref` that points to an HTML element that's rendered, and get the `ref.current.offsetWidth` value to get the width. I tried this on the elements rendered by this function, but it returns undefined. According to [this comment](https://stackoverflow.com/a/64814269/1940886) on StackOverflow, using the above method we may be able to get the initial width of the object. However, to listen we still need to set up a resize listener. We currently do this using the `ReactResizeObserver` HOC component, which is the source of the `height` props seen here.

src/components/data-management/ProjectsListContainer.jsx
```diff
@@ -1,153 +1,47 @@
...
-    rules.UNIQUE_NAME_CASE_INSENSITIVE,
-  ];
+  const [filterParam, setFilterParam] = useState('');
 
-  const validationParams = {
-    existingNames: projectNames,
-  };
+  const debouncedSetFilterParam = useCallback(
+    _.debounce((value) => {
+      setFilterParam(new RegExp(value, 'ig'));
```
cosa65: what's 'ig'?

aerlaut: It's the flags for Regex : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/RegExp

i : case insensitive search
g : global search - find all matches instead of stopping at the first one

cosa65 (approved):

src/components/data-management/ProjectsListContainer.jsx
aerlaut: I think the `g` flag can be removed as we only need the first hit, not every occurence of the string. It'll help speed up the searching. I'll change this.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,115 @@
...
+                  setDeleteProjectUuid(project.uuid);
+                  setDeleteModalVisible(true);
+                }}
+                validationFunc={
+                  (newName) => validateInputs(
+                    newName,
+                    validationChecks,
+                    validationParams,
+                  ).isValid
+                }
```
ivababukova: Here you send different variations of `project.uuid` as props, which is not great because of multiple reasons:
- Creates unnecessary coupling between `ProjectsList` and `ProjectCard`, now `ProjectsList` has to know some internal information about how exactly and what `ProjectCard` needs in order to compute stuff.
- Makes `ProjectCard` component harder to use in other places -- for any other place that needs to use `ProjectCard`, that place will need to work out and pass over all these things.
- Code is hard to read and hard to maintain.

Very, very visible example is the `validateInputs` function: so that you use it in `ProjectCard`, you have to now import it in `ProjectsList`. Why? It will only make `ProjectsList` become our next `ProjectDetails` monster.

You can easily remove all these props and have that handled internally from within `ProjectCard`.

src/components/data-management/ProjectCard.jsx
```diff
@@ -0,0 +1,98 @@
...
+
+const { Item } = Descriptions;
+
+const activeProjectStyle = {
+  backgroundColor: blue[0],
+  cursor: 'pointer',
+  border: `2px solid ${blue.primary}`,
+};
+
+const ProjectCard = (props) => {
```
ivababukova, line 36: Can you settle on one convention and make sure that all component names follow it? Are we going to use Project**s** for all component names that are about multiple projects? Then rename this and every other component to be `Projects<blablabal>`. If not, then again make sure all components are consistent.

src/components/data-management/ProjectCard.jsx
```diff
@@ -0,0 +1,98 @@
...
+  } = props;
+
+  const project = useSelector((state) => state.projects[uuid]);
+
+  return (
+    <Card
+      data-test-class='project-card'
+      key={uuid}
+      type='primary'
+      style={isActive ? activeProjectStyle : { cursor: 'pointer' }}
```
ivababukova: It will make the code cleaner I feel if you also have a variable for the style of the inactive project and use it here similarly to `activeProjectStyle`:

```
const inactiveProjectStyle = {
    cursor: 'pointer'
}

....

style={isActive ? activeProjectStyle : inactiveProjectStyle}
```

src/__test__/components/data-management/ProjectCard.test.jsx
```diff
@@ -0,0 +1,59 @@
...
+      createdDate,
+      lastAnalyzed,
+      lastModified,
+    },
+  },
+};
+
+const mockStore = configureMockStore([thunk]);
+
+describe('ProjectCard', () => {
```
ivababukova, line 37: after you move the functionality of `ProjectCard` to be within the component itself, you will need to move the unit tests from `ProjectsList` to here.

src/__test__/components/data-management/ProjectsList.test.jsx
```diff
@@ -0,0 +1,122 @@
...
+    meta: {
+      activeProjectUuid: project1.uuid,
+    },
+    [project1.uuid]: project1,
+    [project2.uuid]: project2,
+    [project3.uuid]: project3,
+  },
+});
+
+describe('ProjectsList', () => {
```
ivababukova, line 91: I don't see any unit tests to test for the functionality of the `matcher` function.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,115 @@
...
+  }, [projects.ids]);
+
+  const activeProjectUuid = useSelector((state) => state.projects.meta.activeProjectUuid);
+
+  const deleteProject = () => {
+    dispatch(deleteProjectAction(deleteProjectUuid));
+    setDeleteModalVisible(false);
+  };
+
+  const matcher = (searchObj, filterRegex) => searchObj.projectName.match(filterRegex)
```
ivababukova: Rename this -- throughout the code, in the tests we already use this name `matcher` and it is also a package. I would call the function something like `filterByRegex`. Also, what is this `searchObj`? Is this the list of all projects? I think we need rename that too, it is confusing.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,115 @@
...
+      || searchObj.projectUuid.match(filterRegex);
+
+  return (
+    <>
+      {deleteModalVisible && (
+        <ProjectDeleteModal
+          visible={deleteModalVisible}
+          onCancel={() => { setDeleteModalVisible(false); }}
+          onDelete={deleteProject}
+          projectName={projects[deleteProjectUuid]?.name}
```
ivababukova: same here as for the tons of props you are passing in `ProjectCard`  -- you you really need to have these as props to `ProjectDeleteModal`?

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,115 @@
...
+      projectUuid,
+      projectName: project.name,
+      experiments: project.experiments,
+    };
+
+    return [
+      searchObj,
+      project,
+    ];
+  });
```
ivababukova: what's the point of this function `searchable`? I am really confused here.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,115 @@
...
+          onCancel={() => { setDeleteModalVisible(false); }}
+          onDelete={deleteProject}
+          projectName={projects[deleteProjectUuid]?.name}
+        />
+      )}
+      <Space direction='vertical' style={{ width: '100%', height: height - 90 }}>
+        {
+          searchable
+            .filter(([searchObj]) => matcher(searchObj, filter))
+            .map(([_, project]) => (
```
ivababukova: you have 1 loop inside this `searchable` function, now you have another loop. Is that necessary? Can you try to avoid adding all those loops?

src/components/data-management/ProjectCard.jsx
aerlaut: I was differentiating it depending on whether the component relates to a single project (e.g. **Project**Card) or multiple projects (e.g. **Projects**List). However, in the interest of clarity, I think it's better to use **Projects** for the prefix of all components, to have it in line with the other parts too.

ivababukova: Ok, so there was a reasoning behind that differentiating, good, this is all I was asking for. Then I would leave it to `ProjectCard` if it applies to a single project

aerlaut: On 2nd thought, I think the current naming is good. This is called `ProjectCard` because it contains only 1 project, not multiple project.

`ProjectsList` is a list of projects
`ProjectsListContainer` is the container for `ProjectsList`.

src/components/data-management/ProjectsList.jsx
aerlaut: This is reduced to a variable that stores the value of comparisons instead of a function

src/components/data-management/ProjectsList.jsx
aerlaut: It was made to make it easier to search. It's removed in latest refactor.

src/components/data-management/ProjectsList.jsx
aerlaut: Yes, refactored.

src/__test__/components/data-management/ProjectsList.test.jsx
aerlaut: It's the test on line 110

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+          const matchFilter = project.name.match(filter)
+              || project.experiments.some((experimentId) => experimentId.match(filter))
+              || projectUuid.match(filter);
+
+          if (!matchFilter) return <></>;
+
+          return (
+            <ProjectCard
+              key={project.uuid}
```
ivababukova, line 35: this `key` is not being used in the `ProjectCard` component -- I think you forgot to delete it. Also, make sure you add a key to each Card rendered within ProjectCard, because right now when I open the code locally, I can see a warning message about `Warning: Each child in a list should have a unique "key" prop.`

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+import { Space } from 'antd';
+
+import ProjectCard from './ProjectCard';
+
+const ProjectsListContainer = (props) => {
```
ivababukova: rename this to `ProjectsList` -- the components names have to match the file names.

ivababukova: Just noticed that we have this `ProjectsList` component defined both under `ProjectsListContainer.jsx` and `ProjectsList.jsx` files. This must be wrong, make sure you delete whichever of them is not used.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,46 @@
...
+          );
+        })
+      }
+    </Space>
+  );
+};
+
+ProjectsListContainer.propTypes = {
+  height: PropTypes.number,
+  filter: PropTypes.object.isRequired,
```
ivababukova, line 47: This is not right -- I just checked index.jsx page, you are not passing `filter` in there. Either correct index.jsx, or mark this prop as not required and make sure the code won't fail if it is missing.

src/components/data-management/ProjectsList.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+import ProjectCard from './ProjectCard';
+
+const ProjectsListContainer = (props) => {
+  const { height, filter } = props;
+
+  const projects = useSelector((state) => state.projects);
+
+  return (
+    <Space direction='vertical' style={{ width: '100%', height: height - 90 }}>
```
ivababukova: I just checked what happens if we don't define height and guess -- nothing. The component looks the same. So you can remove all these height passings happening in this component and also from the other `ProjectsListContainer` component from from index.jsx when you call it and this should remove some extra unneeded code.

src/components/data-management/ProjectsList.jsx
ivababukova: Actually I worked it out, there are two components called `ProjectsListContainer`, which you will separately need to fix.

src/components/data-management/ProjectCard.jsx
```diff
@@ -0,0 +1,136 @@
...
+            labelStyle={{ fontWeight: 'bold' }}
+            label='Modified'
+          >
+            <PrettyTime isoTime={project.lastModified} />
+
+          </Item>
+          <Item
+            labelStyle={{ fontWeight: 'bold' }}
+            label='Last analyzed'
+          >
```
ivababukova: instead of saying `labelStyle={{ fontWeight: 'bold' }}` could you not use styled components: https://styled-components.com/ and create a new `BoldItem` component (at the top of this file since it is internal) and use it here instead of `Item`?

src/__test__/components/data-management/ProjectsList.test.jsx
```diff
@@ -0,0 +1,122 @@
...
+      <Provider store={filledStore}>
+        <ProjectsList />
+      </Provider>,
+    );
+
+    // Expect all projects to be shown
+    expect(component.find(ProjectCard).length).toEqual(projects.length);
+  });
+
+  it('Filters the correct project given a filter', () => {
```
ivababukova, line 145: sounds great, but we also need some more tests on what will happen if: we there was nothing matched (so there are 0 results) and what if we pass in none instead of filter, also case sensitivity.

ivababukova: there might be more cases too -- have you seen if the coverage report shows whether there are any of the lines not covered by unit tests and which ones?

ivababukova (changes requested):
way better, but there is still some small things to fix

src/components/data-management/ProjectsList.jsx
aerlaut: Refactored

src/components/data-management/ProjectsList.jsx
aerlaut: Refactored into `ProjectCard`

src/components/data-management/ProjectsList.jsx
aerlaut: The error is actually because of line 25 in `ProjectsListContainer`. The `Skeleton`s doesn't have keys.  I've added them in fbbbaa4

src/components/data-management/ProjectsList.jsx
aerlaut: Whoops, my bad. fixed

src/components/data-management/ProjectsList.jsx
aerlaut: The height here is added to constraint the height of the Space element and force it to show a scroll bar. If we remove height, when you add more projects, the scrollbar will not appear as the element's height expands to accommodate the projects, so you can not scroll to select the projects below.

ivababukova: Ok, you are right, we can't remove it. But then, why `height: height - 90`? Why not `height: height - 10`? Or `-5`?

src/components/data-management/ProjectsList.jsx
ivababukova: great!

src/components/data-management/ProjectCard.jsx
aerlaut: I've tried using Styled components but the styles aren't transferred to the Item texts. So, I refactored the style out instead.

src/components/data-management/ProjectsList.jsx
aerlaut: This is addressed in commit dd8faee. The value is changed to 130. The value comes from adding padding, button size and gap between buttons in the UI. The breakdown can be seen in `ProjectsList` line 8-16

src/__test__/components/data-management/ProjectsList.test.jsx
aerlaut: Additional cases added, including checking for searches with `projectUuid` and `experimentId`

src/components/data-management/ProjectCard.jsx
ivababukova: ok, sounds good

ivababukova (approved):

PR #499, opened 2021-09-07
[HACKATHON] Server-side handling of backend status

marcellp:
# Background
#### Link to issue
N/A 

#### Links to any Pull Requests related to this
N/A

#### Anything else the reviewers should know about the changes here
N/A

# Changes
### Code changes
This is a simple fix that ensures we fetch backend status information during initial render from the server instead of the client. It simplifies code and it also means we do not have to have a skeleton state between launching analysis and starting data processing, breaking immersion.

# Definition of DONE
Your changes will be ready for merging after each of the steps below have been completed:

### Testing
- [x] Unit tests written
- [x] Tested locally with Inframock (with latest production data downloaded with biomage experiment pull)
- [ ] Deployed to staging
- [ ] Passed integration tests 

To set up easy local testing with inframock, follow the instructions here: https://github.com/biomage-ltd/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/biomage-ltd/biomage-utils

### Documentation updates
Is all relevant documentation updated to reflect the proposed changes in this PR?

- [ ] Relevant Github READMEs updated
- [ ] Relevant wiki pages created/updated

### Approvers
- [ ] Approved by a member of the core engineering team
- [ ] (UX changes) Approved by vickymorrison (this is her username, tag her if you need approval)

### Just before merging:
- [ ] After the PR is approved, the `unstage` script in here: https://github.com/biomage-ltd/biomage-utils is executed. This script cleans up your deployment to staging

### Optional
- [ ] Photo of a cute animal attached to this PR

PR #506, opened 2021-09-14
[BIOMAGE-1383] UI shows old backend status state before rerunning gem2s.

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1383

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


aerlaut:
This might be fixed by #499

StefanBabukov:
Yeah i branched off from that PR. There were further changes needed though:
API: before we checked if the gem2s parameters are changed when we run createGem2s with `shouldRunGem2s`. I moved that to the pipeline-status endpoint so when we get the backend status we may return "NEEDS_RERUN", which triggers gem2s.

UI: Added a `deleteBackendStatus` reducer to delete the status when you remove a project. Further changes are to where `runGem2s` is called - we need to call it in `ContentWrapper` with a condition based on the status. So now we don't run gem2s everytime and check if it needs to be run in the API, but we do that in the UI based on the backend status.

src/redux/actions/projects/deleteProject.js
```diff
@@ -71,6 +71,10 @@ const deleteProject = (
       },
     });
 
+    dispatch({
+      type: BACKEND_STATUS_DELETED,
```
cosa65: You're only using `BACKEND_STATUS_DELETED` to react to an experiment getting deleted, right? I think you could reuse the `EXPERIMENTS_DELETED` action, look into this: https://redux.js.org/style-guide/style-guide#allow-many-reducers-to-respond-to-the-same-action.

Basically you can just add in backend status a reducer for EXPERIMENTS_DELETED, that way you can get rid of `BACKEND_STATUS_DELETED`.

There is a good example of this in `EXPERIMENT_SETTINGS_PIPELINE_START`, you can see it is handled across a lot of small reducers. 

Ideally we should move away from dispatching many actions sequentially for one action creator. I think the problem is that we were creating and using redux actions to modify the store while we should instead have think of the store as reacting as a response to actions.

src/redux/actions/pipeline/runGem2s.js
```diff
@@ -7,24 +8,26 @@ import {
...
+  const runningStatus = {
+    ...backendStatus[experimentId].status,
+    gem2s: {
+      status: pipelineStatus.RUNNING,
+      completedSteps: [],
+    },
+  };
   dispatch({
-    type: BACKEND_STATUS_LOADING,
+    type: BACKEND_STATUS_LOADED,
```
cosa65, line 25: Similar thing here, we could have the backend status reducer react to EXPERIMENT_SETTINGS_PIPELINE_START (or EXPERIMENT_SETTINGS_GEM2S_START as a new action, I am not sure what is the best option here) and as a consequence set itself to the status you want in the reducer.

Moving this logic to a reducer will also help with this https://redux.js.org/style-guide/style-guide#put-as-much-logic-as-possible-in-reducers

src/components/data-management/AnalysisModal.jsx
```diff
@@ -48,15 +44,13 @@ const AnalysisModal = (props) => {
...
     onLaunch(experimentId);
-    const analysisPath = '/experiments/[experimentId]/data-processing';
     const lastViewed = moment().toISOString();
     dispatch(updateExperiment(experimentId, { lastViewed }));
     dispatch(updateProject(activeProjectUuid, { lastAnalyzed: lastViewed }));
-
-    dispatch(runGem2s(experimentId));
-    router.push(analysisPath.replace('[experimentId]', experimentId));
+    const analysisPath = '/experiments/[experimentId]/data-processing';
+    window.location.href = await analysisPath.replace('[experimentId]', experimentId);
```
cosa65, line 53: Why use `window.location.href` and make it async instead of using the router?

src/components/GEM2SLoadingScreen.jsx
```diff
@@ -46,6 +46,14 @@ const GEM2SLoadingScreen = (props) => {
       image: '/undraw_Dev_focus_re_6iwt.svg',
       alt: 'A woman working in front of a computer.',
     },
+    needsRerun: {
```
cosa65: I couldn't find where this is being used, was it missing?

src/components/ContentWrapper.jsx
```diff
@@ -96,15 +86,12 @@ const ContentWrapper = (props) => {
 
       io.on(`ExperimentUpdates-${experimentId}`, (update) => cb(experimentId, update));
     })();
-  }, [experimentId]);
-
-  useEffect(() => {
-    if (backendStatusRequested) {
-      return;
+    if ((![RUNNING, SUCCEEDED].includes(gem2sStatusKey)
```
cosa65, line 89: Checking whether we can re run or not gem2s or the qc pipeline is handled in the api, I don't think we need to add more checks here too, just to keep it simple.

src/components/ContentWrapper.jsx
```diff
@@ -51,41 +50,32 @@ const ContentWrapper = (props) => {
...
 
   const changedQCFilters = useSelector(
     (state) => state.experimentSettings.processing.meta.changedQCFilters,
   );
 
-  // This is used to prevent a race condition where the page would start loading immediately
-  // when the backend status was previously loaded. In that case, `backendLoading` is `false`
-  // and would be set to true only in the `loadBackendStatus` action, the time between the
-  // two events would allow pages to load.
-  const [backendStatusRequested, setBackendStatusRequested] = useState(false);
```
cosa65, line 79: Why was this removed?

src/redux/reducers/backendStatus/index.js
```diff
@@ -21,6 +25,18 @@ const backendStatusReducer = (state = initialState, action) => {
     case BACKEND_STATUS_ERROR: {
       return backendStatusError(state, action);
     }
+    case BACKEND_STATUS_DELETED: {
+      return backendStatusDeleted(state, action);
+    }
+    case HYDRATE: {
```
cosa65, line 30: Out of ignorance, what is this used for?

src/utils/pipelineStatusValues.js
```diff
@@ -8,7 +8,7 @@ export default {
   TIMED_OUT: 'TIMED_OUT',
   ABORTED: 'ABORTED',
   SUCCEEDED: 'SUCCEEDED',
-
+  NEEDS_RERUN: 'NEEDS_RERUN',
```
cosa65, line 11: We won't need this if we remove the checks for gem2s in the ui, right?

src/redux/actions/backendStatus/loadBackendStatus.js
```diff
@@ -17,7 +17,9 @@ const loadBackendStatus = (experimentId) => async (dispatch) => {
 
   const url = `/v1/experiments/${experimentId}/backendStatus`;
   try {
-    const response = await fetchAPI(url);
+    const response = await fetchAPI(url, {}, {
+      uiUrl,
```
cosa65, line 21: What is this being used for?

src/redux/actions/projects/deleteProject.js
StefanBabukov: Yeah I thought of that, but in the future we will have many experiments associated to a project right? So in this case we might need to use this backend_status_deleted in other places (when you delete a particullar experiment, not the entire project)

src/components/data-management/AnalysisModal.jsx
StefanBabukov: these are changes introduced by Marcell from this PR -https://github.com/biomage-ltd/ui/pull/499

src/components/ContentWrapper.jsx
StefanBabukov: I removed those checks because its much faster and i think better to only call the api when needed. If the checks are done in the api it is delayed a little bit so thats why the user is in data-processing first and then gem2s is triggered again, what do you think?

src/components/ContentWrapper.jsx
StefanBabukov: these are changes introduced by Marcell from this PR -https://github.com/biomage-ltd/ui/pull/499

src/redux/reducers/backendStatus/index.js
StefanBabukov: these are changes introduced by Marcell from this PR -https://github.com/biomage-ltd/ui/pull/499 i believe its for when the SSR happens to update the status

src/redux/actions/backendStatus/loadBackendStatus.js
StefanBabukov: these are changes introduced by Marcell from this PR -https://github.com/biomage-ltd/ui/pull/499

src/redux/actions/projects/deleteProject.js
StefanBabukov: What do you think?

src/utils/pipelineStatusValues.js
StefanBabukov: Maybe i can remove the checks and pass the status to the api? So the checks are done there?

src/redux/actions/backendStatus/loadBackendStatus.js
ivababukova: still, do figure out what this is for and let us know

src/components/ContentWrapper.jsx
ivababukova: still, do work out what it is for and why it is added. Maybe Marcell mistakenly removed it? We cannot be merging in things in the code that we don't understand, because we risk breaking something.

src/components/data-management/AnalysisModal.jsx
ivababukova: again, do figure that out

ivababukova (changes requested):
Code coverage is falling here. We need to add unit tests for what is changed, but not covered

StefanBabukov:
I'll add them once im settled with the final changes

src/components/ContentWrapper.jsx
cosa65: @kafkasl what do you think about this? I am not sure whether the check for whether gem2s should be rerun should be done in the api or the ui

ivababukova:
Closing this PR, as these changes are getting implemented as part of Agi's ticket

PR #513, opened 2021-09-20
[BIOMAGE-1429] - Remove unnecessary marker heatmap params

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1429

https://ui-agi-remove-params.scp-staging.biomage.net

UI - https://github.com/biomage-ltd/ui/pull/513
API - https://github.com/biomage-ltd/api/pull/234
Worker - https://github.com/biomage-ltd/worker/pull/201

- Remove unnecessary parameters

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional

### DIFF
--- a/src/redux/actions/genes/loadMarkerGenes.js
+++ b/src/redux/actions/genes/loadMarkerGenes.js
@@ -6,26 +6,16 @@ import { fetchWork } from '../../../utils/work/fetchWork';
 import getTimeoutForWorkerTask from '../../../utils/getTimeoutForWorkerTask';
 
 const loadMarkerGenes = (
-  experimentId, resolution, plotUuid, numGenes = 5, selectedCellSets = false,
+  experimentId, resolution, plotUuid, numGenes = 5, selectedCellSets,
 ) => async (dispatch, getState) => {
   // Disabled linter because we are using == to check for both null and undefined values
   // eslint-disable-next-line eqeqeq
   if (experimentId == null || resolution == null) throw new Error('Null or undefined parameter/s for loadMarkerGenes');
 
-  const { experimentSettings } = getState();
-  const { processing } = experimentSettings;
-  const { hierarchy } = getState().cellSets || [];
-  const clusters = hierarchy?.filter((node) => node.key === selectedCellSets)[0]?.children;
-  const { method } = processing.configureEmbedding.clusteringSettings;
   const body = {
     name: 'MarkerHeatmap',
     nGenes: numGenes,
-    type: method,
-    config: {
-      resolution,
-    },
-    typeOfSets: selectedCellSets || 'louvain',
-    clusters,
+    cellSetKey: selectedCellSets || 'louvain',
   };
 
   dispatch({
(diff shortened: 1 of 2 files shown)


src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -6,26 +6,16 @@ import { fetchWork } from '../../../utils/work/fetchWork';
 import getTimeoutForWorkerTask from '../../../utils/getTimeoutForWorkerTask';
 
 const loadMarkerGenes = (
-  experimentId, resolution, plotUuid, numGenes = 5, selectedCellSets = false,
+  experimentId, resolution, plotUuid, numGenes = 5, selectedCellSets,
```
ivababukova: why do you still keep this `selectedCellSets` parameter?

src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -6,26 +6,16 @@ import { fetchWork } from '../../../utils/work/fetchWork';
...
   const body = {
     name: 'MarkerHeatmap',
     nGenes: numGenes,
-    type: method,
-    config: {
-      resolution,
-    },
-    typeOfSets: selectedCellSets || 'louvain',
-    clusters,
+    cellSetKey: selectedCellSets || 'louvain',
```
ivababukova: why do we need to pass `cellSetKey` at all? What does the worker use this for?

aerlaut: The worker use this to access the `cellSet` data inside the `cellSet` object which is keyed by the method name (e.g. `louvain`).

src/redux/actions/genes/loadMarkerGenes.js
aerlaut: This is used by the worker to select the `cellSet` that will be used to generate the marker heatmap.

ivababukova: Yes, makes sense. I think it will be clearer if you rename it and all its references to something like `selectedClusteringType` and also add `'louvain'` as its default value at the function declaration.

src/redux/actions/genes/loadMarkerGenes.js
ivababukova: Makes sense. Btw, looking at the current state of the code, I think that this line would break if you pass nothing as selectedCellSets (since you are removing the default = false value). But this will be fixed with my suggestion in the comment above.

aerlaut: I moved the default value initialization in commit 44ad796 and add a test for it,

src/redux/actions/genes/loadMarkerGenes.js
aerlaut: It's the `cellSet` that's selected, so we can use `scratchpad` as well, not only limited to the clustering method. I think this name is already suitable

ivababukova: Oh, I see

ivababukova: let's keep it like this then

ivababukova (approved):

PR #517, opened 2021-09-22
reflect changes to gene umi data and fit type

alexvpickering:
# Background
#### Link to issue 

https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1338

https://ui-martinfosco-ui517-api250-p.scp-staging.biomage.net/

https://github.com/biomage-ltd/api/pull/250
https://github.com/biomage-ltd/pipeline/pull/164
https://github.com/biomage-ltd/iac/pull/197

- splits up data into `pointsData` and `linesData`. This is mainly so that we can plot non-jittery upper and lower thresholds (I also tried all vega smoothings and downsampling in the UI without avail). This approach has a bonus in that we send quite a bit less data back to the UI.
- `migrateFeaturesVsUMIsPlotData` was implemented to provide backwards compatibility with `plotData` generated from previous runs of the pipeline. It results in linear upper and lower thresholds as before.

# Changes
### Code changes

# Definition of DONE

### Testing
- [ ] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional
image


ivababukova:
@alexvpickering I don't understand what is the problem that this PR is trying to solve and how it is solving it -- can you explain (for example what this `migrateFeaturesVsUMIsPlotData` function is for)? Can you go over this PR and your changes with @marcellp on Monday during your pair-up session too?

alexvpickering:
> @alexvpickering I don't understand what is the problem that this PR is trying to solve and how it is solving it -- can you explain (for example what this `migrateFeaturesVsUMIsPlotData` function is for)? Can you go over this PR and your changes with @marcellp on Monday during your pair-up session too?

This PR address the issue identified in [this ticket](https://biomage.atlassian.net/secure/RapidBoard.jspa?rapidView=1&projectKey=BIOMAGE&modal=detail&selectedIssue=BIOMAGE-1338). The cells with high gene and molecule counts (top right) should not be filtered but are:

image

In order to address this, the fit is changed from a linear model to a splines model.

The `migrateFeaturesVsUMIsPlotData` provides backwards compatibilty to data generated with previous runs of the QC pipeline. The previous data looked like this:

```
plotData = [ {log_molecules: 5, log_genes: 5, upper_threshold: 5.5, lower_threshold: 4.5 }, ... ]
```
The upper and lower thresholds were plotted by grabbing the first and last item of `plotData` (note that most of plotData was not used)

The new `plotData` that I am proposing looks like this:

```
plotData = {
 pointsData: [{log_molecules: 5, log_genes: 5 }, ...],
 linesData: [{upper_threshold: 5.5, lower_threshold: 4.5, log_molecules: 5}, ...]
}
```
Note: `plotData.linesData.length = 10` and is evenly spaced along the plotted x-axis so that a smooth curve can be plotted.

`migrateFeaturesVsUMIsPlotData` converts old plotData to the new specification. I originally tried to keep the same specification but it caused issues where there were too many points along the upper and lower threshold lines and it resulted in jaggy dashed lines (ugly). A potential solution to this could have been to sample along the interval of the plotted x-axis and grab 10 points. This would have been quite a complex sampling and did not have the advantage of the proposed solution: it halves the size of the `plotData` object.

Marcell was not available for the pair programming this AM.

src/utils/plotSpecs/generateFeaturesVsUMIsScatterplot.js
```diff
@@ -1,9 +1,9 @@
 import { stdev } from '../mathFormulas';
 
-const generateSpec = (config, plotData) => {
-  const sd = stdev(plotData.map((p) => p.log_genes));
-  const lowerCutoff = Math.min(...plotData.map((p) => p.lower_cutoff)) - sd;
-  const upperCutoff = Math.max(...plotData.map((p) => p.upper_cutoff)) + sd;
+const generateSpec = (config, { pointsData, linesData }) => {
+  const sd = stdev(pointsData.map((p) => p.log_genes));
```
aerlaut, line 4: This form is easier to read : const { pointsData, linesData } = plotData

alexvpickering:
```suggestion
const generateSpec = (config, plotData) => {
  const { pointsData, linesData } = plotData;
  const sd = stdev(pointsData.map((p) => p.log_genes));
```

src/utils/plotSpecs/generateFeaturesVsUMIsScatterplot.js
```diff
@@ -103,27 +107,25 @@ const generateSpec = (config, plotData) => {
         },
       },
       {
-        type: 'rule',
+        type: 'line',
```
alexvpickering, line 112: switch order of upper and lower cutoff

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -3,23 +3,27 @@ import PropTypes from 'prop-types';
...
 import generateSpec from '../../utils/plotSpecs/generateFeaturesVsUMIsScatterplot';
 
 const FeaturesVsUMIsScatterplot = (props) => {
   const {
     config, plotData, actions,
   } = props;
 
+  // can remove if we formally migrate
+  const newPlotData = migrateFeaturesVsUMIsPlotData(plotData);
+
```
aerlaut:
```suggestion
const newPlotData = Array.isArray(plotData) ? migrateFeaturesVsUMIsPlotData(plotData) : plotData
```

aerlaut: I think it looks cleaner, then you don't have to check inside the function

src/components/plots/helpers/migrateFeaturesVsUMIsPlotData.jsx
```diff
@@ -0,0 +1,27 @@
+const migrateFeaturesVsUMIsPlotData = (plotData) => {
```
aerlaut: Maybe change to something more descriptive? something along `transformOldDataToDataNewData`

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -3,23 +3,27 @@ import PropTypes from 'prop-types';
...
   const [plotSpec, setPlotSpec] = useState(config);
 
   useEffect(() => {
-    if (config && plotData?.length) {
-      setPlotSpec(generateSpec(config, plotData));
+    if (config && newPlotData?.pointsData?.length) {
```
aerlaut, line 20: `dataPoints` instead of `pointsData`?

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -3,23 +3,29 @@ import PropTypes from 'prop-types';
...
 import EmptyPlot from './helpers/EmptyPlot';
+import transformOldFeaturesVsUMIsPlotData from './helpers/transformOldFeaturesVsUMIsPlotData';
 import generateSpec from '../../utils/plotSpecs/generateFeaturesVsUMIsScatterplot';
 
 const FeaturesVsUMIsScatterplot = (props) => {
   const {
     config, plotData, actions,
   } = props;
 
+  // can remove if we formally migrate
```
aerlaut: Can you correct this comment to be more contextual? I'm afraid it'll confuse others. Maybe you can add something like "We can remove this if we migrate old plotdata for this plot to the new schema"

ivababukova:
Thanks for the explanation Alex, now I understand why you have this "migrateFeaturesVsUMIsPlotData" function -- to maintain backwards compatibility because of previous QC runs. It will be best to have everyone using the same schema. Do you have a plan or an idea when and how we can move everyone to be using the new schema you are proposing?

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -3,23 +3,29 @@ import PropTypes from 'prop-types';
...
 
   useEffect(() => {
-    if (config && plotData?.length) {
-      setPlotSpec(generateSpec(config, plotData));
+    if (config && newPlotData?.pointsData?.length) {
+      setPlotSpec(generateSpec(config, newPlotData));
     }
   }, [config, plotData]);
 
   const render = () => {
```
ivababukova: instead of having:

```
return (
    <>
      { render()}
    </>
  );
```
and that render function, why don't you directly return whatever you need like that:

```
  if (!newPlotData?.pointsData?.length) {
    return (
      <EmptyPlot mini={config.miniPlot} />
    );
  }

  return (
    <center data-testid='vega-container'>
      <Vega spec={plotSpec} renderer='canvas' actions={actions} />
    </center>
  );
```
Seems much easier to read to me.

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -3,23 +3,29 @@ import PropTypes from 'prop-types';
...
 
 const FeaturesVsUMIsScatterplot = (props) => {
   const {
     config, plotData, actions,
   } = props;
 
+  // can remove if we formally migrate
+  const newPlotData = (!Array.isArray(plotData) || !plotData.length)
+    ? plotData
+    : transformOldFeaturesVsUMIsPlotData(plotData);
```
ivababukova: This is very difficult to read and understand. Can you rewrite these 3 lines to be easier to read? I suggest you break them down into multiple statements.

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -41,7 +47,10 @@ const FeaturesVsUMIsScatterplot = (props) => {
 
 FeaturesVsUMIsScatterplot.propTypes = {
   config: PropTypes.object.isRequired,
-  plotData: PropTypes.array,
+  plotData: PropTypes.oneOfType([
```
ivababukova, line 44: Can you draw a `FeaturesVsUMIsScatterplot` without having `plotData`? Because looking at the usage and implementation of `FeaturesVsUMIsScatterplot`, it looks to me like this `plotData` property should be required and the code will break if it was `null`, which is the default value to it in `FeaturesVsUMIsScatterplot .defaultProps`.

src/components/plots/helpers/transformOldFeaturesVsUMIsPlotData.js
```diff
@@ -0,0 +1,29 @@
...
+    (p) => ({ log_molecules: p.log_molecules, log_genes: p.log_genes }),
+  );
+
+  // mock up linesData by grabbing first and last points
+  const logMolecules = plotData.map((p) => p.log_molecules);
+
+  const lastIndex = plotData.length - 1;
+  const linesData = [
+    {
+      lower_cutoff: plotData[0].lower_cutoff,
```
ivababukova, line 15: what is `plotData[0]`?

ivababukova: Why `plotData[0]`?

ivababukova: nevermind, I saw that it was in that vega schema before.

src/utils/plotSpecs/generateFeaturesVsUMIsScatterplot.js
```diff
@@ -132,14 +136,14 @@ const generateSpec = (config, plotData) => {
...
-    fontSize: config.title.fontSize,
-  },
+    {
+      text: config.title.text,
+      color: config.colour.masterColour,
+      anchor: config.title.anchor,
+      font: config.fontStyle.font,
+      dx: config.title.dx,
+      fontSize: config.title.fontSize,
+    },
```
ivababukova, line 146: you sure it has to be nested 1 more level?

ivababukova:
What dataset do I have to use to test this? I tried with Vicky's datasets and the plot looks unchanged

alexvpickering:
> Thanks for the explanation Alex, now I understand why you have this "migrateFeaturesVsUMIsPlotData" function -- to maintain backwards compatibility because of previous QC runs. It will be best to have everyone using the same schema. Do you have a plan or an idea when and how we can move everyone to be using the new schema you are proposing?

src/components/plots/FeaturesVsUMIsScatterplot.jsx
alexvpickering: I tried to make this more clear but let me know if it you had something else in mind

src/components/plots/FeaturesVsUMIsScatterplot.jsx
alexvpickering: done

src/components/plots/FeaturesVsUMIsScatterplot.jsx
alexvpickering: Looks better to me too - great suggestion!

src/utils/plotSpecs/generateFeaturesVsUMIsScatterplot.js
alexvpickering: Not sure. This is automated formatting (opened ui with `code ui.code-workspace`). If I try to tab these lines back by one they re-indent on save. Maybe my vscode is missing something though? Let me know if yours produces a different indentation.

src/components/plots/FeaturesVsUMIsScatterplot.jsx
alexvpickering: I just checked, it doesn't break but rather just doesn't get rendered. That being said, I think it's just lucky that it doesn't break. I considered added a test for `null` plotData but I think it makes more sense to just update the `defaultProps.plotData = []` which we are already testing for. Let me know if you'd prefer an alternative

alexvpickering:
> What dataset do I have to use to test this? I tried with Vicky's datasets and the plot looks unchanged

I would check:
-  any dataset that has already been run through the QC pipeline and check that it still shows a linear fit (aka the transform has worked for old plotData)
- any new dataset. The issue is particularly apparent for any of David Waxman's samples but the current RAM allocation for staging is insufficient. Let me know if you want to look at his though and I can either show you locally or temporarily bump the RAM in the pipeline PR.


I've changed the PR slightly because of an issue that @ogibson-biomage noticed: the splines fit is sometimes worse for datasets that do not show the saturation. For example, the covid19 MISC samples filter less of the RBCs. As a result, we are proposing that linear is the default fit and spline is an option that can be selected as needed.

src/components/plots/FeaturesVsUMIsScatterplot.jsx
ivababukova: That sounds like a good solution

src/utils/plotSpecs/generateFeaturesVsUMIsScatterplot.js
ivababukova: hmm, mine shows the same as yours. Maybe it is ok then

src/components/plots/FeaturesVsUMIsScatterplot.jsx
ivababukova: it looks better :)

ivababukova (approved):

ivababukova:
Codewise it looks good, happy to merge if you and Oliver think the plot looks well with different datasets in staging

alexvpickering:
> Thanks for the explanation Alex, now I understand why you have this "migrateFeaturesVsUMIsPlotData" function -- to maintain backwards compatibility because of previous QC runs. It will be best to have everyone using the same schema. Do you have a plan or an idea when and how we can move everyone to be using the new schema you are proposing?

I discussed this with @aerlaut and he felt the current solution (format old plotData so that it looks new) was the most reasonable. We could have alternatively migrated old data but that was suggested to be too painful (would require hitting dynamodb just to get the S3 file names for the genes vs umi plot data objects). At some point in the future we could remove the transformation code if we decide that sufficient time has passed.

alexvpickering:
> Codewise it looks good, happy to merge if you and Oliver think the plot looks well with different datasets in staging

Great - we will test it out and merge when ready. One question I have is that I updated the default `regressionType` from `gam` to `linear` (it's not a gam fit). It seemed to "just work" - I was expecting the API to break but perhaps the API doesn't check the config specification anymore?

ivababukova:
> regressionType

The API should check this and here is the schema for it: https://github.com/biomage-ltd/api/blob/master/src/specs/models/processing-config-bodies/ProcessingConfigNumGenesVsNumUMIs.v1.yaml Maybe it doesn't fail because the schema doesn't break if we add a property that is not defined there? I would double check on this, because yesterday we caught a schema that was actually never working.

PR #524, opened 2021-10-01
[HACKATHON] init vitessce heatmap

alexvpickering:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1544

https://ui-martinfosco-ui524.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -281,25 +285,32 @@ const Embedding = (props) => {
...
-            updateCellsHover={updateCellsHover}
-            updateViewInfo={updateCellCoordinates}
-            clearPleaseWait={clearPleaseWait}
+            setViewState={({ zoom, target }) => {
+              setCellRadius(cellRadiusFromZoom(zoom));
+
+              setView({ zoom, target });
+            }}
+            getExpressionValue={() => { }}
+            getCellIsSelected={() => { }}
```
cosa65, line 311: These are props that the scatterplot crashes if it doesn't receive them but are not really essential, so we had to fill them in with empty functions

src/components/data-exploration/heatmap/HeatmapCellInfo.jsx
```diff
@@ -0,0 +1,59 @@
...
+      size='small'
+      style={{
+        zIndex: 6,
+        border: 0,
+        position: 'absolute',
+        left: `${coordinates.x + 20}px`,
+        top: `${coordinates.y + 20}px`,
+        pointerEvents: 'none',
+      }}
+    >
```
cosa65, line 23: Perhaps abstract this styling to a common component between `HeatmapTracksCellInfo`, `CellInfo` (rename to `EmbeddingCellInfo`), `HeatmapCellInfo`

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -0,0 +1,27 @@
...
+ * Returns an array of properties of all the cellSets
+ * within the cellSetClass that cellId belongs to.
+ * @param {string} cellId The id of the cell we are searching for.
+ * @param {string} cellSetClassKey The cellSetClass we are searcing inside of
+ *  (e.g.: sample, louvain, scratchpad, etc..).
+ * @param {string} hierarchy CellSets hierarchy.
+ * @param {string} properties CellSets properties.
+ */
+
+const getCellClassProperties = (cellId, cellSetClassKey, hierarchy, properties) => {
```
cosa65, line 13: Return only the first matching cellSet properties since we never use the other ones

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -33,22 +48,31 @@ const HeatmapPlot = (props) => {
...
+  const [heatmapData, setHeatmapData] = useState(null);
+  const [highlightedTrackData, setHighlightedTrackData] = useState(null);
+
   const [isHeatmapGenesLoading, setIsHeatmapGenesLoading] = useState(false);
-  const currentHeatmapSettings = useRef();
 
+  const [geneHighlight, setGeneHighlight] = useState(null);
+  const [cellHighlight, setCellHighlight] = useState(null);
+
+  const [maxCells, setMaxCells] = useState(1000);
```
ivababukova: what do we use this maxCells for?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -33,22 +48,31 @@ const HeatmapPlot = (props) => {
...
+  const [vitessceData, setVitessceData] = useState(null);
+
+  const cellCoordinatesRef = useRef({ x: 200, y: 300 });
+  const currentHeatmapSettingsRef = useRef(null);
```
ivababukova: shouldn't the default be equal to some default settings, not `null`? Do these settings refer to the settings that the user selects on the top right corner of the marker heatmap window?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: At the time it was probably related to downsampling I woudl assume, but now it doesn't seem to be doing anything

cosa65: At the time it was probably related to downsampling I woudl assume, but now it doesn't seem to be doing anything

cosa65: It was already not doing anything from before vitessce)

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: Yes, they are those settings.
Usually we store the default settings in redux, so from the first `useSelector` this is filled in with the actual default settings

cosa65: from what I can see, this `useRef` is not used to have those settings accessed and used, it is just to compare with previous settings and know whether to repopulate the heatmap data or not

cosa65: I think this useRef could be removed now that I look at it, it is used to see if this changed on a `useSelector` trigger:
`const heatmapSettings = useSelector((state) => state.componentConfig[COMPONENT_TYPE]?.config)`

but actually this should be enough to replace the `useRef` functionality:
`const heatmapSettings = useSelector((state) => state.componentConfig[COMPONENT_TYPE]?.config, _.isEqual)`

src/components/data-exploration/heatmap/HeatmapPlot.jsx
ivababukova: how about we remove it then?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
ivababukova: sounds good, let's remove it if we don't need it then!

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -2,28 +2,43 @@
...
-import PlatformError from '../../PlatformError';
-import { updateCellInfo } from '../../../redux/actions/cellInfo';
-import { loadGeneExpression, loadMarkerGenes } from '../../../redux/actions/genes';
-import { loadCellSets } from '../../../redux/actions/cellSets';
 
-import { loadComponentConfig } from '../../../redux/actions/componentConfig';
-import populateHeatmapData from '../../plots/helpers/populateHeatmapData';
-import Loader from '../../Loader';
+import { loadCellSets } from 'redux/actions/cellSets';
+import { loadGeneExpression, loadMarkerGenes } from 'redux/actions/genes';
```
ivababukova, line 16: Can you check if these imports are correct? I am getting a warning when I open them on my vs code:
Screenshot 2021-12-02 at 15 41 25

src/components/data-exploration/heatmap/HeatmapPlot.jsx
ivababukova: If you remove it, you will also need to remove this useEffect:

```
  useEffect(() => {
    setMaxCells(Math.floor(width * 0.8));
  }, [width]);
```

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -231,29 +282,79 @@ const HeatmapPlot = (props) => {
...
+            <HeatmapCellInfo
+              cellId={cellHighlight}
+              geneName={geneHighlight}
+              geneExpression={focusedExpression?.rawExpression.expression[cellHighlight]}
+              coordinates={cellCoordinatesRef.current}
+            />
+          )
+        }
+      </div>
     </div>
```
ivababukova: isn't there a situation when you don't want to display any of the two cell info components, but just nothing?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -106,29 +139,21 @@ const HeatmapPlot = (props) => {
...
-  useEffect(() => {
-    if (cellSetsLoading || cellSetsHierarchy.length === 0) {
-      return;
-    }
-
-    const legends = legendIsVisible ? spec.legends : [];
-    setVegaSpec({ ...spec, legends });
-  }, [legendIsVisible]);
-
   useEffect(() => {
```
ivababukova: somewhere here you have a useEffect that gets triggered if either of the following changes:

`[selectedGenes, loadingGenes, markerGenesLoading]`

What is the technical difference between `loadingGenes` and `markerGenesLoading`? Aren't `markerGenesLoading` just a type of `loadingGenes`?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -106,29 +139,21 @@ const HeatmapPlot = (props) => {
...
-    currentHeatmapSettings.current = heatmapSettings;
+    currentHeatmapSettingsRef.current = heatmapSettings;
 
     const data = populateHeatmapData(
-      cellSets, heatmapSettings, expressionData, selectedGenes, true,
+      cellSets, heatmapSettings, expressionData, selectedGenes, true, true,
     );
-    setVegaDataWithDebounce(data);
+
+    setHeatmapDataWithDebounce(data);
```
ivababukova: somewhere here you have a useEffect that gets triggered if either of the following changes:

`  }, [
    selectedGenes,
    heatmapSettings,
    maxCells,
    markerGenesLoading,
    cellSetsLoading,
    cellSetsHidden,
  ]);`

why only depending on `markerGenesLoading`, but not `loadingGenes`?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -152,26 +177,52 @@ const HeatmapPlot = (props) => {
...
+    // scale so that each gene has minimum 0 max 255
+    const scaledCellByGeneMatrix = cellByGeneMatrix.map((row) => {
+      const geneMin = Math.min(...row);
+      const geneMax = Math.max(...row);
+
+      return row.map((x) => convertRange(x, [geneMin, geneMax], [0, 255]));
+    });
+
+    // vitesse Heatmap uses:
+    // array with shape [cell_1 gene_1, ..., cell_1 gene_n, cell_2 gene_1, ... ]
```
ivababukova: gene_x numbers are messed up in this comment

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -152,26 +177,52 @@ const HeatmapPlot = (props) => {
...
+    // accomplish with transpose and flatten
+    const cellOrderedExpression = _.unzip(scaledCellByGeneMatrix).flat();
+
```
ivababukova: this function is difficult to read and follow, let's look into how to simplify it!

src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js
```diff
@@ -0,0 +1,27 @@
+import getCellClassProperties from 'utils/cellSets/getCellClassProperties';
+import { hexToRgb } from 'components/plots/helpers/heatmap/utils';
+
+const generateVitessceHeatmapTracksData = (trackOrder, hierarchy, properties, cells) => {
```
ivababukova: this function is only used in `src/components/plots/helpers/heatmap/populateHeatmapData.js`, so why don't you move it to be there directly?

src/components/plots/helpers/heatmap/populateHeatmapData.js
```diff
@@ -1,10 +1,12 @@
 import _ from 'lodash';
-import SetOperations from '../../../utils/setOperations';
-import { union } from '../../../utils/cellSetOperations';
+import generateVitessceHeatmapTracksData from 'components/plots/helpers/heatmap/generateVitessceHeatmapTracksData';
+
+import SetOperations from 'utils/setOperations';
+import { union } from 'utils/cellSetOperations';
 
 const populateHeatmapData = (
```
ivababukova, line 9: have you thought about refactoring this function, splitting it into 2: one for generating vitessce heatmap data and one for non-vitessce heatmap data?

ivababukova: It is a 270 lines long function that includes nested functions, it will benefit a lot from some better re-organisation. Right now it is really hard to follow

src/components/plots/helpers/heatmap/utils.js
```diff
@@ -0,0 +1,38 @@
+const hexToRgb = (hex) => {
```
ivababukova: there is already a duplicate `hexToRgb` helper function in `src/utils/embeddingPlotHelperFunctions`. Since now you will be using this function in the heatmap too, better put it in a common place where both the embedding and the heatmap can access it

src/components/plots/helpers/heatmap/utils.js
```diff
@@ -0,0 +1,38 @@
...
+      // eslint-disable-next-line no-plusplus
+      k++;
+      matrix[k] = [];
+    }
+
+    matrix[k].push(list[i]);
+  }
+
+  return matrix;
+};
```
ivababukova: Why writing this function in this way? I know it works, but it is different style from the rest of the code in the ui.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: image
I'm not getting that warning, perhaps an `npm ci` fixes your setup?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: This case is handled inside `HeatmapCellInfo`, but you're right in that that case's handling it should be more explicit

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: They are 2 different worker requests: `markerGenesLoading` is for loading the top N marker genes and `loadingGenes` for loading genes in particular

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: I'll try removing `markerGenesLoading`, because this shoudl change when selected genes data changes (so on `selectedGenes` changing)

src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js
cosa65: I feel like `populateHeatmapData` already is way too large, ideally I would even think about moving those away into utils or separate files to make it easier to follow

src/components/plots/helpers/heatmap/populateHeatmapData.js
cosa65: it is kind of already split up, the only part that handles vitessce stuff is right at the end, and it's only this:

  ```
// Directly generate track data.
  if (!vitessce) {
    const trackData = trackOrder.map((rootNode) => generateTrackData(
      cells,
      rootNode,
    ));

    data.trackColorData = trackData.map((datum) => datum.trackColorData).flat();
    data.trackGroupData = trackData.map((datum) => datum.groupData).flat();
    data.clusterSeparationLines = trackData.length > 0 ? trackData[0].clusterSeparationLines : [];
  } else {
    data.trackColorData = generateVitessceHeatmapTracksData(
      trackOrder, cellSets.hierarchy, cellSets.properties, cells,
    );
  }
```

Comment (cosa65) [reply src/components/plots/helpers/heatmap/populateHeatmapData.js:9]: refactoring is another matter, but I don't know if we should do everything in this PR, would refactoring the non-vitessce specific parts of `populateHeatmapData` be part this ticket?

File: src/components/plots/helpers/heatmap/utils.js (hunk 14 above)
Comment (cosa65) [reply src/components/plots/helpers/heatmap/utils.js]: I agree, good catch

File: src/components/plots/helpers/heatmap/utils.js (hunk 15 above)
Comment (cosa65) [reply src/components/plots/helpers/heatmap/utils.js]: yep, I'll review/rewrite it

File: src/components/data-exploration/heatmap/HeatmapPlot.jsx (hunk 10 above)
Comment (cosa65) [reply src/components/data-exploration/heatmap/HeatmapPlot.jsx]: why? it is all genes for first cell, then all genes for second cell, etc.

Comment (ivababukova) [reply src/components/data-exploration/heatmap/HeatmapPlot.jsx]: no, the comment is right, I got confused

File: src/components/data-exploration/heatmap/HeatmapPlot.jsx (hunk 7 above)
Comment (ivababukova) [reply src/components/data-exploration/heatmap/HeatmapPlot.jsx]: it will make more sense if it is handled from here

File: src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js (hunk 12 above)
Comment (ivababukova) [reply src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js]: it definitely is too large, I did also add a comment about that, but it doesn't seem like that moving this function away from the file will make it easier to read and understand. It seems to me that that file needs other type of refactoring.

File: src/components/plots/helpers/heatmap/populateHeatmapData.js (hunk 13 above)
Comment (ivababukova) [reply src/components/plots/helpers/heatmap/populateHeatmapData.js:9]: could be a good opportunity to do it, but I understand that this PR has been on for a long while and we need to finish it off.

Comment (ivababukova) [reply src/components/plots/helpers/heatmap/populateHeatmapData.js:9]: Leaving it up to you, if you think it is going to take more effort, then leave it for other time

Comment (cosa65) [reply src/components/plots/helpers/heatmap/populateHeatmapData.js:9]: Actually I think `buildExpressionMatrix` is just transforming what we get from `populateHeatmapData` to another format, so you're right, the `buildExpressionMatrix` part is also part of populateHeatmapData for vitessce.

I will split it up into two functions when it is necessary.

File: src/components/data-exploration/heatmap/HeatmapPlot.jsx (hunk 6 above)
Comment (ivababukova) [reply src/components/data-exploration/heatmap/HeatmapPlot.jsx:16]: hmm, it still didn't ...

Comment (cosa65) [reply src/components/data-exploration/heatmap/HeatmapPlot.jsx:16]: Can I resolve this? I don't know what is the problem there, but it seems like a specific issue, not of this PR

File: src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js (hunk 12 above)
Comment (cosa65) [reply src/components/plots/helpers/heatmap/generateVitessceHeatmapTracksData.js]: I did some ordering and separation into files/structure, we can go over it and you give me your opinion

File: src/components/plots/helpers/heatmap/utils.js (hunk 15 above)
Comment (cosa65) [reply src/components/plots/helpers/heatmap/utils.js]: After some refactoring `listToMatrix` wasn't necessary anymore, so I removed the function

File: src/components/data-exploration/heatmap/HeatmapPlot.jsx [hunk 16]
```diff
@@ -231,29 +230,79 @@ const HeatmapPlot = (props) => {
...
+        width={width - heatmapRightMargin}
         height={height}
+        colormap='plasma'
+        colormapRange={[0.0, 1.0]}
+        expressionMatrix={vitessceData?.expressionMatrix}
+        cellColors={vitessceData?.metadataTracks.dataPoints}
+        cellColorLabels={vitessceData?.metadataTracks.labels}
+        transpose
+        viewState={viewState}
+        setViewState={({ zoom, target }) => { setViewState({ zoom, target }); }}
```
cosa65: try `setViewState={setViewState}`

src/components/plots/helpers/heatmap/vitessce/utils/generateVitessceHeatmapTracksData.js
```diff
@@ -0,0 +1,27 @@
+import getCellClassProperties from 'utils/cellSets/getCellClassProperties';
```
cosa65, line 1: Add comments explaining what is going on here and in other `populteHeatmap` related files

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
-          {
-            key: 'louvain-6',
-          },
-          {
-            key: 'louvain-7',
-          },
-        ],
-      }]));
-  });
+// import { Heatmap } from 'vitessce/dist/umd/production/heatmap.min';
```
ivababukova: remove this

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
+jest.mock('utils/work/seekWorkResponse', () => ({
+  __esModule: true,
+  seekFromAPI: jest.fn(),
+  seekFromS3: () => Promise.resolve(null),
+}));
 
-    expect(component.find('HeatmapPlot').length).toEqual(1);
-    expect(component.find(Empty).length).toEqual(1);
-  });
+jest.mock('next/dynamic', () => () => () => 'Sup Im a heatmap');
```
ivababukova: :)

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
 
-    component = mount(
-      <Provider store={store}>
-        <HeatmapPlot experimentId={experimentId} width={200} height={200} />
+let storeState = null;
+const loadAndRenderDefaultHeatmap = async () => {
+  storeState = makeStore();
+
+  storeState.dispatch(loadProcessingSettings(experimentId));
+  storeState.dispatch(loadBackendStatus(experimentId));
```
ivababukova, line 77: why do you need these for the heatmap to work?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
-    });
+  it('Renders the heatmap component by default if everything loads', async () => {
+    fetchMock.mockIf(/.*/, mockAPI(generateDefaultMockAPIResponses(experimentId, fake.PROJECT_ID)));
 
-    component = mount(
-      <Provider store={store}>
-        <HeatmapPlot experimentId={experimentId} width={200} height={200} />
-      </Provider>,
-    );
+    seekFromAPI.mockClear();
```
ivababukova: you can move this and the other `seekFromAPI.mockClear();` to the `beforeEach`

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
 
-    expect(component.find('HeatmapPlot').length).toEqual(1);
+    seekFromAPI.mockClear();
+    seekFromAPI.mockImplementation((a, b, c, requested) => customWorkerResponses[requested]());
 
-    const expectedMarkerGenes = 3;
+    await loadAndRenderDefaultHeatmap();
 
-    expect(loadMarkerGenes).toHaveBeenCalledWith('123', 10, 'interactiveHeatmap', expectedMarkerGenes);
+    expect(screen.getByText(/We're getting your data.../i)).toBeInTheDocument();
```
ivababukova: I am very pedantic here, but can you add space between `data` and `...` in here and in the actual code implementation text?

ivababukova: there should be spaces between these

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
...
-    await act(async () => {
-      component.find('PlatformError').props().onClick();
-    });
+  // Add test for markerGenesLoadingError
 
-    // on clicking platform error button it reloads
-    const expectedMarkerGenes = 3;
-    expect(loadMarkerGenes).toHaveBeenCalledWith('123', 10, 'interactiveHeatmap', expectedMarkerGenes);
-  });
+  // Think more tests
```
ivababukova: Can you test for:
1. If cells are hidden from embedding, they also get hidden from the heatmap. Same for unhiding.
2. If user selects to view expression of gene from the gene list, the heatmap updates correctly.
3. The heatmap cell info components show and hide when appropriate.
4. The settings on the heatmap work -- when you click and unclick stuff there the heatmap updates accordingly.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,163 @@
 import _ from 'lodash';
```
ivababukova, line 1: Nice test rewrite 😍

src/__test__/components/data-exploration/heatmap/__snapshots__/HeatmapPlot.test.jsx.snap
```diff
@@ -1,7 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
-
-exports[`HeatmapPlot dispatches loadCellSets action when no cell sets are in the store 1`] = `
-Object {
-  "type": "cellSets/loading",
-}
```
ivababukova: it doesn't seem like you are using this

src/components/data-exploration/heatmap/HeatmapPlot.jsx
ivababukova: yes, let's resolve this

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -149,29 +162,14 @@ const HeatmapPlot = (props) => {
...
-    }
+  useEffect(() => {
+    if (!heatmapData) return;
 
-    const { cellId: cellName } = args[1].datum;
-
-    dispatch(updateCellInfo({ cellName }));
-  };
+    setVitessceData(heatmapData);
+  }, [heatmapData]);
```
ivababukova: for more unit test inspirations -- code coverage on my VS code is telling me that these are not tested, so you can write unit tests for them

vickymorrison:
1. In the default view, I suggest removing "Cells" and "Genes" because they are unnecessary and take up a lot of space.
Screenshot 2021-12-10 at 11 30 28

2. From the UI, it is not clear that I can zoom on the heatmap. To be fair, this is also true for the embedding. Having thought about it more, I actually think that we should add zoom in/out controls to both the embedding and the heatmap to improve UX. This should be a separate ticket. You can ignore this comment for now.

3. I thought you said you would remove the settings->expression values menu and options because they are no longer relevant?

4. Similarly, there is no legend so the settings->legend option in the menu can also be removed.

5. I added metadata tracks and then changed their order - this broke the heatmap. I could reproduce it easily:
Screenshot 2021-12-10 at 11 35 49

6. The metadata track is called 'Custom cell sets' in the menu but 'Scratchpad' on the heatmap itself (see above screenshot).

src/components/plots/helpers/heatmap/populateHeatmapData.js
ivababukova: Much better than before! A small thing you can do to make the code more readable: move line 33 `const trackOrder = Array.from(selectedTracks).reverse();` to be just before lines 149 and 150:

```  const cellOrder = generateCellOrder(groupedTracks);
  const geneOrder = selectedGenes;
```

In that way, you have the `populateHeatmapData` function clearly separated into two sections: section with all the inner functions used within it, and section that triggers the actual executions of these functions.

cosa65:
> 5. I added metadata tracks and then changed their order - this broke the heatmap. I could reproduce it easily:
> 
> Screenshot 2021-12-10 at 11 35 49

This issue is independent from vitessce, it can reproduced in `ui-default` right now too. I want to fix it but first wanted to finish this ticket.

cosa65:
> I thought you said you would remove the settings->expression values menu and options because they are no longer relevant?

Yes, but I was waiting to begin work on this until you and alex had coordinated on whether we also wanted to remove this from plots & tables, but I can just do it here for now.

There's also a couple of questions more or less related to this I have for alex that I need to wait until monday for (like whether we should use the truncated or the raw expressions to display).

UPDATE: Removed (changes wont propagate to staging because tests are broken)

cosa65:
> 1. In the default view, I suggest removing "Cells" and "Genes" because they are unnecessary and take up a lot of space.

The spaces are unnecessary initially, but when zooming in these legends are replaced by the gene/cell names labels, in fact I believe the space allocated to the margin depends on the length of the gene's/cell's names. You can see the difference here:
image

image

They are both the same code, the "Cells" and "Genes" texts are just added but don't define the margin sizes

Apart from this, if the heatmap can be resized, is this likely to be an issue?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
cosa65: processingSettings to know the resolution of configure embedding, which is sent in the worker task.
backend status is used in the `fetchWork` call

vickymorrison:
> > 5. I added metadata tracks and then changed their order - this broke the heatmap. I could reproduce it easily:
> > 
> > Screenshot 2021-12-10 at 11 35 49
> 
> This issue is independent from vitessce, it can reproduced in `ui-default` right now too. I want to fix it but first wanted to finish this ticket.

No problem! As I said, just made a ticket for the bug and don't worry about it for now.

vickymorrison:
> > I thought you said you would remove the settings->expression values menu and options because they are no longer relevant?
> 
> Yes, but I was waiting to begin work on this until you and alex had coordinated on whether we also wanted to remove this from plots & tables, but I can just do it here for now.
> 
> There's also a couple of questions more or less related to this I have for alex that I need to wait until monday for (like whether we should use the truncated or the raw expressions to display).

The plots and tables heatmap is a separate question. Don't wait on an answer about the P&T heatmap to complete this one. You and Alex's conclusion is that this menu is not relevant here in Data Exploration so just remove it and don't make any changes in P&T for now.

vickymorrison:
> > 1. In the default view, I suggest removing "Cells" and "Genes" because they are unnecessary and take up a lot of space.
> 
> The spaces are unnecessary initially, but when zooming in these legends are replaced by the gene/cell names labels, in fact I believe the space allocated to the margin depends on the length of the gene's/cell's names. You can see the difference here: image
> 
> They are both the same code, the "Cells" and "Genes" texts are just added but don't define the margin sizes
> 
> Apart from this, if the heatmap can be resized, is this likely to be an issue?

It's clear from the onboarding callt that many users have smaller screens than my macbook. So I think we need to consider the best use of the available space as high priority, especially in the Data Exploration module. Can the Genes / Cells labels be removed in the default view and then appear when the plot resizes on zoom? As an alternative compromise, I suppose I could live with the Gene label being there on the default view, but the Cells label is completely useless and really does take up a lot of space.

src/__test__/components/data-exploration/heatmap/__snapshots__/HeatmapPlot.test.jsx.snap
cosa65: I didn't need the snapshots for this so it was deleted I guess, why?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
ivababukova: indeed, that's for the marker heatmap to work. Thanks!

src/__test__/components/data-exploration/heatmap/__snapshots__/HeatmapPlot.test.jsx.snap
ivababukova: it looked like you added it .. or I got confused

ivababukova: my point was that since this file needs to be removed, because it is not used

cosa65: oh, I think it was removed, I looked in file changes and it said file deleted

ivababukova (commented):
Code-wise, looks great so far. When you add those more tests and address Vicky's comments, let me know and I will have another look.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
cosa65: 1. Impemented.
2. Implemented.
3. Implemented.
4. We just receive the data from a `useSelector`, but I implemented a check that the heatmap responds to that data changing.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,342 @@
...
-    const store = mockStore({ ...initialState });
+    // Renders correctly
+    expect(screen.getByText(/Sup Im a heatmap/i)).toBeInTheDocument();
 
-    component = mount(
-      <Provider store={store}>
-        <HeatmapPlot experimentId={experimentId} width={200} height={200} />
-      </Provider>,
-    );
+    // If a new gene suddenly is suddenly selected
```
ivababukova: can you fix this comment?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,342 @@
...
-    expect(component.find('HeatmapPlot').length).toEqual(1);
+    seekFromAPI.mockImplementation((a, b, c, requested) => customWorkerResponses[requested]());
+
+    await loadAndRenderDefaultHeatmap();
+
+    // Renders correctly
+    expect(screen.getByText(/Sup Im a heatmap/i)).toBeInTheDocument();
 
-    const expectedMarkerGenes = 5;
+    // If a new gene suddenly is suddenly selected
```
ivababukova: can you fix this comment?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,342 @@
...
+    await act(async () => {
+      storeState.dispatch(loadGeneExpression(experimentId, [...markerGenesData5.order, 'loading_gene_id'], 'interactiveHeatmap'));
+    });
```
ivababukova, line 206: why would this result in an error?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -1,545 +1,342 @@
...
+    expect(screen.getByText(/Sup Im a heatmap/i)).toBeInTheDocument();
 
-    expect(component.find('HeatmapPlot').length).toEqual(1);
+    // If groupedTracks change
+    await act(async () => {
+      storeState.dispatch(
+        updatePlotConfig('interactiveHeatmap', {
+          groupedTracks: ['sample', 'louvain'],
+        }),
+      );
```
ivababukova, line 268: was there not an easy way to trigger change of groupedTracks via the ui rather than having to dispatch an action?

src/components/data-exploration/heatmap/HeatmapExpressionValuesSettings.jsx
```diff
@@ -1,52 +0,0 @@
-import React from 'react';
-import { useSelector, useDispatch } from 'react-redux';
-import { Radio } from 'antd';
-import PropTypes from 'prop-types';
-
-import { updatePlotConfig } from '../../../redux/actions/componentConfig';
-
-const HeatmapExpressionValuesSettings = (props) => {
```
ivababukova, line 8: why do we no longer need the ExpressionValueSettings?

src/components/data-exploration/heatmap/HeatmapLegendVisibilitySettings.jsx
```diff
@@ -1,48 +0,0 @@
-import React, { useState } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
-import { Radio } from 'antd';
-import PropTypes from 'prop-types';
-
-import { updatePlotConfig } from '../../../redux/actions/componentConfig';
-
-const HeatmapLegendVisibilitySettings = (props) => {
```
ivababukova, line 8: Why do we no longer need this file?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
cosa65: No, because we are rendering the heatmapplot only, the groupedTracks are changed in a different component

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
cosa65: Because I set it so that the api returns an error here:

```
const customWorkerResponses = _.merge(
      _.cloneDeep(mockWorkerResponses),
      { 'loading_gene_id-expression': errorResponse },
);
```

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
cosa65: What is wrong with it? The idea was to show what happens "if something happens" and then check a lot of things through expect

src/components/data-exploration/heatmap/HeatmapExpressionValuesSettings.jsx
cosa65: Because we only show one type of expression values, as agreed on with vicky and alex

ivababukova: cool, that should already help with the speed of the heatmap too

src/components/data-exploration/heatmap/HeatmapLegendVisibilitySettings.jsx
cosa65: Because the new type of expression values we show is calculated in a way that is gene-specific, so we can't show a bar saying "this intensity of color means a gene's expression in a cell is 3" universally because there is a different bar of intensity for each gene. This was discussed and agreed on with vicky and alex already.

ivababukova (approved):

vickymorrison:
Looking really good! Only 1 comment: the bottom gene in the heatmap is cut off. This happens even though my browser window is the full height of my screen. Considering that most users have smaller screens than me, it would be good to fix this.
Screenshot 2021-12-14 at 15 59 25

vickymorrison (approved):

PR #529, opened 2021-10-08
[BIOMAGE-133] - Dot plot container

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1330

https://ui-agi-ui529.scp-staging.biomage.net/

- Added dotplot

# Changes
### Code changes

# Definition of DONE

### Testing
- [X] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


kafkasl:
I've changed it into draft as it has mock data and it's not ready for review nor has enough tests.

aerlaut:
> I've changed it into draft as it has mock data and it's not ready for review nor has enough tests.

@kafkasl for this ticket / PR, we are using mock data instead of the real one. Linking the front end and the backend is done in another ticket : https://biomage.atlassian.net/browse/BIOMAGE-1451

vickymorrison (approved):

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
```diff
@@ -135,7 +135,7 @@ describe('ViolinIndex', () => {
   it('allows selection of the grouping/points, defaulting to louvain and All', async () => {
     await renderViolinIndex();
 
-    const dataSelection = rtl.screen.getByText('Select Data');
+    const dataSelection = rtl.screen.getByText('Select data');
```
kafkasl: consider using case-insensitive matching to avoid all these issues: 
https://testing-library.com/docs/queries/about#textmatch-examples

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -9,28 +9,28 @@ import _ from 'lodash';
...
-jest.mock('../../../utils/socketConnection', () => ({
+jest.mock('components/plots/Header', () => () => <div />);
+jest.mock('utils/socketConnection', () => ({
   __esModule: true,
   default: new Promise((resolve) => {
     resolve({ emit: jest.fn(), on: jest.fn(), id: '5678' });
   }),
 }));
-jest.mock('../../../utils/work/fetchWork', () => ({
+jest.mock('utils/work/fetchWork', () => ({
```
kafkasl: the actual `ListGenes` response contains more stuff like `"{"total": 16031, "rows": [{"gene_names": "S100A9", "dispersions": 23.4435, "_row": "ENSG00000163220"},`. Unless there is a specific reason not include them I would add them for the sake of completeness. At least the `total` field.

kafkasl: for the `GeneExpression`, I was not able to find the `rawExpression`, the response looks like `"{"S100A8": {"truncatedExpression": {"min": 0.0, "max": 5.7651, "expression": [0, null, null, 0, 0, 0, 0, 5.7651, null, 0, 0, null, 1.4358, 0, 0, null, 0, null, 1.2752, 5.4748, null, 0, null, null, 4.5487, null, 0, 0, null, 0, null...`. Verify whether this mock is outdated and update it if needed.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -182,14 +182,14 @@ let loadedProcessingConfigSpy;
...
-  rtl.render(
+  render(
     <Provider store={store}>
       <MarkerHeatmap
         experimentId={experimentId}
       />
     </Provider>,
   );
-  await rtl.waitFor(() => expect(loadConfigSpy).toHaveBeenCalledTimes(1));
+  await waitFor(() => expect(loadConfigSpy).toHaveBeenCalledTimes(1));
```
kafkasl: what is the purposes of using `expect` outside actual test cases? Unless there is a good reason I find it's just confusing, if possible add it inside the test where it belongs

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -182,14 +182,14 @@ let loadedProcessingConfigSpy;
...
 };
 
 describe('Marker heatmap plot', () => {
```
kafkasl, line 77: the `loads initially` just checks that the load & config have been called right? Isn't it possible to actually test that the user will be seeing a heatmp?

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
```diff
@@ -217,48 +217,48 @@ describe('Marker heatmap plot', () => {
...
+    userEvent.click(markerGenes);
+    const nGenesInput = screen.getByRole('spinbutton', { name: 'Number of genes input' });
     userEvent.type(nGenesInput, 10);
 
-    const runButton = rtl.getByText(geneSelection.parentElement, 'Run');
-    userEvent.click(runButton);
+    userEvent.click(screen.getByText('Run'));
 
-    await rtl.waitFor(() => expect(loadMarkersSpy).toHaveBeenCalled());
+    await waitFor(() => expect(loadMarkersSpy).toHaveBeenCalled());
```
kafkasl: I think this check could be expanded somehow. I can imagine very easily situations where the users selects 10 genes and the call is actually done (so the test won't fail) but the actual call or the result is wrong.

kafkasl (dismissed):

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+    ],
+  },
+];
+
+const dotPlot = (props) => {
+  const { experimentId } = props;
+
+  const dispatch = useDispatch();
+  const config = useSelector((state) => state.componentConfig[plotUuid]?.config);
+  const fetching = useSelector((state) => state.genes.properties.views[plotUuid]?.fetching);
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: you could do this instead:
`const {fetching: genesFetching, data: highestDispersionGenes} = useSelector(...`

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -0,0 +1,193 @@
...
+            onUpdate={updatePlotWithChanges}
+            cellSets={cellSets}
+          />
+        ) : <Skeleton.Input style={{ width: 200 }} active />}
+      </Panel>
+    </>
+  );
+
+  const renderPlot = () => {
+    if (!config) {
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: this will never get called, right? Since before this we are doing 
```
  if (!config) {
    return <Skeleton />;
  }
```

after cleaning this up I think it woudl be a good idea to move all returns to the bottom of the page so these kind of problems are more easily noticeable.

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
```diff
@@ -47,49 +49,30 @@ CardItem.propTypes = {
 };
 
 const PlotsTablesHome = ({ experimentId, experimentData, route }) => {
-  let lastUpdatedVolcano = useSelector(
-    (state) => state.componentConfig.volcanoPlotMain?.lastUpdated,
+  const lastUpdatedVolcano = useSelector(
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/index.jsx:52]: Nice!

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -313,6 +331,72 @@ const MarkerHeatmap = ({ experimentId }) => {
     return (<Skeleton />);
   }
 
+  const getCellOptions = (type) => {
+    const filteredOptions = hierarchy.filter((element) => (
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: This looks a lot like a selector that stefan created, is there any chance we could reuse it?

src/components/plots/styling/frequency/SelectCellSets.jsx
```diff
@@ -64,7 +64,7 @@ const SelectCellSets = (props) => {
         <Select
           aria-label='cell sets'
           value={{
-            key: firstLetterUppercase(config.proportionGrouping),
+            key: _.upperFirst(config.proportionGrouping),
```
cosa65: the key property here is the one that is used to display? or only to store in redux? in the case it goes to redux, why do we to this `upperFirst`?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: This will fail if `state.genes.properties.views[plotUuid]` evaluaes to `undefined`, as you will be asking the UI to `destructure property of undefined`. We can do this if we add a `|| {}` at the end, to which the variables will evaluate to `undefined`.

src/components/plots/styling/frequency/SelectCellSets.jsx
aerlaut: This is only used to display, not the value that goes to redux.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: This will not be called if `config` is not available (still loading). If `config` is available, this will be called.

This is placed after the `return` by purpose so that the function is not instantiated if it's not needed (if `config` does not exist). Is that the right way to think about it?

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
aerlaut: test for plot added

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: definitely!

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
aerlaut: List genes is not used, so I'm removing it.

src/utils/work/seekWorkResponse.js
```diff
@@ -95,8 +95,11 @@ const seekFromAPI = async (
       });
     });
 
-    return Promise.race([timeoutPromise, responsePromise]);
+    result = Promise.race([timeoutPromise, responsePromise]);
   }
+
+  io.emit('WorkRequest', request);
```
aerlaut, line 101: `io.emit` is moved here to address a potential bug where the response from API might return before we have successfully registered listeners.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
aerlaut: The other piece of data is already changed to using captured real data.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I have seen the `|| {}` pattern in other places, why not use it here then?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: but if config is available then `!config` will never be true, right? I mean it's an if that will never evaluate to true so we could just get rid of that

src/components/plots/styling/frequency/SelectCellSets.jsx
aerlaut: `key` is not used anymore see in Antd `Select`, so I'm changing it to value : `https://ant.design/components/select/`. I think it was legacy key, and the authors did not want to be break the API, so they kept the `key` there.

aerlaut: I'm removing this and the code above it which uppercases the value.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: yea, I just added it

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: ooh yea, I understand what you mean. Yea you're right, this code will never be called. I'll remove it.

src/__test__/components/plots/DotPlot.test.jsx
```diff
@@ -0,0 +1,75 @@
...
+  </Provider>
+);
+
+const config = initialPlotConfigStates.dotPlot;
+
+const state = {};
+
+describe('DotPlot', () => {
+  it('Renders a plot', async () => {
+    getCellSets.mockImplementation(() => () => ({
```
aerlaut: See if can use integration test

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
```diff
@@ -0,0 +1,122 @@
...
+    experimentId: fake.EXPERIMENT_ID,
+  }, customProps);
+
+  // eslint-disable-next-line react/jsx-props-no-spreading
+  return <DotPlotPage {...props} />;
+};
+
+let storeState;
+
+const mockFetchAPI = (req) => {
```
Comment (aerlaut) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx]: use the new intergration test framework

src/__test__/test-utils/backendStatus.mock.js
```diff
@@ -0,0 +1,31 @@
+const backendStatus = {
```
aerlaut: See if this can be recorded as JSON object under `data`

src/__test__/test-utils/markerHeatmapWorkResponse.mock.js
```diff
@@ -0,0 +1,46 @@
+const workResponse = {
```
aerlaut: Check if this one can be removed

src/__test__/test-utils/mockAPI.js
```diff
@@ -0,0 +1,60 @@
...
+  results: [
+    {
+      body: JSON.stringify(body),
+    },
+  ],
+  response: { error },
+}));
+
+const generateDefaultMockAPIResponses = (mockExperimentId) => ({
+  [mockExperimentId]: () => promiseResponse(
```
aerlaut: add `experiments/{experimentId}`

src/components/plots/DotPlot.jsx
```diff
@@ -0,0 +1,89 @@
...
+        AvgExpression: Math.random(),
+        cellsFraction: Math.random(),
+      });
+    }
+  }
+
+  return mockPlotData;
+};
+
+const plotData = generateMockData(3, 14);
```
aerlaut, line 29: Function to generate mock data for the plot

src/components/plots/styling/SelectData.jsx
```diff
@@ -109,9 +109,16 @@ const SelectData = (props) => {
     </>
   );
 };
+
 SelectData.propTypes = {
   onUpdate: PropTypes.func.isRequired,
   config: PropTypes.object.isRequired,
   cellSets: PropTypes.object.isRequired,
+  isValueForYAxis: PropTypes.bool,
```
aerlaut: Change signature

src/__test__/test-utils/markerHeatmapWorkResponse.mock.js
aerlaut: removed

src/__test__/test-utils/mockAPI.js
aerlaut: added

src/components/plots/styling/SelectData.jsx
aerlaut: changed

src/__test__/test-utils/experimentData.mock.js
```diff
@@ -0,0 +1,14 @@
+import fake from '__test__/test-utils/constants';
```
cosa65: I think it is better to just directly import the constants with `{}`.

cosa65: Also this file `__test__/test-utils/constants` I would call `__test__/test-utils/defaultAttributes` or something like that

src/__test__/test-utils/mockBackend.js
```diff
@@ -0,0 +1,67 @@
+import * as socketConnectionMocks from 'utils/socketConnection';
```
cosa65: could you do `import socketConnectionMocks from 'utils/socketConnection';` isntead of using the `import * as...` pattern?

src/__test__/test-utils/mockBackend.js
```diff
@@ -0,0 +1,67 @@
...
+import * as seekWorkResponseMocks from 'utils/work/seekWorkResponse';
+
+import SocketMock from 'socket.io-mock';
+
+const socketMock = new SocketMock();
+
+jest.mock('utils/work/seekWorkResponse', () => {
```
cosa65: why do we mock `seekWorkResponse` if we are mocking s3 and the api?

src/__test__/test-utils/mockWorker.js
```diff
@@ -0,0 +1,15 @@
+/* eslint-disable import/prefer-default-export */
+const mockETag = (
```
cosa65: The file is called mockWorker but it exports `mockETag`, maybe renmae it to `mockETag`? And export `mockETag` by default

src/__test__/utils/work/seekWorkResponse.test.js
```diff
@@ -76,9 +77,23 @@ describe('seekFromAPI unit tests', () => {
...
+        response: {
+          error: false,
+        },
+      };
+
+      // This is a mocked response emit response from server
+      socketMock.socketClient.emit(`WorkResponse-${requestBody.ETag}`, responseBody);
+    });
+
+    socketConnectionMocks.mockOn.mockImplementation((channel, f) => {
```
cosa65: what is `f`?

src/components/plots/DotPlot.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+const DotPlot = (props) => {
+  const { config } = props;
+
+  const [numClusters, setNumClusters] = useState(0);
+
+  const { loading: cellSetsLoading, error: cellSetsError, hierarchy } = useSelector(getCellSets());
+
+  useEffect(() => {
+    if (Object.keys(hierarchy).length === 0) return;
```
cosa65: can you check `cellSetsLoading` instead? It is a more explicit check on the same hting, right?

src/components/plots/DotPlot.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+    const clusters = _.find(
```
cosa65: Can this not use one of the selectors perhaps?

src/components/plots/styling/MarkerGeneSelection.jsx
```diff
@@ -0,0 +1,97 @@
...
+  const [numGenes, setNumGenes] = useState(config.nMarkerGenes);
+  const [genesLoaded, setGenesLoaded] = useState(true);
+
+  useEffect(() => {
+    if (numGenes === config.numGenes) {
+      setGenesLoaded(true);
+    } else if (numGenes !== config.numGenes && genesLoaded) {
+      setGenesLoaded(false);
+    }
+  }, [config.numGenes, numGenes]);
```
cosa65: what is `config.numGenes` used for as opposed to `config.nMarkerGenes`? I see we have a `useState` `numGenes` initialized with `config.nMarkerGenes` that is then compared to `config.numGenes`.

Could we have a call to go over this tomorrow?

src/components/plots/styling/SelectData.jsx
```diff
@@ -7,12 +7,13 @@ import {
   Select,
 } from 'antd';
 
-import composeTree from '../../../../utils/composeTree';
+import composeTree from '../../../utils/composeTree';
```
cosa65: We have absolute filepath imports

src/components/plots/styling/SelectData.jsx
```diff
@@ -109,9 +109,16 @@ const SelectData = (props) => {
...
+
 SelectData.propTypes = {
   onUpdate: PropTypes.func.isRequired,
   config: PropTypes.object.isRequired,
   cellSets: PropTypes.object.isRequired,
+  isValueForYAxis: PropTypes.bool,
 };
+
+SelectData.defaultProps = {
+  isValueForYAxis: true,
```
cosa65: Reminder to change this for an enum as discussed in the call

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -0,0 +1,178 @@
...
+const dotPlot = (props) => {
+  const { experimentId } = props;
+
+  const dispatch = useDispatch();
+  const config = useSelector((state) => state.componentConfig[plotUuid]?.config);
+  const {
+    fetching: genesFetching,
+    data: highestDispersionGenes,
+  } = useSelector((state) => state.genes.properties.views[plotUuid] || {});
+  const cellSets = useSelector((state) => state.cellSets);
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:79]: in other places you are destructuring all the properties of `cellSets` instead of using the `cellSets` object, also here the genes are being destructured, I would suggest deciding on one pattern and going with it everywhere

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -0,0 +1,178 @@
...
+
+  const PROPERTIES = ['dispersions'];
+  const tableState = {
+    pagination: {
+      current: 1, pageSize: config?.nMarkerGenes ?? 3, showSizeChanger: true, total: 0,
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: `config?.nMarkerGenes ?? 3` shouldn't this default `3` be set in its initialState?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -0,0 +1,178 @@
...
+          option: {
+            positions: 'top-bottom',
+          },
+        },
+      },
+    ],
+  },
+];
+
+const dotPlot = (props) => {
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: use `DotPlot` for a component

Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: also perhaps call it something differnet, the actual `DotPlot` is a different component, this one is a `DotPlotPage` or something.

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
```diff
@@ -7,15 +7,17 @@ import {
...
+import Header from 'components/Header';
+
 import heatmap from '../../../../../public/static/media/heatmap.png';
 import embeddingContinuous from '../../../../../public/static/media/embeddingContinuous.png';
 import embeddingCategorical from '../../../../../public/static/media/embeddingCategorical.png';
 import violin from '../../../../../public/static/media/violin.png';
+import dotPlot from '../../../../../public/static/media/dotplot.png';
 import volcano from '../../../../../public/static/media/volcano.png';
 import frequency from '../../../../../public/static/media/frequency.png';
 import markerHeatmap from '../../../../../public/static/media/marker_heatmap.png';
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/index.jsx:20]: use absolute imports

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -1,15 +1,26 @@
...
 import PropTypes from 'prop-types';
 import pushNotificationMessage from 'utils/pushNotificationMessage';
 import endUserMessages from 'utils/endUserMessages';
+
 import { getCellSets, getCellSetsHierarchyByKey } from 'redux/selectors';
+import getSelectOptions from 'utils/plots/getSelectOptions';
+import MarkerGeneSelection from 'components/plots/styling/MarkerGeneSelection';
 import loadProcessingSettings from '../../../../../redux/actions/experimentSettings/processingConfig/loadProcessingSettings';
 import PlotStyling from '../../../../../components/plots/styling/PlotStyling';
 import { updatePlotConfig, loadPlotConfig } from '../../../../../redux/actions/componentConfig';
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: use absolue imports

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -20,7 +31,6 @@ import { loadCellSets } from '../../../../../redux/actions/cellSets';
 import PlatformError from '../../../../../components/PlatformError';
 import Loader from '../../../../../components/Loader';
 import populateHeatmapData from '../../../../../components/plots/helpers/populateHeatmapData';
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: use absolute imports

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -83,15 +96,19 @@ const MarkerHeatmap = ({ experimentId }) => {
         pushNotificationMessage('error', endUserMessages.NO_CLUSTERS);
       }
     }
-  }, [config?.selectedCellSet, config?.numGenes, hierarchy]);
+  }, [config?.selectedCellSet, config?.nMarkerGenes, hierarchy]);
 
   useEffect(() => {
     if (louvainClustersResolution
       && louvainClustersResolutionRef.current !== louvainClustersResolution
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: We might be able to remove this `useRef` and its comparison: `louvainClustersResolutionRef.current !== louvainClustersResolution` by adding a comparer in `useSelector` like this: `useSelector(..., _.isEqual)`, could you try and see if it actually works?

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -315,10 +352,78 @@ const MarkerHeatmap = ({ experimentId }) => {
...
 
+  const getCellOptions = (type) => {
+    const filteredOptions = hierarchy.filter((element) => (
+      properties[element.key].type === type
+    ));
+    if (!filteredOptions.length) {
+      return [];
+    }
+    return filteredOptions;
+  };
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: We might be able to use a selector for this, right?

src/utils/plotSpecs/generateDotPlotSpec.js
```diff
@@ -0,0 +1,217 @@
+// The goal is to minimize space with the largest possible dot size along the larger axis
+const getDotDimensions = (config, numClusters) => {
+  const plotWidth = config.dimensions.width;
+  const plotHeight = config.dimensions.height;
+  const padding = 1;
```
cosa65, line 5: Just checking: this padding doesn't depend on any other values?

cosa65 (changes requested):

src/__test__/test-utils/experimentData.mock.js
aerlaut: Hmm.. I think it's better to prefix the constants with `fake` as it would emphasize that it is fake data. I also wouldn't call it default attributes because it's something that we set. I don't want it to be confused as if it's something that's default to the system. I'm following a pattern that's done in several places in the API, and I think it's good,

src/__test__/test-utils/mockBackend.js
aerlaut: No, because we want to import everything (default and named exports) from `utils/socketConnection`, not only the default export. Doing `import socketConnectionMocks from 'utils/socketConnection';` only assigns the default export to `socketConnectionMocks`.

aerlaut: for reference : https://medium.com/@etherealm/named-export-vs-default-export-in-es6-affb483a0910

src/__test__/test-utils/mockBackend.js
aerlaut: This is so that we can mock `seekFromS3` to return `null` on line 54 inside `fetchWork`. Returning `null` from `seekFromS3` forces the component to execute `seekFromAPI` in `fetchWork` line 141. Inside `seekFromAPI`, we have`seekFromS3` on line 88, but we can not mock this `seekFromS3` because it is internal to the module, it is not an import. So this call to `seekFromS3` will call the real function body at `seekWorkResponse` line 9, and do a `fetch` on line 18, which we then handle with `fetchMock`.

src/__test__/test-utils/mockWorker.js
aerlaut: Good idea. I'll rename it to `mockWorkResultEtag` to make it clearer what the `ETag` we are mocking is for.

src/__test__/utils/work/seekWorkResponse.test.js
aerlaut: `f` is the callback function that is passed as the 2nd parameter of `io.on` on line 58 of `seekWorkResponse`.  We repalce `io.on` on line 28 of `seekWorkResponse.test.js` to use `mockOn`, which is a `jest.fn()`. On line 58 of `seekWorkResponse`, the mocked function receives 2 params : `${experimentId}-{bodyName}`, and `f`, the callback function. So inside the `mockOn` implementation, we call `f` again to continue the execution.

src/components/plots/DotPlot.jsx
aerlaut: I think checking for the length more explicit because we are checking for the actual data that will be used later on (`hierarchy`) instead of a proxy (`cellSetLoading`). That said, if we want to be more explicit, we can check for both : that `cellSetsLoading !== false && hierarchy.length > 0`. Only checking for `cellSet` would make the condition return `false` if there is an error., and the following cold block run. What do you think?

src/components/plots/DotPlot.jsx
aerlaut: Yea, this can use the selector.

src/components/plots/styling/MarkerGeneSelection.jsx
aerlaut: whoops, `config.numGenes` shouldn't be used anymore. it should be `config.nMarkerGenes`. Thanks for catching this!

src/components/plots/styling/SelectData.jsx
aerlaut: this is already changed, but not yet pushed

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Yea, it's already set inside the `initialConfig` inside the `dotPlot` component. It was set here as I was working on the DotPlot in the beginning. This also had to be done because the constant variable is put outside and there is no check for the existence of `config`. I think I can move this closer  to`loadPaginatedGeneProperties`, the place where it's called, where there is also a check for the existence for `config` .

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: The other plots have "Index" ending in them. Better end it with "Page" actually,

src/utils/plotSpecs/generateDotPlotSpec.js
aerlaut: No, it's a free value we set.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: yup

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/index.jsx:20]: Turns out this can not be changed to absolute imports. The `public` folder is on the same level of `src`. Doing a relative `../public/static...` is not recognized by the linter. I can make the `public` folder as the, but I don't think this is a good idea. It leaves the remaining file path to `static/media/...`, and it's also confusing (is it `src/static/media`? or `public/static/media`?)

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I think we are doing this because we want to prevent rerender that is caused by other changes (not changes to `louvainClusterResolution`).

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
kafkasl: how has this been addressed? I don't see that you added the insensitive case selector option

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
kafkasl: ok

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
kafkasl: I meant that this is outside a test case delimited by `it` so it looks weird. I assume you do that to check to ensure the expect is called for every test calling `renderHeatmapPage`

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
kafkasl: try to set how many time you expect it to be called

aerlaut: This is not there anymore

src/components/plots/styling/MarkerGeneSelection.jsx
aerlaut: This is not used anymore. It was used to disable the input to select the number of marker genes to show if the selected number of genes per cluster is the same as what is currently selected. This is replaced with direct comparison on line 54.

src/__test__/test-utils/backendStatus.mock.js
aerlaut: removed

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: I remove this because the marker heatmap plot does not have controls to change `louvainClusterResolution`. I don't know why this code is still here. `louvainClusterResolution` is changed in the last step of configure embedding, and the heatmap plot is not shown there. So this can be removed

src/__test__/test-utils/mockBackend.js
cosa65: but you are also mocking amplify, right? if you mock `seekFromS3` you don't need to mock amplify because it never gets called

src/components/plots/DotPlot.jsx
cosa65: you're right about error, so perhaps you should check for loading, error and length === 0?

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/index.jsx:20]: hmm maybe we should have set the root for aboslute paths at one level higher than `src/`

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: > I think we are doing this because we want to prevent rerender that is caused by other changes (not changes to louvainClusterResolution).

the `useEffect` only depends on `louvainClustersResolution`, so adding `_.isEqual` in the `useSelector`'s second param would effectively do the same

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: The line is removed because we don't use it anymore

src/__test__/test-utils/mockBackend.js
aerlaut: Mocking `seekFromS3` only mocks the function inside `fetchWork` but can not mock `seekFromS3` inside `seekWorkResponse`. The `seekFromS3` inside `seekWorkResponse` calls `amplify`, so we still need to mock `amplify`

src/components/plots/DotPlot.jsx
aerlaut: Implemented

src/components/plots/styling/SelectData.jsx
aerlaut: updated to use absolute file paths

src/components/plots/styling/SelectData.jsx
aerlaut: pushed in later commit

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:79]: added line below to destructure cellSets

src/components/plots/DotPlot.jsx
aerlaut: Turns out this is not used anymore because we are using the `getHierarchyByKey` selector which returns the array

src/components/plots/styling/MarkerGeneSelection.jsx
```diff
@@ -0,0 +1,88 @@
...
+import propTypes from 'prop-types';
+import {
+  Space,
+  Radio,
+  InputNumber,
+  Select,
+  Button,
+} from 'antd';
+
+const MarkerGeneSelection = (props) => {
```
cosa65, line 11: I couldn't find tests for this comp, could you add? It should be pretty straightforward, just checking the `on` and the render based on `config.useMarkerGenes` or not

package.json
```diff
@@ -115,14 +115,15 @@
...
+    "socket.io-mock": "^1.3.2",
     "styled-components": "^5.3.1",
     "swr": "^0.3.2",
     "terser-webpack-plugin": "^1.1.0",
     "thread-loader": "^2.1.3",
     "uuid": "^8.1.0",
     "vega": "^5.13.0",
     "vega-lite": "^4.13.0",
-    "vitessce": "^0.2.1",
+    "vitessce": "^0.2.5",
```
cosa65, line 126: We had talked about removing this version change because it broke the embedding, right?

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+    enableFetchMocks();
+    fetchMock.resetMocks();
+    fetchMock.doMock();
+    fetchMock.mockIf(/.*/, mockAPI(defaultMockResponses));
+    storeState = makeStore();
+
+    storeState.dispatch(loadBackendStatus(fake.EXPERIMENT_ID));
+  });
+
+  it('Renders the plot page correctly', async () => {
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:58]: there's some conditional rendering that isn't being tested here (case `!config`, case `config && !cellSetsLoading && !cellSetsError`)

cosa65 (changes requested):

package.json
aerlaut: I tried again in staging and it doesn't break the embedding, so I'm keeping it.

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:58]: I'm handling the first case here (`!config`). The 2nd case should be hanled by the `SelectData` component I think, so I'm moving the loading state inside that component.

src/components/plots/styling/MarkerGeneSelection.jsx
aerlaut: added

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:58]: The 2nd case is not yet tested, I think it should be added with tests for the `SelectData` component. Maybe with ticket https://biomage.atlassian.net/browse/BIOMAGE-1511?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -126,14 +126,12 @@ const DotPlotPage = (props) => {
         />
       </Panel>
       <Panel header='Select data' key='15'>
-        {config && !cellSetsLoading && !cellSetsError ? (
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Check is moved inside `SelectData`.

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (cosa65) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:58]: Ok, coudl you update that ticket?

cosa65 (approved):

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:58]: Updated telling the one who picks up the ticket to write tests.

src/__test__/components/plots-and-tables/markerHeatmap.test.jsx
aerlaut: Yup, that line is removed.

src/__test__/components/plots-and-tables/ViolinPlotIndex.test.jsx
aerlaut: I have changed selectors to use case insensitive matchers.

aerlaut: Clicking "view changes" links to the old version, it doesn't show the changes. To see the newest version of the file you have to see it under "Files changed"

PR #531, opened 2021-10-11
changed cellscope name to cellenics

kafkasl:
Only changed Cellscope to cellenics, all other changes are automatic sorting of imports by the linter.


marcellp (approved):
lgtm

PR #538, opened 2021-10-25
[BIOMAGE-1539] Fix handling of large work results

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1539

https://ui-martinfosco-ui538-worker20.scp-staging.biomage.net/

https://github.com/biomage-ltd/worker/pull/206

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


cosa65:
The actual problem turned out to be not that the `markerGenes` json response to parse was 500mB+ but that the R worker's response is stringified and wrapped in a new json.

In the UI is was necessary to do first a `response.json()` and later on a `JSON.parse(response.results[0].body)` to get the R worker's data.

The problem was that, during the first json parse: `response.json()`, nodejs found the actual task's response (that is stored as a string) and in very big responses this string is bigger than the string length limit set in nodejs, so it would error out.

To fix this at first I considered parsing the R worker response in the python worker before storing it in s3, but it seems that the whole structure we are using to store the work response is not used at all since in the UI we always do `response.results[0].body`, si I decided to move this unpacking to the python worker (where the error/cacheable properties are last checked) and just directly store in s3 the json response.

So now the UI just gets what it needs so that there isn't a stringified json inside another json (and we also avoid having to do 2 json parses for each response).

src/__test__/utils/work/seekWorkResponse.test.js
```diff
@@ -98,8 +92,8 @@ describe('seekFromAPI unit tests', () => {
 
     expect(socketConnectionMocks.mockOn).toHaveBeenCalledTimes(1);
     expect(response).toEqual({
-      results: [{ body: '{"hello":"world"}' }],
-      response: { error: false },
```
aerlaut, line 102: Is it ok to remove the error flag?

ivababukova (approved):

src/__test__/utils/work/seekWorkResponse.test.js
cosa65: As far as I could see, in the UI we don't use the error flag at all

cosa65: In fact, if you look at `def publish(self):` in the python worker, it calls `__upload(self):` (which performs the upload to s3), but it calls `_upload` only if error is false

aerlaut: ok, so error is not used in the UI 👍

PR #542, opened 2021-10-29
[HACKATHON] - Add new dataset and generate mock test data

aerlaut:
# Background
#### Link to issue 

- I'm working on a new article to give a Biomage tutorial. In the tutorial. To showcase every part of the platform, we need a new sample dataset that is multisample and unfiltered, so the platform can be showcased. This is the PBMC BMMC dataset.
- I noticed that the "Process Project" button is enabled by default. This button will only change if the `backendStatus` has been loaded. If the backendStatus shows that the project is processed, the button will change to "Go to Data Processing", skipping GEM2S. This allows for a brief time window where the user can re-run GEM2S on an already-processed dataset. In this PR, I'm addressing this issue by disabling the button before the required data is available. 

https://ui-agi-ui542.scp-staging.biomage.net/

- Add a new sample dataset: PBMC BMMC. Change the available sample datasets into a list.
- Add loading state to launch analysis button to prevent accidental clicks to process project in case of slow connection.
- Add improvements to how mock test data is generated

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [x] Unit tests written
- [X] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-management/SamplesTable.jsx
```diff
@@ -275,12 +275,25 @@ const SamplesTable = forwardRef((props, ref) => {
     setTableData(newData);
   }, [projects, samples, activeProjectUuid]);
 
-  const downloadPublicDataset = async () => {
-    const s3Object = await Storage.get('PBMC_3k.zip',
+  const sampleData = [
```
kafkasl: this is a list of datasets not samples, change the name

Is there any reason to add all this code to be able to download 2 datasets instead of just changing the default one?

aerlaut: Whoops, it should be `exampleDatasets`. I was confusing `sample` with `example`.

aerlaut: I talked to Vicky about replacing the default file to use the new example dataset. She thinks that we shouldn't remove access to the original 3k PBMC dataset. I agree with her - there is merit to still have the 3k PBMC dataset. It is smaller and faster to process, compared to the PBMC BMMC dataset. So, if we only need to demo quickly, we can use the 3k PBMC dataset.

Having the list will also allow us to easily add more example datasets in the future, if we want to show case a specific part of the platform (e.g. spatial?)

src/__test__/test-utils/mockAPI.js
```diff
@@ -44,6 +47,15 @@ const generateDefaultMockAPIResponses = (experimentId) => ({
   [`experiments/${experimentId}/backendStatus`]: () => promiseResponse(
     JSON.stringify(backendStatusData),
   ),
+  '/experiments': () => promiseResponse(
+    JSON.stringify(experimentsData),
+  ),
+  '/projects': () => promiseResponse(
+    JSON.stringify(projectsData),
+  ),
+  'projects/380c53b4-mock-test-8091-4c82336d6d49/samples': () => promiseResponse(
```
cosa65: Do you think you could make it `projects/${projectId}/samples` as you did with `experiments/${experimentId}/backendStatus`

src/components/data-management/LaunchAnalysisButton.jsx
```diff
@@ -1,31 +1,36 @@
...
-import { updateExperiment } from '../../redux/actions/experiments';
+import fileUploadSpecifications from 'utils/upload/fileUploadSpecifications';
+import UploadStatus from 'utils/upload/UploadStatus';
+import pipelineStatus from 'utils/pipelineStatusValues';
+import integrationTestConstants from 'utils/integrationTestConstants';
+import generateGem2sParamsHash from 'utils/data-management/generateGem2sParamsHash';
+import { runGem2s } from 'redux/actions/pipeline';
+import { updateExperiment } from 'redux/actions/experiments';
 
 const LaunchButtonTemplate = (props) => {
```
cosa65, line 21: Why is this split into a different component? you could move it into the `LaunchAnalysisButton` with a `renderButton()` or something, right?

I am not sure what is considered best practice, but I saw we do that in other places in the ui.

src/components/data-management/SamplesTable.jsx
```diff
@@ -376,3 +396,7 @@ SamplesTable.propTypes = {
 };
 
 export default React.memo(SamplesTable);
```
cosa65, line 399: Out of curiosity, do you know why there is a `React.memo` here?

src/__test__/test-utils/mockAPI.js
aerlaut: I can for this one, but it's not going to be very sustainable. We have to have a way to generate mock data soon.

src/components/data-management/LaunchAnalysisButton.jsx
aerlaut: I decided to make this into a JSX Element because the function signature looks better. For example, if this were to be turned into a function `renderButton(text, disabled, loading, onClick)`, there will be a place where the function call is `renderButton('Launch Analysis', null, null, () => { launchAnalysis() })` which looks ok but not pretty.

src/components/data-management/SamplesTable.jsx
aerlaut: No idea. I think it was just an effort to improve the `SamplesTable` performance. I don't think it's useful as `React.memo` only check for `prop` changes, not changes due to `useEffect` or `useSelector` (https://reactjs.org/docs/react-api.html#reactmemo). I think this can be removed?

src/components/data-management/LaunchAnalysisButton.jsx
cosa65: Ok, we should probably decide on one way to do this and keep it in mind when working in other places.

If we are going with having auxiliary components as opposed to auxiliary render functions we should change the auxiliary render functions to components wehen we run into them in the future.

src/components/data-management/SamplesTable.jsx
cosa65: I do know that there's some rerendering happening in the parent component of `SamplesTable` (it's what I was working on in the upload improvement PR), maybe this is some kind of patch for that.

cosa65: I would keep it for now

cosa65 (approved):

PR #543, opened 2021-11-01
[BIOMAGE-546] Stop hardcoding UMAP embedding

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/jira/software/c/projects/BIOMAGE/boards/1?modal=detail&selectedIssue=BIOMAGE-546

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
```diff
@@ -0,0 +1,15 @@
+const changeEmbeddingAxesIfNecessary = (config, method, onUpdate) => {
+  if (!config || !method) return;
+  const { defaultValues: axesDefaultValues, xAxisText, yAxisText } = config?.axes;
```
ivababukova: no need for `config?.axes` -- if config is null, the code would have exited on line 2 and never entered here

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
```diff
@@ -0,0 +1,15 @@
...
+
+  if (axesDefaultValues?.length) {
+    const methodUppercase = method[0].toUpperCase() + method.slice(1);
```
ivababukova, line 6: can you give me an example of a value for `methodUppercase`? Also an example for values for `method`?

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
```diff
@@ -0,0 +1,15 @@
+const changeEmbeddingAxesIfNecessary = (config, method, onUpdate) => {
```
ivababukova, line 1: I would name this `renameEmbeddingAxes`

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -41,18 +42,42 @@ const ExplorationViewPage = ({
   const { windows, panel } = layout;
   const [selectedTab, setSelectedTab] = useState(panel);
   const [addMenuVisible, setAddMenuVisible] = useState(false);
-
+  const { method } = useSelector((state) => (
+    state.experimentSettings.processing?.configureEmbedding?.embeddingSettings || ' '
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: Looking at the code, I don't see the need for adding this default of `' '`. You are checking everywhere if method is null.

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -41,18 +42,42 @@ const ExplorationViewPage = ({
...
+          first: {
+            ...windows.first.first,
+            first: `${methodUppercase} Embedding`,
+          },
+        },
+      }));
+    }
+  }, [method]);
+
+  const embeddingTileName = `${methodUppercase} Embedding`;
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: I suggest you rename this `embeddingTileName` to `embeddingTitle` and move it to be defined on line 58, then use it also in line 67.

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -41,18 +42,42 @@ const ExplorationViewPage = ({
...
+        first: {
+          ...windows.first,
+          first: {
+            ...windows.first.first,
+            first: `${methodUppercase} Embedding`,
+          },
+        },
+      }));
+    }
+  }, [method]);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: I think we can drastically simplify this approach. 
For a user to change the embedding method, they need to go back to data processing, to the calculate embedding plot and change one of the settings. Then this would trigger data processing pipeline and only after that they will be able to go to data exploration. This means that this use-effect will never be triggered -- change of the method is not possible without leaving data exploration. 

There is an object in this file called `categoryItems`. This object defines what each of the windows in data exploration is, where `key` denotes the title of the window. For your case, I think that all you need to do is to change the `key` for the first item under `Plots` to be equal to `${methodUppercase} Embedding`. And then you can delete this useEffect

src/redux/reducers/layout/initialState.js
```diff
@@ -3,7 +3,7 @@ const initialState = {
     direction: 'row',
     first: {
       first: {
-        first: 'UMAP Embedding', second: 'Data Management', direction: 'row', splitPercentage: 60,
+        first: 'UMAP Embeddingg', second: 'Data Management', direction: 'row', splitPercentage: 60,
```
ivababukova: typo

ivababukova: you can remove `UMAP` from the initial state, just have `Embedding`

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
StefanBabukov: method can be tsne or umap, so methodUppercase - Tsne, Umap

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
StefanBabukov: but we also check so they don't change every time

src/pages/experiments/[experimentId]/data-exploration/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/data-exploration/index.jsx]: No this doesn't work, because the layout of the trees is in the store so this is what i firtly tried to do. It didn't work it was not rendering the windows properly, as it was a misscorrelation between the values in the initialState file. I can show you what I mean.

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
ivababukova: that is true. It is just a suggestion, you can keep the name as it is

src/components/plots/helpers/changeEmbeddingAxesIfNecessary.js
ivababukova: so why does it have to be a list?

ivababukova (approved):

PR #546, opened 2021-11-02
[BIOMAGE-1451] - Dot plot backend

aerlaut:
# Background
#### Link to issue

https://biomage.atlassian.net/browse/BIOMAGE-1451

https://ui-agi-dotplot-final-final.scp-staging.biomage.net/

UI - https://github.com/biomage-ltd/ui/pull/546
API - https://github.com/biomage-ltd/api/pull/253
Worker - https://github.com/biomage-ltd/worker/pull/209

# Changes
### Code changes

# Definition of DONE

- Add actions to fetch DotPlot data from the backend

### Code updates

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/work/fetchWork.js
```diff
@@ -143,13 +143,15 @@ const fetchWork = async (
   });
 
   // First, let's try to fetch this information from the local cache.
-  const data = await cache.get(ETag);
-  if (data) {
-    return data;
-  }
+  // const data = await cache.get(ETag);
```
kafkasl: why is this commented? should it be removed?

src/utils/work/fetchWork.js
```diff
@@ -143,13 +143,15 @@ const fetchWork = async (
...
+  // if (data) {
+  //   return data;
+  // }
 
   // Then, we may be able to find this in S3.
-  let response = await seekFromS3(ETag);
+  // let response = await seekFromS3(ETag);
```
kafkasl: same here

src/utils/work/generatePlotWorkBody.js
```diff
@@ -0,0 +1,26 @@
...
+    cellClassKey: config.selectedCellSet,
+    cellSetKey: config.selectedPoints,
+  },
+});
+
+const generatePlotWorkBody = (plotType, config) => {
+  switch (plotType) {
+    case plotNames.plotType.DOT_PLOT:
+      return composeDotPlotWorkBody(config);
+    default: {
```
kafkasl, line 23: why this switch? is this something you envision will be re-used by other plots at some point?

src/__test__/utils/work/generatePlotWorkBody.test.js
```diff
@@ -0,0 +1,16 @@
+import plotNames from 'utils/plots/plotNames';
+import generatePlotWorkBody from 'utils/work/generatePlotWorkBody';
+
+import { initialPlotConfigStates } from 'redux/reducers/componentConfig/initialState';
+
```
kafkasl, line 5: add ideally two more test cases, one with a wrong plotName, the other with invalid config (to make sure it does not crash the whole UI).

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -89,40 +104,81 @@ const DotPlotPage = (props) => {
...
-      const tableState = {
-        pagination: {
-          current: 1, pageSize: config.nMarkerGenes, showSizeChanger: true, total: 0,
-        },
-        geneNamesFilter: null,
-        sorter: { field: PROPERTIES[0], columnKey: PROPERTIES[0], order: 'descend' },
-      };
+  const currentConfig = useRef(null);
+
+  const getCurrentDataProps = (conf) => ({
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: why do we need this function? what extra fields does `conf` contain that are not needed? do they need to be removed?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -89,40 +104,81 @@ const DotPlotPage = (props) => {
...
+  });
 
-      dispatch(loadPaginatedGeneProperties(experimentId, PROPERTIES, plotUuid, tableState));
+  const sameConfig = () => _.isEqual(
+    currentConfig.current,
+    getCurrentDataProps(config),
+  );
+
+  useEffect(() => {
+    if (config && !sameConfig()) {
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: try to avoid these kind of functions `sameConfig()` where instead of passing parameters you rely on getting them from the class. It's harder to test and reason about these functions. If you have the function `sameConfig(cfg1, cfg2)` it is very easy to write a test for that. Instead, with `sameConfig()` how do you test it? you need to mock the class to get the `currentConfig.current` and then `config` is not even defined. 

Moreover you are using `getCurrentDataProps(config)` but `config` is not a parameter so you are relying on the variable having the same name in the caller function which is very brittle and will break if someone changes the name.

kafkasl (dismissed):
I don't see any test for the dotPlot component itself. any reason for that?

src/utils/work/fetchWork.js
aerlaut: Yes, this was commented to bypass cache in development. Uncommented

src/__test__/utils/work/generatePlotWorkBody.test.js
aerlaut: Thank you for the input on checking for error. The function will throw an error if no configs match the passed plot name. I've added a test for that.

I don't think we can check for the work request body created by this function, unless we want to have some kind of schema validator like we have in the API. So in the case of invalid wokRequest body, the API will return an error response to the UI (validation error)

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: This is a helper function to make it easy to get the props that defines the plot's data. The idea is we want the plot to re-fetch the plot's data if these data changes. If other plot config changes, we want the plot to update, but not refetch the data.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I understand. I'm removing the wrapper function to remove the dependency.

`getCurrentDataProps` (renamed to `getPlotDataConfig`) actually takes and use a parameter - `conf`. It does not use the component's `config` object. I move the function out of the component body, so it's clearer.

src/utils/work/generatePlotWorkBody.js
aerlaut: Yes. Other plots calculate plot data from data that is already available on the platform (e.g. gene expression), so they do not need to define their own work request body. This is the first plot that has its own work request to specifically calculate data for the plot. I think this functionality will be needed for new plots in the platform.

aerlaut:
> I don't see any test for the dotPlot component itself. any reason for that?

This test is already implemented in `DotPlot.test.jsx`, it was added in PR #529. I think it has tested for cases that are available in `DotPlot.jsx`.

src/__test__/components/plots/styling/MarkerGeneSelection.test.jsx
```diff
@@ -55,6 +55,7 @@ describe('MarkerGeneSelection', () => {
 
   it('Should show the number of marker genes input', async () => {
     const mockConfig = _.merge(
+      {},
```
kafkasl: what's the point of merging something empty? is the whole `_.merge` is needed at all?

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
```diff
@@ -99,4 +123,26 @@ describe('Dot plot page', () => {
...
+
+    await act(async () => {
+      render(
+        <Provider store={storeState}>
+          {dotPlotPageFactory()}
+        </Provider>,
+      );
+    });
+
+    expect(screen.getByText(/Error loading cell sets/i)).toBeInTheDocument();
```
Comment (kafkasl) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:157]: for this test file have you tried what we discussed about checking that tests are doing something? e.g. pass a correct config response in the last one and expect it to get an error (so making sure it fails), and so on

src/__test__/components/plots/styling/MarkerGeneSelection.test.jsx
aerlaut: `_.merge` modifies the first object that's passed to it, and also returning the modified object. If i do `_.merge(a, b)`, the object `a` is modified with changes in `b`. If I do `_.merge({}, a, b)`, `a` is not modified, and the function returns a new object that is a combination of `a` and `b`.

In this case, I don't want to modify the `initialPlotConfigStates.dotPlot`, but I want an object that contains `initialPlotConfigStates.dotPlot` and `{ useCustomGenes: false }`. It is like doing 
      
```
{
   ...initialPlotConfigStates.dotPlot,
   useCustomGenes: false
}
```

src/__test__/redux/reducers/componentConfig/loadPlotData.test.js
```diff
@@ -0,0 +1,63 @@
...
+  it('Throws an error if there is an error fetching plot data', async () => {
+    fetchWork.mockImplementation(() => {
+      throw new Error('Error fetching work');
+    });
+
+    await act((async () => {
+      await storeState.dispatch(loadPlotData(experimentId, plotUuid, plotType));
+    }));
+
+    expect(pushNotificationMessage).toHaveBeenCalled();
```
kafkasl: 1. add how many times do you expect the function to be called
2. is checking if the function has been called enough to test that `throws an error if there is an error fetching plot data`?

src/redux/reducers/componentConfig/initialState.js
```diff
@@ -754,7 +754,18 @@ const initialPlotConfigStates = {
 const initialComponentConfigStates = {
   interactiveHeatmap: interactiveHeatmapInitialConfig,
 };
-export { initialPlotConfigStates, initialComponentConfigStates };
+
+const initialPlotDataState = {
```
kafkasl, line 774: where was the initial plot data defined before? is this supposed to be reused (is it already?) by other plots? does it match the other plots' initial data state?

src/redux/reducers/componentConfig/plotDataLoading.js
```diff
@@ -0,0 +1,16 @@
+import { initialPlotDataState } from './initialState';
+
+const plotDataLoading = (state, action) => {
```
kafkasl, line 1: why are we storing this? to show some placeholder while we load? what did we do before that?

src/utils/plots/plotNames.js
```diff
@@ -0,0 +1,26 @@
+const plotType = {
```
kafkasl: this contains more than plot names, I would name the file `contants.js`. It can stay in the plots folder but I believe it's best to place it directly inside the utils folder because otherwise it's easier to miss it, and the next developer might add his contacts somewhere else

src/utils/work/generatePlotWorkBody.js
kafkasl: in this situation it probably makes sense as I can foresee having a new plot that needs worker data. However, be very wary of coding for features not needed because it just adds extra complexity that might not be needed or used in the end. https://en.wikipedia.org/wiki/KISS_principle

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:157]: yes, I've made sure the tests fail if it's passed the correct response

src/redux/reducers/componentConfig/plotDataLoading.js
aerlaut: Yes, to show placeholder while the plot is being loaded. Before this, we are checking for `!plotData.length` to show that the item is loading. I think having this `loading` state is much better, because we might have cases where plot data is empty (there actually is no data) while we are not loading anymore.

src/redux/reducers/componentConfig/initialState.js
aerlaut: `plotData` was initially added as the payload in `loadPlotConfig` when data is being fetched from the server. `plotData` is used to pass data from the server to the plots. Not all plots have `plotData`. Right now, only plots in the filters in `data-processing` uses `plotData`.

src/__test__/redux/reducers/componentConfig/loadPlotData.test.js
aerlaut: Whoops, it should display notification, not throw an error.

src/redux/reducers/componentConfig/initialState.js
aerlaut: I have modified `loadPlotConfig` to make it. I have also inserted `initialPlotDataState` when loading plot config. This way, all plots have the same states, although they might not use it.

src/__test__/redux/reducers/componentConfig/loadPlotData.test.js
aerlaut: I've also modified `toHaveBeenCalled()` into `toHaveBeenCalledTimes(1)`

src/__test__/components/plots/styling/MarkerGeneSelection.test.jsx
kafkasl: and why you didn't use? it looks cleaner and easier to read, is there any drawback?
```
{ ...initialPlotConfigStates.dotPlot, useCustomGenes: false }
```

aerlaut: `_.merge()` is useful when merging objects with multiple layers. In this case it isn't, so it's over the top. I'll change it to use object spreading.

aerlaut: Changed to use simpler merge

src/components/plots/styling/MarkerGeneSelection.jsx
```diff
@@ -15,62 +15,62 @@ const MarkerGeneSelection = (props) => {
   const [numGenes, setNumGenes] = useState(config.nMarkerGenes);
 
   const renderOptions = () => {
-    if (config.useCustomGenes) {
```
aerlaut, line 18: This is just switching to make it easier to read

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
```diff
@@ -0,0 +1,108 @@
...
+      await testStore.dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
+    });
+
+    // Expect componentConfig to contain key for plotUuid
+    expect(testStore.getState().componentConfig[plotUuid]).toBeDefined();
+
+    const plotConfig = testStore.getState().componentConfig[plotUuid];
+
+    expect(plotConfig.config).toBeDefined();
+    expect(plotConfig.loading).toBeDefined();
```
kafkasl: why not just check the actual value instead of `toBeDefined` ?

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
```diff
@@ -0,0 +1,108 @@
...
+    expect(plotConfig.error).toBeDefined();
+    expect(plotConfig.plotType).toBe(plotType);
+
+    expect(plotConfig.plotData).toBeDefined();
+    expect(_.isEqual(plotConfig.plotData, mockData.plotData)).toBe(true);
+
+    expect(plotConfig).toMatchSnapshot();
```
kafkasl, line 52: if you are matching the snapshot here, why add all those checks before this line which are essentially doing the same?

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
```diff
@@ -0,0 +1,108 @@
...
+
+    expect(plotConfig.config).toBeDefined();
+    expect(plotConfig.loading).toBeDefined();
+    expect(plotConfig.error).toBeDefined();
+    expect(plotConfig.plotType).toBe(plotType);
+
+    expect(plotConfig.plotData).toBeDefined();
+    expect(plotConfig.plotData.length).toEqual(0);
+
+    expect(plotConfig).toMatchSnapshot();
```
kafkasl, line 67: same here, either check the snapshot or individual fields, I don't see why would you do both

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
aerlaut: This code is removed, checking of the properties is done with `toMatchSnapshot()`

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
aerlaut: Use this instead of manually checking the properties, which is not robust

src/__test__/redux/actions/componentConfig/loadPlotConfig.test.js
aerlaut: This code is removed, checking of the properties is done with `toMatchSnapshot()`

src/__test__/redux/reducers/componentConfig/loadPlotData.test.js
```diff
@@ -0,0 +1,76 @@
...
+  )),
+}));
+
+const experimentId = fake.EXPERIMENT_ID;
+let testStore = null;
+
+const plotUuid = 'DotPlotMain';
+const plotType = plotTypes.DOT_PLOT;
+
+describe('plotDataLoaded', () => {
```
aerlaut: `loadPlotData` is changed to `fetchPlotDataWork`. Change the test name accordingly

src/redux/actions/componentConfig/fetchPlotDataWork.js
```diff
@@ -0,0 +1,46 @@
+import { initialPlotConfigStates } from 'redux/reducers/componentConfig/initialState';
+import getTimeoutForWorkerTask from 'utils/getTimeoutForWorkerTask';
+import { PLOT_DATA_LOADED, PLOT_DATA_LOADING, PLOT_DATA_ERROR } from 'redux/actionTypes/componentConfig';
+
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from 'utils/endUserMessages';
+import generatePlotWorkBody from 'utils/work/generatePlotWorkBody';
+import { fetchWork } from 'utils/work/fetchWork';
+
+const fetchWorkerPlotData = (experimentId, plotUuid, plotType) => async (dispatch, getState) => {
```
aerlaut: Change the name to be the same as the file name

src/components/plots/DotPlot.jsx
```diff
@@ -39,7 +20,7 @@ const DotPlot = (props) => {
     export: true,
     source: false,
     compiled: false,
-    editor: false,
+    editor: true,
```
aerlaut: change editor to false

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
-  const cellSets = useSelector((state) => state.cellSets);
+
+  const cellSets = useSelector(getCellSets());
   const {
     loading: cellSetsLoading,
     error: cellSetsError,
-    hierarchy,
+    hierarchy: cellSetHierarcy,
+    properties: cellSetProperties,
   } = cellSets;
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:111]: Use selector directly on line 111

src/redux/reducers/componentConfig/plotDataLoading.js
```diff
@@ -0,0 +1,16 @@
...
+  const { plotUuid } = action.payload;
+  return {
+    ...state,
+    [plotUuid]: {
+      ...initialPlotDataState,
```
aerlaut: Check if we still need this

src/utils/work/fetchWork.js
```diff
@@ -119,7 +119,7 @@ const fetchWork = async (
   }
 
   // If caching is disabled, we add an additional randomized key to the hash so we never reuse
-  // past results.
+  // past results. Use the string 'true' because localStorage only store values as strings
```
aerlaut: Moved to another PR

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:111]: changed

src/components/plots/DotPlot.jsx
aerlaut: changed

src/__test__/redux/reducers/componentConfig/loadPlotData.test.js
aerlaut: changed

src/redux/actions/componentConfig/fetchPlotDataWork.js
aerlaut: changed

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
+
+    // filterBy has the shape louvain/louvain-1
+    const [filterRootNode, filterKey] = filterCluster.split('/');
+
+    // If 'All" is chosen for the dropdown, there will always be representation from more than 1 group
+    if (filterRootNode === 'All') return true;
+
+    const filterClusterCellIds = Array.from(cellSetProperties[filterKey].cellIds);
+
+    const baseClusterParent = cellSetHierarcy.find((cellSet) => cellSet.key === baseCluster);
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: Check if you can use the cellsets selectors here I think we can get hierarchy by key.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
+    return baseClusterCellIds.some((baseCellIds) => {
+      if (filterClusterCellIds.some((filterCellId) => baseCellIds.has(filterCellId))) {
+        numIntersections += 1;
+      }
+
+      return numIntersections === 2;
+    });
+  };
+
+  const hasGroups = useMemo(
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Isnt it better to use `useCallback` here?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
-        },
-        geneNamesFilter: null,
-        sorter: { field: PROPERTIES[0], columnKey: PROPERTIES[0], order: 'descend' },
-      };
+    if (cellSetsLoading) return;
 
-      dispatch(loadPaginatedGeneProperties(experimentId, PROPERTIES, plotUuid, tableState));
+    // Marker genes calculation needs that the cellIds in groupBy (refer to fn definition)
+    // be represented by more than one groups in filterBy to enable comparison
+    // if (config?.useMarkerGenes && !hasGroupsToCompare(config.selectedCellSet, config.selectedPoints)) {
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: remove this

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
+      if (filterClusterCellIds.some((filterCellId) => baseCellIds.has(filterCellId))) {
+        numIntersections += 1;
+      }
+
+      return numIntersections === 2;
+    });
+  };
+
+  const hasGroups = useMemo(
+    () => hasGroupsToCompare(config?.selectedCellSet, config?.selectedPoints),
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Initially when the config is not defined we still do this calculation in the function, maybe its better to put a check in `hasGroupsToCompare` if config is defined?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: No, because the return value is a boolean, not a function. `useCallback` memoizes a function.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
+      },
+      geneNamesFilter: null,
+      sorter: { field: PROPERTIES[0], columnKey: PROPERTIES[0], order: 'descend' },
+    };
+
+    dispatch(loadPaginatedGeneProperties(experimentId, PROPERTIES, plotUuid, tableState));
+  };
 
   useEffect(() => {
+    if (config?.selectedGenes.length === 0 && !highestDispersionGenes?.length > 0 && !genesFetching) {
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:198]: What is this expression - `!highestDispersionGenes?.length > 0`? maybe you mean `!highestDispersionGenes.length`

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: The comment? I think it's good information so that people knows why we need to check for this,

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: There is already a check inside `hasGroupsToCompare` to check for the existence of the inputs. I'll move it out into the `useMemo` function body

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -71,75 +89,249 @@ const DotPlotPage = (props) => {
...
+    setMoreThanTwoGroups(true);
+
+    const currentComparedConfig = getComparedConfig(config);
+    if (config && !_.isEqual(previousComparedConfig.current, currentComparedConfig)) {
+      previousComparedConfig.current = currentComparedConfig;
+      dispatch(fetchPlotDataWork(experimentId, plotUuid, plotType));
+    }
+  }, [config, cellSetProperties]);
+
+  const loadInitialCustomGenes = () => {
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:184]: Other plots use this function such as the violin plot and embedding continuous, can you check if you can move this function and reuse it in those places?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:198]: We want this check to pass if `highestDispersionGene` is not loaded.

`!highestDispersionGenes?.length > 0` is a combination of 2 checks : 
-if `highestDispersionGenes` is `undefined`, then it evaluates to `!undefined` which is true
-if `heighestDispersionGenes` is defined and the data is still loading (`length` is 0), then it evaluates to `!([].length > 0)` will evaluate to `true`

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:184]: I don't want to do that refactoring in this PR because it's too big already. I can make another ticket for it. Sounds good?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Oh the commented line of code. Good catch! removed.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:198]: yeah but !highestDispersionGenes?.length will give the same output in those cases, its just a bit confusing when you see it at first maybe its me

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: Whoops, you've just uncovered a bug in React! I tried using a selector in the function (which is called inside a `useMemo`). Looking at the error, I don' think we can do this yet.

image

Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: no its not good to use hook inside a hook in general i think its one of the react hook rules, maybe you can do it on the top of the file as a variable.

Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: Also alternatively you can call the actual selector directly not through useSelector i guess?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:198]: Yup, which is what we want. I just checked `loadPaginatedGeneProperties`. If `highestDispersionGenes` is undefined (first call), it is the call to initialize the state. On error, it populates `data` to an empty array. So we also want to be able to re-dispatch this call if there is a previous error (`data` is empty).

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: I don't think we can do that, because `getCellSetsSelector` requires `state`, which is passed by `useSelector`.

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: I can move that line outside the function, but it makes it messy because we get the `baseClusterParent` as another global dependency from outside of the function and we lose the use of `baseCluster` as one of the keys of memoization.

Using the selectors is a good thing - the result is memoized. The result of the `hasGroups...` function is also memoized and the cellSetsHierarchy array is short, so I don't think the performance hit of not using the memoized selector would be a noticeable hit.

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: This works,  but as I said, we lose the functionality of `baseCluster` as a key for useMemo, and we can't use `baseClusterParent` as a key as it is an object (whose reference changes everytime)

image

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: So I'd still use the current approach. wdyt?

src/redux/reducers/componentConfig/plotDataLoading.js
aerlaut: not needed anymore as it is initially initialized when loading the plot config, so this is removed.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: You can pass the hierarchy and keys directly and  call `getCellSetsHierarchyByKeys`? As for the other things im not sure if it doesnt impact performance then maybe its good?

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:142]: Yup, that's what I did, as I show in the example above. However, I didn't really like it because : 
- It splits the code so that `baseClusterParent` away from where it's used, and causes `hasGroupsToCompare` to have another dependency on a global variable. 
- The data that is fetched by `getCellSetsHierarchyByKeys` already exists in `cellSets`, which is used in other parts of the code.
- Performance-wise, it's not that much of a difference. So I think the current code (without the selector) looks better as it has better legibility.

So I think the way it's currently coded looks fine.

StefanBabukov (approved):

PR #550, opened 2021-11-09
[BIOMAGE-1526] - Add option in Volcano Plot to automatically calculate axes

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1526

https://ui-agi-volcano.scp-staging.biomage.net/

N.A.

- The current volcano plot filters out data that does not fall between the axes boundaries. Data should not be filtered out, so this filtering is removed in this PR.
- The current volcano plot does not have an intuitive way to set the axes to default. New checkboxes are added to provide a way for users to use auto-adjusted axes. Users are still able to set the axes manually.
- We still want to have the ability to specify the range of the axes that is shown. This would enable side-by-side comparison between multiple volcano plots with the same axes ranges.

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [x] Unit tests written
- [x] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/utils/plotSpecs/generateVolcanoSpec.js
```diff
@@ -1,28 +1,8 @@
...
-  });
-
-  const logFoldChangeFilterExpr = (config.logFoldChangeDomain)
-    ? `datum.logFC > ${config.logFoldChangeDomain * -1} && datum.logFC < ${config.logFoldChangeDomain}`
-    : 'true';
-
-  const negativeLogpValueFilterExpr = (config.maxNegativeLogpValueDomain)
-    ? `datum.neglogpvalue < ${config.maxNegativeLogpValueDomain}`
-    : 'true';
-
```
aerlaut, line 25: This is removed because we do not want to filter out user's data that lies out of the plot area

src/__test__/components/plots-and-tables/Volcano.test.jsx
```diff
@@ -1,128 +0,0 @@
-import React from 'react';
```
aerlaut, line 1: Moved to `volcano/index.test.jsx`

src/components/plots/styling/volcano/DimensionsRangeEditor.jsx
```diff
@@ -21,29 +21,48 @@ const VolcanoDimensionsRangeEditor = (props) => {
...
+        label='Y-Axis Range'
       >
+        <Checkbox
+          onChange={() => {
+            onUpdate({ yAxisAuto: !config.yAxisAuto });
+          }}
+          defaultChecked
+          checked={config.yAxisAuto}
+        >
+          Auto
```
StefanBabukov, line 37: instead of using auto why dont we just keep the dimensions accurate according to the axis range when you change it? This way we won't need auto?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -203,12 +218,11 @@ const VolcanoPlot = ({ experimentId }) => {
...
 
-    if (plotData.length === 0 || loading || _.isEmpty(spec.spec)) {
+    if (plotData.length === 0 || diffExprLoading) {
       return <Loader experimentId={experimentId} />;
     }
-    return (
-      <Vega data={{ data: plotData }} spec={spec.spec} renderer='canvas' />
-    );
+
+    return <Vega spec={generateSpec(config, plotData)} renderer='canvas' />;
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx]: Is it good to call `generateSpec` on every re-render? Does it slow things down more?

src/components/plots/styling/volcano/DimensionsRangeEditor.jsx
aerlaut: I don't quite understand. Can you please explain again?

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx]: Good point. I've wrapped the call to `generateSpec` around a use effect and a check to see if the config is still the same. It initially rendered the plot 3 times. After that, it renders the plot only once.

src/components/plots/styling/PointDesign.jsx
```diff
@@ -31,19 +31,22 @@ const PointDesign = (props) => {
...
-            min={1}
-            max={10}
-            onChange={(value) => {
-              handleChange({ marker: { opacity: value } });
-            }}
-            marks={{ 1: 1, 10: 10 }}
-          />
-        </Form.Item>
+
+        {config.marker.showOpacity ? (
```
StefanBabukov, line 35: Why don't you just add a check in `PlotStyling` for if config is defined? This way we don't need to check for the config in any of the plot controls which all rely on config.

StefanBabukov: And if config is undefined, show loader on the controls, what do you think?

vickymorrison (approved):

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
```diff
@@ -155,10 +179,10 @@ const VolcanoPlot = ({ experimentId }) => {
 
     const date = moment.utc().format('YYYY-MM-DD-HH-mm-ss');
     const fileName = `de_${experimentId}_${cellSet}_vs_${compareWith}_${date}.csv`;
-    const disabled = plotData.length === 0 || loading || _.isEmpty(spec.spec) || error;
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:158]: so if the spec is empty initially while loading everything, will the export button be disabled?

src/components/plots/styling/PointDesign.jsx
aerlaut: This check selectively enable / disable the slider for point opacity.  if `showOpacity` is true, it shows the slider. This is the case for categorical embedding and continuous embedding

Feedback from Vicky, we want this to be disabled for volcano plot, so this control is hidden by setting `showOpacity` to `false` for volcano plot.

src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/volcano/index.jsx:158]: The CSV link does not depend on the spec, so I remove it. That allows the users to download the data without haveing to load for the spec to load. `loading` and `error` is renamed to `diffExprLoading` and `diffExprError` respectively.

StefanBabukov (approved):

PR #559, opened 2021-11-18
[BIOMAGE-1535] Stop the experiments pollution in the UI

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/jira/software/c/projects/BIOMAGE/boards/1?modal=detail&selectedIssue=BIOMAGE-1535
https://ui-experiment-pollution.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -11,7 +11,7 @@ const loadMarkerGenes = (
   // Disabled linter because we are using == to check for both null and undefined values
   // eslint-disable-next-line eqeqeq
   if (experimentId == null || resolution == null) throw new Error('Null or undefined parameter/s for loadMarkerGenes');
-
+  console.log('loading marker genes ', experimentId, resolution, plotUuid, numGenes, selectedCellSet);
```
StefanBabukov: remove this

src/redux/reducers/index.js
```diff
@@ -31,3 +32,17 @@ export default combineReducers({
...
   networkResources: networkResourcesReducer,
 });
+
+const rootReducer = (state, action) => {
+  if (action.type === EXPERIMENTS_SWITCH) {
+    const newState = {
+      networkResources: state.networkResources,
+      experimentSettings: state.experimentSettings,
+    };
+    console.log('lmao HEREXDD', newState);
```
StefanBabukov: remove this

src/redux/store.js
```diff
@@ -33,7 +33,7 @@ const makeStore = () => {
       store.replaceReducer(createNextReducer(rootReducer));
     });
   }
-
+  console.log('CREATING STORE ', store);
```
StefanBabukov: remove this

src/components/data-management/LaunchAnalysisButton.jsx
```diff
@@ -56,6 +56,7 @@ const LaunchAnalysisButton = () => {
     const lastViewed = moment().toISOString();
     dispatch(updateExperiment(experimentId, { lastViewed }));
     dispatch(updateProject(activeProjectUuid, { lastAnalyzed: lastViewed }));
+    dispatch(switchExperiment());
```
ivababukova, line 59: we were talking that you will call this from the content wrapper. Why only here?

StefanBabukov: As discussed with Agi yesterday, because if you open the same experiment again in the browser it will remove everything. This way we make sure we clean the store only when launch analysis is pressed.

aerlaut: Experiment pollution could only happen when going back from a page with experiment, to data management, and then choosing another experiment/project. So we should reset the store when moving from a page with out experiment id to a page with experiment id. That's why it's set at the launch button.

Changing the experimentId directly from the URL should cause a hard reset, which will cause the redux store to be recreated. A hard reset will cause the redux store to reset right?

Am I understanding correctly or did I miss something?

src/__test__/redux/actions/experiments/switchExperiment.test.js
```diff
@@ -0,0 +1,21 @@
...
+  it('switches the experiment to its initial values', () => {
+    const store = createStore(rootReducer, storeState, applyMiddleware(thunk));
+    store.dispatch(switchExperiment());
+    expect(store.getState().backendStatus).toEqual(storeState.backendStatus);
+    expect(store.getState().experimentSettings).toEqual(storeState.experimentSettings);
+    expect(store.getState().samples).toEqual(storeState.samples);
+    expect(store.getState().projects).toEqual(storeState.projects);
+    expect(store.getState().cellSets).toEqual(cellSetsInitialState);
+    expect(store.getState().experiments).toEqual(experimentsInitialState);
+  });
```
aerlaut: Hmm... This test doesn't look like it does anything. You haven't made any modifications to the store that would allow us whether the store is really reset or not.

Some testing idea : 
Setup the store, setup the mockAPI endpoints, load cellsets, check cell set exists, reset the store, check if cellsets is empty (which should be if we reset the store right?)

And instead of checking the properties one by one, I think you can use `matchSnapshot`.

aerlaut: Can we use the real store (via `makeStore` instead of bringing `createStore`?) that would greatly involve other codes in our platform too.

src/components/data-management/LaunchAnalysisButton.jsx
StefanBabukov: Yes hard reset would cause it correct

src/__test__/redux/actions/experiments/switchExperiment.test.js
StefanBabukov: Why should we load everything like we are  testing a component. This is a simple action to reset some things to their initial state which uses the actual file thats used and keep the value of the selected store parts as is? Initially i was going to do that  dispatch loading and everything, but it just seemed unnecessarily complex when this small functionality can be tested in a simple way. What do you think? Also I created a file which has all the values in the store so we can check by that.

aerlaut: Ooh I see. I didn't understand that `storeState` actually contains a store that's loaded with experiments, projects, etc. Can you rename `storeState` to something more descriptive (e.g. `storeWithExperimentData`)?

And yea, I think instead of checking that the properties equal to their initial state one by one, you can use `matchSnapshot` because there are more properties which you have not checked for (e.g. embeddings, genes, cellMeta, cellInfo, etc.)

aerlaut (approved):
Looks good and tested. Just need to update the tests,

PR #573, opened 2021-12-01
[BIOMAGE-1583] - Fix source file download

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1583

https://ui-agi-s3-download.scp-staging.biomage.net/

UI : https://github.com/biomage-ltd/ui/pull/573
API : https://github.com/biomage-ltd/api/pull/266

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [x] Unit tests written
- [x] Deployed to staging
- [x] Passed integration tests 

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/redux/actions/samples/updateSampleFile.js
```diff
@@ -32,6 +32,10 @@ const updateSampleFile = (
       };
 
       const newSample = mergeObjectWithArrays(sample, diffObject);
+
+      // Do not save the fileObject to DynamoDB
+      delete newSample.files[fileName].fileObject;
```
cosa65: shouldn't this delete be done in `saveSamples`? Since that's where it is being saved to dynamo, because if `saveSamples` is run outside of this `updateSampleFIle` we stil have the problem, right?

src/components/data-management/UploadDetailsModal.jsx
```diff
@@ -39,7 +36,7 @@ const UploadDetailsModal = (props) => {
 
   useEffect(() => {
     if (replacementFileBundle) {
-      bundleToFile(replacementFileBundle, SELECTED_TECH).then((newFile) => {
+      fileObjectToFileRecord(replacementFileBundle, SELECTED_TECH).then((newFile) => {
```
cosa65: You have the "File bundle" syntax you are changing in other places too, like `uploadFileBundle`

cosa65: If part of what you are doing in this ticket is renaming file bundle to file object I would change it everywhere

src/utils/data-management/downloadSingleFile.js
```diff
@@ -1,14 +1,9 @@
-import { saveAs } from 'file-saver';
-import { Storage } from 'aws-amplify';
+import downloadFromUrl from 'utils/data-management/downloadFromUrl';
+import { getFromApiExpectOK } from 'utils/getDataExpectOK';
 
-const downloadSingleFile = async (activeProjectUuid, sampleUuid, fileName, bundleName) => {
```
cosa65, line 4: We were using amplify? And it was failing because the permissions to access this from amplify had already been removed?

aerlaut: Yup

src/components/data-management/UploadDetailsModal.jsx
aerlaut: replaced

src/redux/actions/samples/updateSampleFile.js
aerlaut: Hmm... right, that's true. I'll move this delete inside `saveSamples`.

aerlaut: This call is moved inside `saveSamples`. Please refer to the new changes by looking at the file from the "Files changed" tab.

src/components/EditableParagraph.jsx
```diff
@@ -10,6 +10,7 @@ const EditablePagrapraph = (props) => {
 
   const paragraphEditor = useRef();
 
+  const [text, setText] = useState(value);
```
cosa65, line 13: What are the changes done here for? They don't seem related to the ticket

cosa65, line 13: It seems the `text` `useState` is being used as a local version of `value`, so why not call it `localValue` or something like that to make that clearer? It is better to use one cohesive name.

If you feel `value` is not good I would change `value` to `text` and `text` to `localText`,

src/utils/upload/processUpload.js
```diff
@@ -217,16 +216,16 @@ const processUpload = async (filesList, sampleType, samples, activeProjectUuid,
   });
 };
 
-const bundleToFile = async (bundle, technology) => {
+const fileObjectToFileRecord = async (fileObject, technology) => {
```
cosa65, line 234: `bundleToFile` wasn't very self explanatory, but `fileObjectToFileRecord` isn't very explanatory either, what do you think of perhaps coming up with a better name or adding comments explaining what is the input and what is the output?

Example of comments.

  ```
/**
   * Does cool things.
   * @param {string} fileObject blalbalba.
   * @param {string} technology blsblslfs.
   * @returns {object} fileRecord blablabladblsal
   */
   
```

File: src/components/EditableParagraph.jsx (hunk 4 above)
Comment (cosa65) [reply src/components/EditableParagraph.jsx:13]: Ah you added unit tests so codecov doesn't cry I think, but the refactoring here was to make this respond better or something?

Comment (aerlaut) [reply src/components/EditableParagraph.jsx:13]: So that it can have internal value, and the component can be independently without outside values

Comment (aerlaut) [reply src/components/EditableParagraph.jsx:13]: I think this is already pretty clear. We usually pass in `value` to an input component. However, inside the component I used `text` because I'm not really using a real `input` component, but a div with `contenteditable`, which should display some kind of `text` inside it.

File: src/utils/upload/processUpload.js (hunk 5 above)
Comment (aerlaut) [reply src/utils/upload/processUpload.js:234]: I've added a comment to explain the function,

File: src/components/EditableParagraph.jsx (hunk 4 above)
Comment (cosa65) [reply src/components/EditableParagraph.jsx:13]: But `text` is the same as `value` but internal, so why use a completely different name?

I think it is clear right now because there isn't much code, the problem is more noticeable in larger files but using cohesive names is still a good practice.

Other naming ideas could be `inputText` `text` or `inputValue` and `value`, you can rename those on destructuring  the `props` so that from outside we only see `value` if you want.

Comment (aerlaut) [reply src/components/EditableParagraph.jsx:13]: Hmm... ok, that's a good argument. I'll change the name,

File: src/components/data-management/UploadDetailsModal.jsx [hunk 6]
```diff
@@ -25,30 +25,27 @@ const UploadDetailsModal = (props) => {
   } = props;
   const { fileCategory, sampleUuid } = uploadDetailsModalDataRef.current || false;
   const file = uploadDetailsModalDataRef.current?.file || {};
-  const {
-    upload = {}, bundle = {},
-  } = file;
+  const { upload } = file || {};
```
cosa65: Use `??` instead for null checks

cosa65: also perhaps do this same change to 
```
const { fileCategory, sampleUuid } = uploadDetailsModalDataRef.current || false;
  const file = uploadDetailsModalDataRef.current?.file || {};
```

cosa65: Also if you can change the usage of false to null instead in `uploadDetailsModalDataRef.current || false`

src/redux/actions/samples/saveSamples.js
```diff
@@ -15,21 +16,29 @@ const saveSamples = (
...
   // add new sample to payload
   if (addSample) {
     const projectSamples = projects[projectUuid].samples.reduce((acc, sampleId) => {
       acc[sampleId] = samples[sampleId];
       return acc;
     }, {});
 
+    // Do not save the fileObject to DynamoDB as it can not be serialized
+    Object.keys(newSampleCopy.files).forEach((file) => {
+      delete newSampleCopy.files[file].fileObject;
```
cosa65: We decided to disable retry button when the file is fetched from dynamo (because then we don't have fileObject to perform the retry).

src/redux/actions/samples/saveSamples.js
```diff
@@ -15,21 +16,29 @@ const saveSamples = (
...
+    });
+
     payload = projectSamples;
     payload = {
       ...payload,
-      [newSample.uuid]: newSample,
+      [newSample.uuid]: newSampleCopy,
```
cosa65: rename to `newSampleToUpload`

src/utils/upload/processUpload.js
```diff
@@ -66,13 +66,20 @@ const compressAndUploadSingleFile = async (
...
+  };
+
   try {
     loadedFile = await loadAndCompressIfNecessary(file, () => (
       dispatch(
         updateSampleFile(
           sampleUuid,
           fileName,
-          { upload: { status: UploadStatus.COMPRESSING } },
+          {
```
cosa65: super nitpick: compress again this and other line

src/components/data-management/UploadDetailsModal.jsx
aerlaut: changed

src/redux/actions/samples/saveSamples.js
aerlaut: changed

src/utils/upload/processUpload.js
aerlaut: changed

src/redux/actions/samples/saveSamples.js
aerlaut: added

cosa65 (approved):

vickymorrison (approved):

PR #584, opened 2021-12-14
[BIOMAGE-1252] Add advanced filtering button and modal

StefanBabukov:
# Background
#### Link to issue 
https://biomage.atlassian.net/jira/software/c/projects/BIOMAGE/boards/1?modal=detail&selectedIssue=BIOMAGE-1252
https://ui-advanced-filtering-ui584.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+        ))}
+      </Menu>
+    );
+  };
+
+  return (
+    <Modal
+      visible
+      title='Advanced filters'
+      onCancel={() => onCancel()}
```
cosa65: I think you could do `onCancel={onCancel}`

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+        criteria: 'logfc',
+        condition: 'lt',
+        value: 0,
+      },
+      Significant: {
+        criteria: 'pValue',
+        condition: 'lt',
+        value: 0.05,
+      },
+    };
```
cosa65: `presetFilters` can be outside the `renderPresetFilters` function, at the same level as `conditionOptions`

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+  return (
+    <Modal
+      visible
+      title='Advanced filters'
+      onCancel={() => onCancel()}
+    >
+      <Form>
+
+        <Form.List
+          name='filterForm'
```
cosa65, line 80: What is this attribute for?

src/__test__/components/data-exploration/differential-expression-tool/DiffExprResults.test.jsx
```diff
@@ -336,4 +336,20 @@ describe('DiffExprResults', () => {
     // Expect table to contain Empty component
     expect(empty.length).toEqual(1);
   });
+
+  it('Advanced filter button opens and closes the modal', () => {
```
cosa65, line 340: We could also have a test for checking that the advanced filter button doesn't show up when diff expr results aren't being shown

cosa65 (commented):
Nice! Some things I think we could do to improve it:
	- Could you change the "OK" button for "Apply filters" as in the mockups?
	- Increase the width a little bit so that the placeholder text isn't cut off.

Also tagging @vickymorrison in case she wants to take a look.

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
StefanBabukov: I use it to access the form items, used to calculate the value for min and max for the `InputNumber`

src/__test__/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.test.jsx
```diff
@@ -0,0 +1,55 @@
...
+import '@testing-library/jest-dom';
+
+import {
+  render, screen, fireEvent, waitFor,
+} from '@testing-library/react';
+import AdvancedFilteringModal from 'components/data-exploration/differential-expression-tool/AdvancedFilteringModal';
+
+describe('Advanced filtering modal', () => {
+  const renderAdvancedFilteringModal = () => {
+    const onCancel = jest.fn();
```
cosa65: You could add a test for `onCancel` callback being triggered too

src/__test__/components/data-exploration/differential-expression-tool/DiffExprResults.test.jsx
StefanBabukov: The advanced filter button is only rendered in DiffExprResults so if `DiffExprResults` is not there, the button is not rendered too

cosa65: ahh you're right, ok

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,162 @@
...
+    >
+      {Object.keys(presetFilters).map((filter) => (
+        <Menu.Item key={filter}>
+          {filter}
+        </Menu.Item>
+      ))}
+    </Menu>
+  );
+
+  const changeCriteria = (criteria, index = false) => {
```
cosa65: It better to use `null` for elements that aren't set, javascript allows us to do this but it is usually cleaner to avoid having many different possible types in a single variable.

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,162 @@
...
+                      />
+                    </Form.Item>
+
+                    <Form.Item
+                      name={[field.name, 'value']}
+                      fieldKey={[field.value, 'condition']}
+                    >
+                      <InputNumber
+                        style={{ width: 140 }}
+                        min={valuesDomain[index] ? valuesDomain[index][0] : 0}
```
cosa65: couldn't you just get the min from `valueRestrictions[field.key]` instead?

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
```diff
@@ -0,0 +1,162 @@
...
+                      />
+                    </Form.Item>
+                  </Space>
+
+                  <CloseOutlined onClick={() => remove(field.name)} />
+                </Space>
+              ))}
+
+              <Space direction='horizontal'>
+                <Button onClick={() => add()} block icon={<PlusOutlined />}>
```
cosa65: this could be `onClick={add}`

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
cosa65: would this allow you to get rid of `valuesDomain`? Or are you using it for something different?

StefanBabukov: When the form is empty fields is empty and this will crash and we want the max and min values to change when you switch between different criteria

vickymorrison:
1.The "Apply Filters" button should be disabled when there are no filters added.
2. I suggest a small change to the text of the 2 'add filter' buttons to improve UX (as per the 'don't make me think' ethos): I think "add custom filter" and "add preset filter" makes it much clearer to the user.
Screenshot 2021-12-16 at 12 40 25
3. Until the functionality of the advanced filtering is implemented, I suggest showing the "Apply filters" button as disabled with a tooltip stating "Coming soon!" This way the current ticket can be merged and released to production whilst maintaining a good experience for our current users. Though you may already have a solution for this.

src/components/data-exploration/differential-expression-tool/AdvancedFilteringModal.jsx
cosa65: I didn't understand the empty fields causing a crash thing.

On the `max` and `min` I understand that they have to change, but I don't see why accessing the criteria directly will not work, right now you are using `form.setFieldsValue({});` to trigger a re render, right? So it would still re render.

We can have a call tomorrow if you want to talk about this, perhaps I am missing something.

cosa65: This comment might be overkill, so feel free to ignore if you prefer not to do this, but it is a suggestion to perhaps avoid doing some re renders.

If you want to think of an alternative to having to re render to update the `min` and `max` we could just not use `min` and `max` and make sure they are within range with the `onChange` of `InputNumber` and setting the values manually.
 
If we went this way we could maybe even implement a wrapper component for `InputNumber` that allows us to have `min` and `max` as functions and not have to do everything here (and make it reusable).

StefanBabukov: The re-render occurs when you change something so it is not slowing down. I'll leave it like this for now, because I don't think the change will make a noticable difference in experience.

vickymorrison:
Please be consistent with the buttons - have a '+' on both buttons.
Screenshot 2022-01-04 at 13 01 26

Otherwise, everything looks good.

vickymorrison (approved):

cosa65 (commented):
Nice! looks very good and close to finish probably, just a couple of things:

- Could you set custom steps (how much the value is increased/decreased when clicking the up/down arrows) for each of the filter labels? (like for example when the range is [0,1] the arrows could have a step that is smaller than 1).
- there is a warning in the console on render, could you see how to make it go away? 
image

-

cosa65 (approved):

PR #591, opened 2021-12-17
[BIOMAGE-1623] Fix metadata tracks infinite loop

cosa65:
# Background
#### Link to issue 
https://biomage.atlassian.net/browse/BIOMAGE-1623

https://ui-martinfosco-ui591.scp-staging.biomage.net/

# Changes
### Code changes

# Definition of DONE

### Code updates

### Testing
- [ ] Unit tests written
- [ ] Deployed to staging

### Documentation updates

### Approvers

### Just before merging:

### Optional


src/components/ReorderableList.jsx
```diff
@@ -54,7 +54,10 @@ const ReorderableList = (props) => {
     <Button
       size='small'
       shape='circle'
-      disabled={currentPosition === 0}
+      disabled={
+        reorderableList[currentPosition].disabledReorder
```
cosa65, line 58: To make the ordering less messy the metadata tracks now work in this way: If disabled, a track option gets moved to the bottom along with any other metadata tracks and it can't be reordered, (it also makes more sense with the redux data we have).

To this end, I changed the `ReorderableList` so that it can receive a property that tells it to disable reordering for one item in particular (reordering's enabled by default so it doesn't affect other usage of this).

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useEffect, useState, useRef } from 'react';
```
cosa65, line 1: I rewrote most of this component because it was a little bit hard to follow.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -248,7 +248,7 @@ const HeatmapPlot = (props) => {
 
     const [cellIndexStr, trackIndex, mouseX, mouseY] = info;
 
-    const trackOrder = Array.from(heatmapSettings.selectedTracks).reverse();
```
cosa65, line 251: Removed `Array.from`, `selectedTracks` already is an array.

cosa65: Also removed `Array.from`, `selectedTracks`  already are arrays

cosa65:
Still need to write/fix tests

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -14,95 +14,95 @@ import { updatePlotConfig } from '../../../redux/actions/componentConfig';
 import ReorderableList from '../../ReorderableList';
 import { getCellSetsHierarchy } from '../../../redux/selectors';
```
ivababukova, line 15: while you are at it, you can also fix those imports and remove the `../`

ivababukova:
I noticed that this warning appears on the console after I added custom cell sets track on the heatmap, where custom cell sets is empty (I haven't created any new clusters), and on hover over that track with the mouse. Can you fix it:
Screenshot 2021-12-22 at 08 57 30

ivababukova:
I saw that this warning appears on the console the first time I hover over the heatmap. Can you look into it?
Screenshot 2021-12-22 at 09 00 15

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -14,95 +14,95 @@ import { updatePlotConfig } from '../../../redux/actions/componentConfig';
...
     (state) => state.componentConfig[componentType].config.selectedTracks,
   );
 
-  const getTrackData = () => selectedHierarchy.map(
-    (data) => ({ selected: selectedTracks.includes(data.key), key: data.key }),
-  );
+  const [listData, setListData] = useState([]);
 
-  const isInitialRenderRef = useRef(true);
+  useEffect(() => {
```
ivababukova, line 40: I noticed that this `useEffect` gets triggered twice every time `componentConfig/update` action happens (like when I move a group up with the arrows: changing `selectedTracks` is effectively a change of `componentConfig`), which doesn't seem right. Shouldn't the useEffect get triggered only once?

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -14,95 +14,95 @@ import { updatePlotConfig } from '../../../redux/actions/componentConfig';
...
     dispatch(
       updatePlotConfig(componentType, {
-        selectedTracks: getEnabledTracks(),
+        selectedTracks: reorderedSelectedTrackKeys,
       }),
     );
-  }, [trackData]);
+  };
 
   useEffect(() => {
```
ivababukova: looks like you need to remove this entire empty `useEffect`

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -14,95 +14,95 @@ import { updatePlotConfig } from '../../../redux/actions/componentConfig';
...
-  );
+    const unselectedTracks = _.difference(hierarchy.map(({ key }) => key), selectedTracks);
+    const unselectedTracksListData = convertToReorderableListData(
+      unselectedTracks, false, hierarchy,
+    );
 
-  useEffect(() => {
-    // Prevent initial dispatch when object appears
-    if (isInitialRenderRef.current) return;
+    setListData([...selectedTracksListData, ...unselectedTracksListData]);
```
ivababukova: I see what you're doing here, but I wonder if there is a way to update only the element that changed, rather than having to recreate the `selectedTracks` from scratch every time. For example, if I enable a track, instead of re-creating the entire `selectedTracks`, can we not just go and change `disabledReorder` for that track to `true`?

It is also true that we won't have many `cellClassKeys` to loop over (we only have custom cell sets, samples and louvain), so there is little value into optimising this.

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
```diff
@@ -14,95 +14,95 @@ import { updatePlotConfig } from '../../../redux/actions/componentConfig';
...
+    } else {
+      _.pull(newSelectedTracks, key);
+    }
 
-  const getEnabledTracks = () => trackData.filter((entry) => entry.selected).map((o) => o.key);
+    dispatch(
+      updatePlotConfig(componentType, {
+        selectedTracks: newSelectedTracks,
+      }),
+    );
```
ivababukova: this triggers the `useEffect` in lines 39-52, is this intended?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -27,6 +27,8 @@ import HeatmapTracksCellInfo from 'components/data-exploration/heatmap/HeatmapTr
 import getCellClassProperties from 'utils/cellSets/getCellClassProperties';
 import useConditionalEffect from 'utils/customHooks/useConditionalEffect';
 
+import { reversed } from 'utils/arrayUtils';
```
ivababukova: I don't see where you use this `reversed`

src/components/plots/helpers/heatmap/populateHeatmapData.js
```diff
@@ -151,13 +150,13 @@ const populateHeatmapData = (
 
   if (!vitessce) {
     return generateVegaData(
-      cellOrder, geneOrder, trackOrder,
+      cellOrder, geneOrder, reversed(selectedTracks),
```
ivababukova, line 153: why reversing?

src/__test__/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.test.jsx
```diff
@@ -1,154 +1,130 @@
...
+      ['louvain clusters', 'Custom cell sets', 'Samples'],
+      [true, false, false],
+      items,
+    );
   });
 
-  it('order and shown tracks are changed when the switches and the buttons are pressed', () => {
-    const store = mockStore({
-      ...initialState,
+  it.only('Can enable and reorder correctly', async () => {
```
ivababukova: this `only` causes the other test to be skipped -- don't forget to remove it

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: In `populateHeatmapData`, when we call `generateVegaData`

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
cosa65: good point, I'll look into improving the rendering

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
cosa65: Resolving this one because it seems like the same topic as https://github.com/biomage-ltd/ui/pull/591#discussion_r773702048

cosa65:
> I saw that this warning appears on the console the first time I hover over the heatmap. Can you look into it? Screenshot 2021-12-22 at 09 00 15

Should be fixed now

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
cosa65: Hmm so the thing is that the way we are rendering this is by looking at an array: 
`selectedTracks` which just holds which tracks are enabled, so for example if we `louvain` and `samples` enabled then: `selectedTracks = ['louvain', 'sample'];`

So when a track is hidden (`sample` for example) we do: `selectedTracks = ['louvain']`. Everything with `disabledReorder` and all of that is generated using `selectedTracks`.

(To reorder we just reorder the `selectedTracks`).

I think if we wanted to change this to make order and selection not be dependent on each other we would need to change what is being stored in redux for something more complex, which I am not sure if it is worth it since the primary objective here is creating `selectedTracks` for `populateHeatmapData`.

ivababukova: I agree, and considering the fact that this is super small amount of data and there is very little impact on performance, it is not really worth it.

ivababukova: you can spend your time better by optimising other things that really do matter

src/components/data-exploration/heatmap/HeatmapMetadataTrackSettings.jsx
ivababukova: @cosa65 did you manage to figure this out?

cosa65: I've been looking into it and it seems that `HeatmapMetadataTrackSettings` is being rendered twice (as in there are two components that are `HeatmapMetadataTrackSettings`) and that's why things are done twice, I am not sure exactly how yet so I will take a look at the parent components of this

cosa65: It seems this is the reason for the issue: 
image

and they also say this which I don't understand too well:
image

after wrapping the items we show within <menu.item> components it doesn't render twice, however the menu.item changes some of the menu's behavior so I will have to look into that now

cosa65: In the end for the solution I implemented a `Menu.Item` wrapper that removes all custom styling and behavior from it so that the metadata tracks and group by settings components don't change anything (except that this time there's only one of each instead of duplicates).

ivababukova: Great find! Can you make sure that you show this to the rest of the team, particularly Agi and Stefan, to avoid them struggling with this in the future?

ivababukova (approved):

PR #606, opened 2022-01-14
[BIOMAGE-1648] - Correct data management navigation

aerlaut:
# Background
#### Link to issue 

https://biomage.atlassian.net/browse/BIOMAGE-1648

https://ui-agi-ui606.scp-staging.biomage.net/

N/A

- Issue was caused because navigating using the menu bars did not update `experimentInfo` when moving away from Data Management, unlike Launch Analysis. `experimentInfo` is used by Data Processing to loop over the samples and show data. 
- This happens because navigating using the left sidebar was using `navigateTo`, while launch analysis directly move to Data Processing using `router.push`.
- Solution is implemented by making Launch Analysis to use `navigateTo` to move to Data Processing, and updates to `experimentInfo` are moved to `navigateTo`.

# Changes
### Code changes

- Refactor `switchExperiment` to be fired inside `AppRouteProvider` instead of in `WrappedApp`.
- update experiment info inside of `switchExperiment`.

# Definition of DONE

### Code updates

### Testing
- [x] Unit tests written
- [x] Deployed to staging

To set up easy local testing with inframock, follow the instructions here: https://github.com/hms-dbmi-cellenics/inframock
To deploy to the staging environment, follow the instructions here: https://github.com/hms-dbmi-cellenics/biomage-utils

### Documentation updates

### Approvers

### Just before merging:
- [ ] After the PR is approved, the `unstage` script in here: https://github.com/hms-dbmi-cellenics/biomage-utils is executed. This script cleans up your deployment to staging

### Optional


src/utils/AppRouteProvider.jsx
```diff
@@ -10,9 +15,15 @@ const AppRouterContext = React.createContext(null);
 const AppRouteProvider = (props) => {
```
kafkasl, line 43: could you explain what this `AppRouterProvider` does?, Is it used for all the navigation inside cellenics? or only for some routes

src/utils/AppRouteProvider.jsx
```diff
@@ -10,9 +15,15 @@ const AppRouterContext = React.createContext(null);
...
   const { children } = props;
   const router = useRouter();
+  const dispatch = useDispatch();
 
   const [renderIntercept, setRenderIntercept] = useState(null);
 
+  const activeProjectUuid = useSelector((state) => state?.projects?.meta?.activeProjectUuid);
```
kafkasl: if this is only going to be used when `previousRoute.match('/data-management')` then it should be put inside the if. Here this will be run for every request.

src/utils/AppRouteProvider.jsx
```diff
@@ -37,6 +48,19 @@ const AppRouteProvider = (props) => {
       setRenderIntercept(availableIntercepts.DATA_PROCESSING(nextRoute, hardNavigate));
       return;
     }
+
+    if (previousRoute.match('/data-management')) {
+      const lastViewed = moment().toISOString();
+
+      dispatch(updateExperiment(experimentId, { lastViewed }));
+      dispatch(updateProject(activeProjectUuid, { lastAnalyzed: lastViewed }));
```
kafkasl: this is out of the scope of this PR, but these two lines do not make sense:
```
```
Having to make 2 extra requests with "modify permission" (PUT) to update `lastViewed` is completely unnecessary. This should be a side effect in the API endpoints. ie., whenever the endpoint corresponding to `updateExperimentInfo` is hit, then set those two variables from the API (possibly with a hook).

src/utils/AppRouteProvider.jsx
```diff
@@ -37,6 +48,19 @@ const AppRouteProvider = (props) => {
       setRenderIntercept(availableIntercepts.DATA_PROCESSING(nextRoute, hardNavigate));
       return;
     }
+
+    if (previousRoute.match('/data-management')) {
+      const lastViewed = moment().toISOString();
```
kafkasl: I would prefer you put all the new code inside a new function (with some comments of why it's needed) in this file so the `AppRouteProvider` remains easy to read. e.g.:
```
if (previousRoute.match('/data-management')) {
     updateExperiments();
     continueNavigation(nextRoute, hardNavigate);
}
```

kafkasl (commented):
I am concerned that although fixing the issue it will be easy to make that mistake again (i.e., forgetting to call the correct `routing`) because it seems easy to use `router.push` instead of `navigateTo`. My question is: should we only use `navigateTo` to change pages? (are we doing it that way?)

src/utils/AppRouteProvider.jsx
aerlaut: `AppRouteProvider` is a component which wraps the application and acts like a middleware for route changes. It intercepts navigation and allow checking of conditions before letting the navigation proceed. Yes, after this PR, all navigation inside Cellenics is done via `AppRouteProvider`.

src/utils/AppRouteProvider.jsx
aerlaut: We can't do this because hooks can only be called from the top level of a functional component, not inside a loop or condition (https://reactjs.org/docs/hooks-rules.html). Trying to move this into the function definition causes an error.

aerlaut: Right now, changes to `experiments`, `activeProjectUuid` or `experimentId` causes this element to re-render. This would happen only on the data-management page. In the other modules, users do not change experiments or projects, so this would be fine.

I checked the profiler, and yes change to these variables cause re-rendering of the app provider. However, changing these variables will not only impact this component, but also other components wrapped by the `provider` which displays the project and experiments visually. I checked in the `Profiler` and the impact of re-rendering of this component is small compared to the other components.

image

src/utils/AppRouteProvider.jsx
aerlaut: Should I make a ticket for this? This will be irrelevant with the SQL changes.

src/utils/AppRouteProvider.jsx
kafkasl: could you add then that description as documentation to the `AppRouteProvider`. Have you checked that no other calls are being made to `router.push`? if there are other calls, we might want to consider this a refactor and have some more tickets to stop using `router.push`. Having both even for some time would induce to error.

src/utils/AppRouteProvider.jsx
kafkasl: yes that would be great. although it won't be irrelevant with SQL changes. SQL can keep track of insertion date easily, tracking access is more complicated and probably not worth it (easier to just update modification date from the API code when doing the update itself).

kafkasl: are we actually using both values `lastViewed` and `lastAnalyzed` for something?  or are we just using one a duplicating the info

aerlaut: Previously it would give an idea of when a project was last opened, and when an experiment within a project was last viewed (last analyzed). The idea was you may have multiple experiments per project, and you can see the last viewed date of each experiment.

aerlaut: So by removing projects, we don't need `lastAnalyzed` on project

src/utils/AppRouteProvider.jsx
aerlaut: refactored

src/utils/AppRouteProvider.jsx
kafkasl: I will remove them in another PR as agreed

aerlaut: Feedback from Vicky during standup on the 18th of Jan 2022 : We can delete last analyzed

src/utils/AppRouteProvider.jsx
aerlaut: Yes, I have checked there is no other places where `router.push` is used. I'll add a documentation for the component.

aerlaut: Documentation added in 3cd1744

src/pages/_app.jsx
```diff
@@ -78,12 +76,6 @@ const WrappedApp = ({ Component, pageProps }) => {
       setAmplifyConfigured(true);
     }
   }, [amplifyConfig]);
-  useEffect(() => {
-    // clear the store only if we navigate to a new experiment from data-management
-    if (router.route === '/data-management') {
-      dispatch(switchExperiment());
```
StefanBabukov, line 84: Will the store reset when you switch an experiment?

src/utils/AppRouteProvider.jsx
```diff
@@ -26,27 +60,45 @@ const AppRouteProvider = (props) => {
...
     continueNavigation(nextRoute, hardNavigate);
   };
 
   const navigateTo = (
-    nextRoute,
+    module,
+    params = {},
     refreshPage,
-  ) => handleRouteChange(router.pathname, nextRoute, refreshPage);
+  ) => handleRouteChange(router.pathname, module, params, refreshPage);
```
StefanBabukov, line 92: Can you not just use `handleRouteChange` why do you wrap it with `navigateTo` ?

src/pages/_app.jsx
aerlaut: Yes, the call to `switchExperiment` is moved into the `AppRouteProvider` component. This action will be dispatched if the user is moving from `data-management` to other parts of the application.

src/utils/AppRouteProvider.jsx
aerlaut: So that the previous route (`router.pathname`) can be passed in as the first parameter. If we use `handleRouteChange` directly (in the component), then the component would know about the. Alternatively, we can have a global `router` variable in `AppRouteProvider` which `handleRouteChange` can access. However, I don't think it is a good idea as it would strongly couple `handleRouteChange` with `router`.

StefanBabukov: But we have `router` already. The pathname can just be removed as an argument in `handleRouteChange` and just get it inside the function right?

src/utils/AppRouteProvider.jsx
```diff
@@ -26,27 +60,45 @@ const AppRouteProvider = (props) => {
...
+    const lastViewed = moment().toISOString();
+
+    dispatch(switchExperiment(experimentId));
+    dispatch(updateExperiment(experimentId, { lastViewed }));
+    dispatch(updateProject(projectUuid, { lastAnalyzed: lastViewed }));
+  };
+
   const continueNavigation = (nextRoute, hardNavigate) => {
     // Hard navigate, causing the page to refresh and fetch data from server
     if (hardNavigate) window.location.href = nextRoute;
```
StefanBabukov: I think you dont need this `window.location.href` router.Push() is good enough?

StefanBabukov: Or am i missing something

src/utils/AppRouteProvider.jsx
aerlaut: yea, but that's what I'm trying to avoid, using the global `router` variable, which introduces the coupling. It's also nice to have the `pathname` scoped locally inside the variable.

src/utils/AppRouteProvider.jsx
aerlaut: It's different. `window.location.href` is used when you want to do "hard navigation", which means you send a request to the server for another page, and the server sends the new page, potentially with new server-side information. Without caching or hydration, this resets all client-side states.

`router.push()` is used for "soft navigation", i.e. the single-page app renders a new page on the client-side only, without requesting a new page to the server. As the page is still the same, this means all the states are kept.

aerlaut: Right now no part of the application uses this feature. When navigating to an outside link, we use the `Link` component directly. I can remove this part and make the code simpler.

StefanBabukov (approved):
looks good

src/utils/AppRouteProvider.jsx
aerlaut: removed as suggested

vickymorrison (approved):

PR #609, opened 2022-01-17
[BIOMAGE-1467] Move redux cell sets to immerjs

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1467

https://ui-martinfosco-ui609-api290.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/api/pull/290

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


cosa65:
I moved away from the pattern of "change something in a cell set so upload all of the cell sets again" by using the `json-merger` library introduced for updates on cell sets done by the worker.

On the UI side some of this updates are a bit more complex compared to the worker's one so the json merger commands are more complicated.

While doing this I noticed that for example, we allow reordering the cell sets (it was broken, but we do) within a cell class BUT we also allowed the cell sets to be moved between cell classes which doesn't really make sense because we have different kids of permissions over them. After consulting with iva it seems this code was a little bit outdated so I restricted it to only allow to move cellsets inside a cell class.

Apart from that I performed the actual change of cellSets reducers to immerjs.

Due to the `json-merger`changes, after this PR is merged the UI won't be using the PUT endpoint of the api cell sets, this means that we will no longer be performing requests of more or les 10MB for the biggest cell sets objects we have (so it will speed up part of the process involving cell sets)

src/redux/actions/cellSets/reorderCellSet.js
```diff
@@ -0,0 +1,70 @@
...
+              },
+            },
+          },
+        ],
+      },
+    },
+  }]
+);
+
+const reorderCellSet = (
```
cosa65, line 28: This action creator/action combo is replacing the one of `updateCellSetHierachy` because reordering the cell sets was the only actual usage `updateCellSetHierachy` was receiving

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
```diff
@@ -71,33 +58,41 @@ describe('HierarchicalTree', () => {
...
-    expect(childrenKeys).toEqual(['2a', '3a', '1a']);
+    expect(mockOnCellSetReorder).toHaveBeenCalledTimes(1);
+    const [cellSetKey, newPosition] = mockOnCellSetReorder.mock.calls[0];
+
+    expect(cellSetKey).toEqual('1a');
+    expect(newPosition).toEqual(1);
   });
 
-  it('Can have child component change parent', () => {
+  it('Can\'t have child component change parent', () => {
```
aerlaut, line 110: Why do you have to use `\'`, can't you just use "... can't..."?

cosa65: because of the single quotes of the string, and using double quotes would go against what we usually use, right?

src/components/EditableField.jsx
```diff
@@ -42,6 +42,7 @@ const EditableField = (props) => {
   }, [editing]);
 
   const deleteEditableField = (e) => {
+    e.stopPropagation();
```
aerlaut, line 45: Why do we need to add `e.stopPropagation` here?

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
aerlaut: oh ok

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
```diff
@@ -42,156 +43,31 @@ const HierarchicalTree = (props) => {
...
+    const fromPosition = parseInt(posFromArray[2], 10);
 
-    if (shouldUpdateState) {
-      props.onHierarchyUpdate(newTreeData);
-    }
+    // If not in the same cellClass, ignore
+    if (!_.isEqual(posFromArray[1], posToArray[1])) return;
+
+    // If was dropped in same place, ignore
+    if (fromPosition === dropPosition) return;
```
aerlaut, line 63: What is the value of `dropPosition`?

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
```diff
@@ -42,156 +43,31 @@ const HierarchicalTree = (props) => {
...
+
+    // If not dropped in gap, ignore
+    if (!dropToGap && !node.rootNode) return;
```
aerlaut, line 68: Why do you need the `!node.rootNode` condition?

src/redux/actionTypes/cellSets.js
```diff
@@ -74,7 +74,7 @@ const CELL_SETS_ERROR = `${CELL_SETS}/error`;
 export {
   CELL_SETS_LOADING, CELL_SETS_LOADED,
   CELL_SETS_CREATE,
-  CELL_SETS_UPDATE_PROPERTY, CELL_SETS_UPDATE_HIERARCHY, CELL_SETS_SET_SELECTED,
+  CELL_SETS_UPDATE_PROPERTY, CELL_SETS_REORDER, CELL_SETS_SET_SELECTED,
```
aerlaut, line 77: Why are some actions in one line and other actions on multiple lines?

src/redux/reducers/cellSets/helpers.js
```diff
@@ -28,14 +30,18 @@ const createPropertiesFromTree = (data) => {
           type,
         };
 
+        if (!rootNode && !_.isNil(parentNode)) {
```
aerlaut, line 33: Why do you use `!_.isNil(parentNode)`? Can't we do the same check with `!parentNode`

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
```diff
@@ -1,26 +1,51 @@
...
-import waitForComponentToPaint from '../../../../utils/tests/waitForComponentToPaint';
+
+import {
+  EditOutlined, DeleteOutlined, // , CheckOutlined, CloseOutlined,
+} from '@ant-design/icons';
+
+import HierarchicalTree from 'components/data-exploration/hierarchical-tree/HierarchicalTree';
+import waitForComponentToPaint from 'utils/tests/waitForComponentToPaint';
+
 import '__test__/test-utils/setupTests';
```
aerlaut: Why do you need the `setupTests`?

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
```diff
@@ -1,26 +1,51 @@
...
 
 const experimentId = 'asd';
```
aerlaut: can you import `fake` from `__test__/test-utils/constants` and use `fake.EXPERIMENTT_ID` instead? It's closer to the real data, so that would be good.

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
```diff
@@ -71,141 +78,399 @@ describe('HierarchicalTree', () => {
     const dropInfo = {
       dragNode: {
         ...firstChild,
-        props: { eventKey: firstChild.key },
+        pos: '0-0-0',
```
aerlaut, line 79: So these `pos` properties are only available during the drag event? It's not saved along with the object?

src/__test__/redux/actions/cellSets/createCellSet.test.js
```diff
@@ -45,14 +51,15 @@ describe('createCellSet action', () => {
     expect(firstAction).toMatchSnapshot();
   });
 
-  it('Last action dispatches cellSetSave event', async () => {
+  it('Send fetch to the api', async () => {
```
aerlaut: A bit more complete description : `Send fetch to the API when creating cell set`

src/__test__/redux/actions/cellSets/deleteCellSet.test.js
```diff
@@ -41,13 +41,15 @@ describe('deleteCellSet action', () => {
     expect(firstAction).toMatchSnapshot();
   });
 
-  it('Last action dispatches cellSetSave event', async () => {
+  it('Send fetch to the api', async () => {
```
aerlaut: A bit more complete : `Send fetch to the API when a cell set is deleted`

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
```diff
@@ -0,0 +1,76 @@
...
+import waitForActions from 'redux-mock-store-await-actions';
+import { waitFor } from '@testing-library/react';
+
+import reorderCellSet from 'redux/actions/cellSets/reorderCellSet';
+import { CELL_SETS_REORDER } from 'redux/actionTypes/cellSets';
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+
+import initialState from 'redux/reducers/cellSets/initialState';
+
+import '__test__/test-utils/setupTests';
```
aerlaut: Why do you need `setupTests`?

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+enableFetchMocks();
+const mockStore = configureStore([thunk]);
```
aerlaut, line 14: Can we use the real store instead of a mock store?

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
```diff
@@ -0,0 +1,76 @@
...
+
+    // Fetch was called correctly
+    expect(fetch).toHaveBeenCalledTimes(1);
+
+    const [url, body] = fetch.mock.calls[0];
+    expect(url).toEqual('http://localhost:3000/v1/experiments/1234/cellSets');
+    expect(body).toMatchSnapshot();
+  });
+
+  it('Does not dispatch on fetch error', async () => {
```
aerlaut: What does not dispatch on fetch error? Can you add a check to test that whatever should be dispatched is not dispatched?

src/redux/actions/cellSets/updateCellSetProperty.js
```diff
@@ -1,25 +1,102 @@
...
+
+    pushNotificationMessage('error', endUserMessages.ERROR_SAVING);
+  }
+
   await dispatch({
     type: CELL_SETS_UPDATE_PROPERTY,
     payload: {
       experimentId,
-      key,
+      cellSetKey: key,
```
aerlaut: If I look at the reducer (src/redux/reducers/cellSets/cellSetsUpdateProperty.js, line 8), it seems to still be getting `key` instead of `cellSetKey`. Can you check whether this is correct?

src/redux/actions/cellSets/reorderCellSet.js
```diff
@@ -0,0 +1,70 @@
...
+      console.error(`fetch ${url} error ${e.message}`);
+    }
+
+    pushNotificationMessage('error', endUserMessages.ERROR_SAVING);
+  }
+
+  await dispatch({
+    type: CELL_SETS_REORDER,
+    payload: {
+      experimentId,
```
aerlaut: This `experimentId` property doesn't seem to be accessed in the reducer. Do you need it?

src/components/EditableField.jsx
cosa65: All the other actions in `EditableField` have this `e.stopPropagation`, you can check `onSubmit` and `onCancel`.

From a more practical standpoint, if I remove the `e.stopPropagation` from `onDelete` then when I delete a cellSet then that `click` event propagates up to the `onCheck` (triggered by antd on `<Tree>`).

As a consequence of this the cellClass of the deleted cellSet gets checked on/off when a delete icon is clicked.

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
cosa65: It is the index on which the item is dropped and the item should be inserted into.

However, it changes to a different value if the mouse is dragged more towards the right, so when it is inserted into an item on its same level, like for example if we wanted to insert a cellSet into another cellSet (although in this case we get a `dropToGap === true` so that's the way we can deal with that case).

There is only one case in which `dropToGap === true` is fine, and that is when the node we are inserting our item into is the rootNode, and this case is only really relevant to us when the item is inserted at `dropPosition === 0`.I know it is kind of confusing but it's the way `Tree` was coded in antd, and I'm sure it makes more sense if you allow multiple levels on the `Tree` (instead of only 2 like we need).

cosa65: If you want us to go over this in more detail we can have a call tomorrow, let me know

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
cosa65: answered in https://github.com/hms-dbmi-cellenics/ui/pull/609#discussion_r791046456

src/redux/actionTypes/cellSets.js
cosa65: I guess to have some kind of grouping? They need to be split into lines anyways.

src/redux/reducers/cellSets/helpers.js
cosa65: With `isNil` we are checking for a `null` or `undefined`, so it is more correct compared to using !`parentNode` which also allows `0` and `false`. 

I know we could assume `0` and `false` never happens, but with javascript it unfortunately it can happen and sometimes we end up with falses or something. I prefer to go with the more strict option when possible so that we catch those `false` or `0` that shouldn't be there.

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
cosa65: We aren't saving any object, you can check the reorder action and its creator, it is only using `experimentId`, `cellSetKey`, `newPosition`, which is the only data we actually need.

src/__test__/components/data-exploration/hierarchical-tree/HierarchicalTree.test.jsx
cosa65: Removed

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
cosa65: Added the check

src/redux/actions/cellSets/updateCellSetProperty.js
cosa65: Nice! thanks, fixed

src/redux/actions/cellSets/reorderCellSet.js
cosa65: Actually we send it everywhere in cellSets unnecessarily, so I removed it from everywhere

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
cosa65: For action creators shouldn't we focus on checking the actions they dispatch? Why do we need the real store?

Also we don't need to see the reactions to the store changing, so I don't see what's the benefit.

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
aerlaut: Can you add this information in a comment?

src/components/data-exploration/hierarchical-tree/HierarchicalTree.jsx
aerlaut: Can you add this information in a comment explaining why you need the `rootNode` check?

src/redux/actionTypes/cellSets.js
aerlaut: oh ok

src/redux/reducers/cellSets/helpers.js
aerlaut: Are the values `0` and `false` possible in this case?

src/__test__/redux/actions/cellSets/reorderCellSet.test.js
aerlaut: Hmm... ok

src/redux/reducers/cellSets/helpers.js
aerlaut: In this case, parent node would be an object if it's defined right?

src/redux/actions/cellSets/updateCellSetSelected.js
```diff
@@ -12,7 +12,6 @@ const updateCellSetSelected = (experimentId, keys, tab) => async (dispatch, getS
   await dispatch({
     type: CELL_SETS_SET_SELECTED,
     payload: {
-      experimentId,
       keys,
```
aerlaut, line 15: Can you remove the `experimentId` parameter from the [function signature as](https://github.com/hms-dbmi-cellenics/ui/blob/1e8b7c97357398b1f1a6f0e928631c2a907800e7/src/redux/actions/cellSets/updateCellSetSelected.js#L3) well?

src/redux/reducers/cellSets/cellSetsUpdateProperty.js
```diff
@@ -1,13 +1,14 @@
...
 
-  const newState = _.cloneDeep(state);
+const cellSetsUpdateProperty = produce((draft, action) => {
+  const { cellSetKey, dataUpdated } = action.payload;
 
-  _.merge(newState.properties[key], dataUpdated);
-
-  return newState;
-};
+  // Adding original because jest fails otherwise, but in reality it works also without original
```
aerlaut: What do you mean by this comment?

src/redux/actions/cellSets/runCellSetsClustering.js
```diff
@@ -43,7 +43,7 @@ const runCellSetsClustering = (experimentId, resolution) => async (dispatch, get
           throw err;
         }
 
-        dispatch(updateCellSetsClustering(experimentId));
```
aerlaut: Looking [updateCellSetsClustering()](https://github.com/hms-dbmi-cellenics/ui/blob/1e8b7c97357398b1f1a6f0e928631c2a907800e7/src/redux/actions/cellSets/updateCellSetsClustering.js#L7), it seems you still need the experimentId.

src/redux/reducers/cellSets/cellSetsUpdateProperty.js
cosa65: Removed the reason for this comment but not the comment :P

cosa65: I was trying out some stuff and at a certain point used `original` from immerjs, later I removed it

src/redux/actions/cellSets/runCellSetsClustering.js
cosa65: fixed

src/redux/reducers/cellSets/helpers.js
cosa65: `0` and `false` are possible in the sense that javascript isn't going to throw if you assign a bool or an int to a string. 

It looks like it doesn't make sense to have `0` and `false` here so we shouldn't worry about it, but I have seen those values show up in some places they don't make sense, so if we can have a little more restrictive check why not do it?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -207,7 +214,8 @@ const Embedding = (props) => {
   const updateCellsSelection = (selection) => {
     if (Array.from(selection).length > 0) {
       setCreateClusterPopover(true);
-      setSelectedIds(selection);
+      const selectedIdsToInt = new Set(Array.from(selection).map((id) => parseInt(id, 10)));
```
StefanBabukov, line 217: Why are the cellIds not int in the first place? Is that how the scatterplot returns selections?

src/redux/actions/cellSets/createCellSet.js
```diff
@@ -17,25 +39,58 @@ const createCellSet = (experimentId, name, color, cellIds) => async (dispatch, g
...
-  pushNotificationMessage('info', endUserMessages.NEW_CLUSTER_CREATED);
+  const url = `/v1/experiments/${experimentId}/cellSets`;
+
+  try {
+    const response = await fetchAPI(
+      url,
+      {
+        method: 'PATCH',
+        headers: {
+          'Content-Type': 'application/boschni-json-merger+json',
```
StefanBabukov, line 75: What made you use this application with the `CellSetJsonMerger` from the way it was before, is it faster?

cosa65: updateCellSets is uploading the whole cellsets object, this only performs the update we need to perform.

In some cases (experiment with a cellSets in S3 of 10 mb) before something like a cellset rename took very long.

src/components/data-exploration/embedding/Embedding.jsx
cosa65: yep, vitessce returns strings

aerlaut (approved):
Tested in staging, LGTM 👍

PR #614, opened 2022-01-20
Refactor react routes

aerlaut:


src/__test__/components/data-exploration/generic-gene-table/LaunchPathwayAnalysisModal.test.jsx
```diff
@@ -60,15 +59,13 @@ const pantherDBResponse = {
 const renderPathwayAnalysisModal = async (filtersApplied = false) => {
   await act(async () => {
     render(
-      <SWRConfig value={{ provider: () => new Map() }}>
```
ivababukova, line 63: What was this `SWRConfig` for and why is it not needed anymore?

src/components/ContentWrapper.jsx
```diff
@@ -48,33 +47,33 @@ const ContentWrapper = (props) => {
...
 
   const activeProject = useSelector((state) => state.projects[activeProjectUuid]);
-
   const samples = useSelector((state) => state.samples);
 
+  // This use effect block is used to get the active experiment ID in cases where experimentId is not evident in the route
+  // This happens in data-management, where we set the active experimentId to be the experimentId of the project
+  // ExperimentId in the route is used as a higher priority
   useEffect(() => {
     if (!activeProjectExperimentID && !routeExperimentId) return;
```
ivababukova, line 62: if this useEffect is used only for data management module, you can also replace this if statement with `if(currentModule !== modules.DATA_PROCESSING)`

src/__test__/components/data-exploration/generic-gene-table/LaunchPathwayAnalysisModal.test.jsx
aerlaut: This was a left-over from the previous implementation to get PantherDB species. I used `useSWR` previously for fetching pantherDB species from their API (http://pantherdb.org/services/openAPISpec.jsp). It worked in local, but failed in staging (HTTP vs HTTPS). So we do not `useSWR` any more, and just store the species list in a JSON.

`useSWR` caches results, which is a problem in the tests. This `SWRConfig` is used to clear out cache in tests (https://swr.vercel.app/docs/advanced/cache#reset-cache-between-test-cases). However, because we are not using `useSWR` anymore and I forgot to remove this.

src/components/ContentWrapper.jsx
aerlaut: Hmm... you're right. I'll try it out.

aerlaut: This is working, I've implemented the changes.

src/utils/AppRouteProvider.jsx
```diff
@@ -43,7 +45,19 @@ const AppRouteProvider = (props) => {
   const router = useRouter();
   const dispatch = useDispatch();
 
+  const pathLookup = useRef(Object.entries(PATH_STUBS));
```
kafkasl: why do we need that `useRef` ? afaik it's used to persist a value and avoid re-renderings when updating a value, but we never update the `pathLookup` variable

aerlaut: So taht it's not calculated again and again on every rerender.

src/utils/AppRouteProvider.jsx
```diff
@@ -43,7 +45,19 @@ const AppRouteProvider = (props) => {
...
+
   const [renderIntercept, setRenderIntercept] = useState(null);
+  const [currentModule, setCurrentModule] = useState(module.DATA_MANAGEMENT);
+
+  useEffect(() => {
+    pathLookup.current.find(([moduleName, path]) => {
```
kafkasl: the `find` function returns the first element that matches the condition but you are ignoring the return value and instead doing the logic inside the condition matching. You should do instead:

```
const moduleName = pathLookup.find(([mn, p]) => p === '/data-processing')[0]
setCurrentModule(moduleName);
```

and I don't think you need to use `.match` which is designed to regex when you are just comparing normal strings.

aerlaut: the value of `router.pathname` is the real pathname (e.g. `http://localhost:5000/experiments/8be39ad79c4f16e611650f8c90a4d3f6/data-processing`). That's why we need to match it to the path, which is only a stub (e.g. `/data-processing`).

I purposefully use `Array.find()` so that it returns on the first instance of `true`, which is when the pathname matches the path and the lookup doesn't have to continue.

kafkasl (approved):

src/utils/AppRouteProvider.jsx
aerlaut: changed

src/utils/AppRouteProvider.jsx
aerlaut: `PATH_STUBS` is a small object, overhead of importing `useRef` might not be worth it, so removed.

PR #617, opened 2022-01-21
[BIOMAGE-1688] - Add browser banner

aerlaut:


aerlaut:
There are 2 major ways to detect browser : 
1. Checking the user agent (UA) string via `navigator.userAgent`
2. Browser-specific feature detection. 

You can see both implemented in this Stackoverflow post : 
https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser

The top voted answer suggests feature detection is the way to go. However, I feel it is not robust because features may get added and removed in subsequent browser updates.

The main reason people say UA checking is not robust is because it can easily be spoofed by setting the `navigator.userAgent` value directly. UA spoofing is done to obfuscate traffic (https://help.trafficguard.ai/knowledge/user-agent-spoofing). This impacts monitoring activities (e.g. for ads). Our purpose is to let the user know that their browser is not detected, not for tracking user's browsers. Aside from that, I can also pass the feature checks by manually adding the properties via a browser onsole.

So, I think in our case, checking the UA string is a more reliable and easier way to check what browser the user is using. There is no advantage to the user spoofing their UA string, as it would not make browser-specific problems go away.

vickymorrison:
The warning message has a ' at the end of it that should be removed.

vickymorrison (approved):

src/__test__/components/ContentWrapper.test.jsx
```diff
@@ -104,6 +109,14 @@ describe('ContentWrapper', () => {
 
     store = makeStore();
 
+    navigator.userAgent = chromeUA;
```
ivababukova, line 112: if you already do `Object.defineProperty(` in line 54 and you set chromeUA there, why do you have to set it again in here as user agent?

aerlaut: To set a ground level for all the tests. This is changed in line `252` to test for the display banner, but for the remaining tests, it's good to be sure that the banner doesn't exist, otherwise it may affect other tests (e.g. during text search using `getByText`).

ivababukova (approved):

PR #618, opened 2022-01-21
[BIOMAGE-1619] - Prevent multiple work result fetches

aerlaut:
# Description
Prevent multiple workrequest downloads

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1619

https://ui-agi-ui618-worker253.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/worker/pull/253

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -160,6 +160,8 @@ const DotPlotPage = (props) => {
 
   useEffect(() => {
     if (cellSetsLoading) return;
+    if (config?.useMarkerGenes && config.nMarkerGenes === 0) return;
+    if (config?.useMarkerGenes === false && config.selectedGenes.length === 0) return;
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: you could do `!config?.useMarkerGenes` here

Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Could you add some comments with explanation of each case?

src/utils/postRequestProcessing.js
```diff
@@ -1,20 +1,17 @@
-// Calculate ZScore
 const calculateZScore = (responseData) => {
-  Object.keys(responseData).forEach((gene) => {
-    const { mean, stdev } = responseData[gene].rawExpression;
+  const dataWithZScore = Object.entries(responseData).reduce((acc, [gene, value]) => {
```
cosa65, line 2: What changed here?

src/utils/postRequestProcessing.js
```diff
@@ -1,20 +1,17 @@
...
+    const { mean, stdev, expression } = value.rawExpression;
+    const zScore = expression.map((x) => (x === null ? null : ((x - mean) / stdev)));
```
cosa65: x `isNil`?

src/utils/work/fetchWork.js
```diff
@@ -68,20 +68,28 @@ const fetchGeneExpressionWork = async (
...
-      null,
       {
         ETagPipelineRun: qcPipelineStartDate,
         ...extras,
       },
     );
   }
 
+  if (!workResultReady) {
+    console.debug(`No response immediately resolved for ${body} (ETag: ${ETag}) -- this is probably an event subscription.`);
```
cosa65: what does "this is probably an event subscription" mean?

src/utils/work/seekWorkResponse.js
```diff
@@ -5,34 +5,43 @@ import fetchAPI from 'utils/fetchAPI';
...
   requestProps = {},
 ) => {
-  console.error('seek from api', body);
+  console.error('dispatching work request', body);
   const { default: connectionPromise } = await import('utils/socketConnection');
   const io = await connectionPromise;
 
   const timeoutDate = moment().add(timeout, 's').toISOString();
   const authJWT = await getAuthJWT();
-  const socketId = !eventCallback ? io.id : 'broadcast';
```
cosa65, line 35: Why was `broadcast` removed? Don't we use it for clustering?

src/utils/work/fetchWork.js
```diff
@@ -139,30 +147,36 @@ const fetchWork = async (
...
-      eventCallback,
       {
         PipelineRunETag: qcPipelineStartDate,
         ...extras,
       },
     );
   }
 
-  if (!response) {
+  if (!workResultReady) {
```
cosa65: `workResultReady` would be false when the task timed out?

src/utils/work/seekWorkResponse.js
```diff
@@ -5,34 +5,43 @@ import fetchAPI from 'utils/fetchAPI';
...
+
+    throwResponseError(response);
+  }
+
+  const { signedUrl } = await response.json();
   const storageResp = await fetch(signedUrl);
+
   if (!storageResp.ok) {
-    return null;
+    throwResponseError(storageResp);
```
cosa65, line 25: This didn't throw before, right? Is it caught later?

src/utils/work/seekWorkResponse.js
```diff
@@ -44,56 +53,33 @@ const seekFromAPI = async (
...
-  if (eventCallback) {
-    io.off(`${experimentId}-${body.name}`);
+  const timeoutPromise = new Promise((resolve, reject) => {
+    const id = setTimeout(() => {
+      clearTimeout(id);
+      reject(new WorkTimeoutError(timeoutDate, request));
+    }, timeout * 1000);
+  });
 
-    io.on(`${experimentId}-${body.name}`, (res) => {
```
cosa65, line 52: Why was this removed? What happens when we run a clustering now? Are cell sets updated?

cosa65: This socket is used for 'broadcast' type work requests (I saw the code for broadcast was removed as well)

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: It's different. In this case we want to check : 
- If they use marker genes, the number of genes should not be 0
- If they use custom genes, the list of genes is not empty

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Ok i'll add comment

src/utils/postRequestProcessing.js
aerlaut: Previously the function modifies its input. When I was fixing the test for marker heatmap, it fails because it says with `can not assign zScore to read only property`. So I checked and it turns out this function is modifying its input object. I don't think that's a good idea, so I refactored the function to return a copy of the response object, with the calculations included.

src/utils/postRequestProcessing.js
aerlaut: No, `x` is a member of the array, it will always be defined, it can not be undefined.

src/utils/work/fetchWork.js
aerlaut: This error happens when there is a work response returned from the API which does not come from the result of a work request. One of them is the subscription of the UI to the socket channel. I don't think we need this anymore because workResultReady will be either `true`, or throw an error. I'll remove this.

src/utils/work/fetchWork.js
aerlaut: No if it times out, it will throw an error. `workResultReady` is not needed, because it either returns true, or it throws an error.

src/utils/work/seekWorkResponse.js
aerlaut: Yea, every implementation of `fetchWork` that we have right now wraps the function call in a `try...catch` block.

src/utils/work/seekWorkResponse.js
aerlaut: This is never used. If you take a look at the current call to `seekFromAPI` in `fetchWork` (https://github.com/hms-dbmi-cellenics/ui/blob/master/src/utils/work/fetchWork.js#L77), the function parameter is explicitly set to `null`.

src/utils/work/seekWorkResponse.js
aerlaut: Yes, broadcast was never used, so I removed it. If you check how `seekFromAPI` is called, the parameter which takes `eventCallback` is explicitly set to null.

I moved the event handling to the end of `fetchWork` (https://github.com/hms-dbmi-cellenics/ui/pull/618/files#diff-37e452fd6bfee4dd173ebb5f7be9fb0e91f42bea39f0d203ce497608dab65edbR176-R178), so that the event handlers are still called at the end of the `fetchWork` request.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: added

src/utils/work/fetchWork.js
aerlaut: removed

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Nice! I just meant `config?.useMarkerGenes === false` to `!config?.useMarkerGenes`

src/utils/postRequestProcessing.js
cosa65: Nice!

src/utils/postRequestProcessing.js
cosa65: you're right

cosa65: ok, very nitpicky, but what if you switch it so that the case in which we do something comes up first?

`x !== null ? ((x - mean) / stdev) : null;`

src/utils/work/seekWorkResponse.js
cosa65: But you linked me to `fetchGeneExpressionWork`. In `fetchWork` we do use it https://github.com/hms-dbmi-cellenics/ui/blob/master/src/utils/work/fetchWork.js#L149

src/utils/work/fetchWork.js
cosa65: Could you check that the timeout works? Because locally I tried killing the worker and the timeout isn't reaching the handler to show the "try again"

src/utils/work/seekWorkResponse.js
cosa65: If the broadcast isn't being used anymore I think then there's something in the worker we aren't using anymore too (that handles 'broadcast').

src/utils/postRequestProcessing.js
aerlaut: hahaha ok

src/utils/work/seekWorkResponse.js
aerlaut: Oh right. Right now `eventCallback` is only used when running celllSets clustering. I think this was meant to automatically update the cell sets automatically if we have multiple users looking at the same experiment in different browsers. This callback will automatically update the cell sets if clustering is changed in one of the browsers. 

If the socketId is passed as `broadcast` by the UI, the worker will send the response to the channel `${experimentId}-${body.name}` (https://github.com/hms-dbmi-cellenics/worker/blob/master/python/src/worker/response.py#L93-L99). Right now this is the channel that's listened to in the UI for the work request. However, we only open a listener for this channel when the UI dispatches `runCellSetsClsutering`. If the user never run cell sets clustering (e.g. jump straight from data management to data exploration), the listener will not be set up, and changes to the cell set by a user will not cause changes to the cell sets of other users looking at the same experiment.

aerlaut: To solve this issue, we can : 
**1. Try to use other channels that we're currently always listening to**

The only socket channel we listen to all the time is `ExperimentUpdates-${currentExperimentId}` which is initialized in the [content wrapper](https://github.com/hms-dbmi-cellenics/ui/blob/c9de20448ba52adcd9017f0098eec3fe319fad05/src/components/ContentWrapper.jsx#L118). Right now the callback for this socket handles updates for GEM2S and QC. We can make a switch in the [experimentUpdatesHandler](https://github.com/hms-dbmi-cellenics/ui/blob/d025c80fd02ed497ca3bc30d694e33ceba9ac8ec/src/utils/experimentUpdatesHandler.js#L12) to also update the cell sets if there is a re-clustering.

Pros : 
- Cleaner, we don't have to add another socket to listen to

Cons : 
- More work in `experimentUpdatesHandler`, but it still falls in the scope of the handler's work.

**2. Make the UI always listen to the 'broadcast' channel (e.g. `${experimentId}-${body.name}`) and handle the cellsets change there**

We can do this by initializing a socket to the channel and registering the callback to the listener of the broadcast channel.

Pros : 
- Dedicated channel to listen for updates. Cleaner binding.

Cons:
- Duplicated code (more channels to listen to)
- Adding a listener on a channel which users may rarely use, especially if they have found a good cluster to use.

Looking at the 2 options, I personally would rather go for the 1st option.

src/utils/work/seekWorkResponse.js
aerlaut: Turns out the broadcast is still used to sync cell sets for concurrent users looking at the same experiment. See this comment : https://github.com/hms-dbmi-cellenics/ui/pull/618#discussion_r794847974

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I think it's better to check for the initialization of `config` above and remove the `?.` checks. I'll do that.

src/utils/work/fetchWork.js
```diff
@@ -68,20 +68,22 @@ const fetchGeneExpressionWork = async (
...
         ...extras,
       },
     );
   }
 
+  response = await seekFromS3(ETag, experimentId);
+
   if (!response[missingGenes[0]]?.error) {
     // Preprocessing data before entering cache
     const processedData = calculateZScore(response);
```
ivababukova, line 96: why do you calculate zscore here?

src/utils/work/fetchWork.js
```diff
@@ -1,14 +1,14 @@
 /* eslint-disable no-underscore-dangle */
-import hash from 'object-hash';
+import { MD5 } from 'object-hash';
```
ivababukova, line 2: Feel free to keep it like this if you think it's out of scope, but we need to change this hashing algorithm to something more secure. Right now it is not a problem, but in the future if we decide to use this hash for something else, it can become a big problem.

src/utils/work/seekWorkResponse.js
```diff
@@ -5,34 +5,43 @@ import fetchAPI from 'utils/fetchAPI';
...
+
   if (!storageResp.ok) {
-    return null;
+    throwResponseError(storageResp);
   }
 
   return unpackResult(storageResp);
 };
 
-const seekFromAPI = async (
```
ivababukova, line 21: why renaming to `dispatchWorkRequest`? `seekFromAPI` sounds clearer to me

src/utils/work/seekWorkResponse.js
cosa65: Some time ago I believe we listened to the same socket for qc, gem2s and worker updates (so clustering updates), so I believe the problem here is that gem2s and qc had their socket changed but the worker wasn't so it became broken. 

To solve this I think the best idea is to bring it back in line with the original solution.

src/utils/work/fetchWork.js
```diff
@@ -139,30 +141,30 @@ const fetchWork = async (
...
         ...extras,
       },
     );
   }
 
-  if (!response) {
-    console.debug(`No response immediately resolved for ${body} (ETag: ${ETag}) -- this is probably an event subscription.`);
-    return response;
-  }
+  response = await seekFromS3(ETag, experimentId);
```
ivababukova, line 173: In `seekWorkResponse` in the deleted code, we used to also resolve or reject the promise like this after calling seekFromS3:
Screenshot 2022-01-31 at 13 47 30

Why is this not necessary anymore?

src/utils/work/fetchWork.js
aerlaut: After discussing with @ogibson-biomage and @cosa65, calculation of the z-score is still left in the UI because we still have to work on passing. Currently, gene expression data sent from the worker to the UI containing the truncated data and the raw data is quite big (300 MB). If we add z-score along with the data, it will increase the work request size by about 50%, making downloading the work request slower for our users with slow connection.

The gene expression data is actually sparse (e.g. contain a lot of `null` values). We use a lot of space to still include these `null` values to make it easy to access the gene expression data. However, this is an issue 

I've actually created a ticket for this (https://biomage.atlassian.net/browse/BIOMAGE-1710) before the discussion with Martin and Oliver. Oliver argues that this ticket should not be done before we address the gene expression work request. He is going to add the reason to **not** do this right now in the ticket, but as a downstream ticket after addressing the gene expression work request issues.

aerlaut: After discussing with @ogibson-biomage and @cosa65, calculation of the z-score is still left in the UI because we still have to work on passing. Currently, gene expression data sent from the worker to the UI containing the truncated data and the raw data is quite big (300 MB). If we add z-score along with the data, it will increase the work request size by about 50%, making downloading the work request slower for our users with slow connection.

The gene expression data is actually sparse (e.g. contain a lot of `null` values). We use a lot of space to still include these `null` values to make it easy to access the gene expression data. However, this is an issue 

I've actually created a ticket for this (https://biomage.atlassian.net/browse/BIOMAGE-1710) before the discussion with Martin and Oliver. Oliver argues that this ticket should not be done before we address the gene expression work request. He is going to add the reason to **not** do this right now in the ticket, but as a downstream ticket after addressing the gene expression work request issues.

src/utils/work/fetchWork.js
aerlaut: If we're going to change the algorithm, I think it should be part of another ticket

src/utils/work/seekWorkResponse.js
aerlaut: Because we do not 'seek the result from the API' anymore, as the function does not return the work result anymore. Rather, it's a 'here's the work request, and let me know if it's done'.

src/utils/work/fetchWork.js
aerlaut: This `resolve`/`reject` in this case is to return the result of fetching the data from S3 (`resolve` if success, `reject` if error). This is not necessary anymore because the call to S3 is moved out from `workflowDispatch` and is placed afterwards.

src/utils/work/seekWorkResponse.js
ivababukova: I get the point, it definitely isn't seeking from the API. I agree with the name change, sounds more accurate

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -62,20 +63,14 @@ jest.mock('lodash/sampleSize', () => ({
...
   '2-marker-genes': () => Promise.resolve(_.cloneDeep(markerGenesData2)),
 };
 
-let storeState = null;
-const loadAndRenderDefaultHeatmap = async () => {
-  storeState = makeStore();
-
-  storeState.dispatch(loadProcessingSettings(experimentId));
-  storeState.dispatch(loadBackendStatus(experimentId));
-
```
ivababukova, line 78: why are these actions not necessary anymore?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -89,26 +84,41 @@ const loadAndRenderDefaultHeatmap = async () => {
   });
 };
 
-const stalledResponse = () => new Promise(() => { });
+const mockAPIResponses = generateDefaultMockAPIResponses(experimentId, fake.PROJECT_ID);
+
+const stalledResponse = () => delayedResponse({ body: 'Not found', status: 404 }, 4000);
```
ivababukova: If it was me writing this, I don't think I would have `stalledResponse` based on `delayedResponse` function: it seems like too much effort for too little value of the abstraction. The code takes longer to understand -- one has to go back to `mockAPI` and look up `delayedResponse`, then has to come back here. And all that `delayedResponse` is, is resolving a promise after some time. I can easily imagine how people forget about this mock and don't use `delayedResponse` at all too. So I am not sure of the value of abstracting it.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -89,26 +84,41 @@ const loadAndRenderDefaultHeatmap = async () => {
...
     fetchMock.resetMocks();
     fetchMock.doMock();
 
-    seekFromAPI.mockClear();
+    fetchMock.mockIf(/.*/, mockAPI(mockAPIResponses));
+
+    seekFromS3
+      .mockReset()
+      .mockImplementationOnce(() => null)
+      .mockImplementationOnce((mockEtag) => mockWorkerResponses[mockEtag]());
```
ivababukova, line 109: You have the mock s3 in the beforeEach, but then you mock them again in most individual tests -- looks like it's better to remove the mock for s3 from the beforeEach in this case and have it in each test

ivababukova: unless this mock is actually used within the beforeEach function for the `loadProcessingSettings` and the `loadBackendStatus` actions?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
```diff
@@ -320,7 +335,7 @@ describe('HeatmapPlot', () => {
 
     // On hovering over heatmap tracks
     await act(async () => {
-      vitesscePropsSpy.setTrackHighlight(['4', 0, 1, 2]);
+      await vitesscePropsSpy.setTrackHighlight(['4', 0, 1, 2]);
```
ivababukova, line 337: why adding all these awaits?

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
```diff
@@ -67,42 +67,51 @@ const customAPIResponses = {
...
 
-const defaultMockResponses = _.merge(
+const mockAPIResponse = _.merge(
   generateDefaultMockAPIResponses(experimentId),
   customAPIResponses,
-  mockWorkerResponses,
 );
 
 const defaultProps = { experimentId };
 const dotPlotPageFactory = createTestComponentFactory(DotPlotPage, defaultProps);
```
Comment (ivababukova) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:76]: I had a peek at the `test-utils` folder and it seems to be going towards the "utils" direction, meaning that there is no real structure there and soon it will be messy. If possible, I suggest to introduce some order there , group and organise files there into different folders

Comment (ivababukova) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:76]: it doesn't have to be as part of this PR, but it seems to me that it needs to be done soon before it ends up being a mess

src/__test__/redux/actions/cellSets/__snapshots__/runCellSetsClustering.test.js.snap
```diff
@@ -6,17 +6,21 @@ Object {
 }
 `;
 
-exports[`runCellSetsClustering action Dispatches all required actions to update cell sets clustering. 2`] = `undefined`;
+exports[`runCellSetsClustering action Dispatches all required actions to update cell sets clustering. 2`] = `
+Object {
+  "type": "cellSets/clusteringUpdated",
+}
+`;
```
ivababukova: how come is this not returning undefined now?

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
aerlaut: Right, I've replaced stalled response to just directly use `delayedResponse`

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
aerlaut: No it's not. I'm also putting it here so that this becomes the default behaviour. If the test requires custom behaviour, then the test can implement its own implementation of `mockSeekFromS3`. If we remove it from here, we'll have to add the default implementation for every test, whether the test requires a custom implementation or not, and it would make it difficult to change the base behaviour if we need to.

ivababukova: I noticed that the base behaviour is used only in 2 places. The majority of the tests override the behaviour

aerlaut: The base case is used in 5 tests, while 4 tests uses custom implementation.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
aerlaut: so that the async functions will resolve before moving on to the next line. If we don't do this, the test will continue without waiting for the resolution of the async calls. This happens a lot in test cases, and this is the source of the `UnhandledPromiseRejectionWarning` logs when we run tests. Sometimes, when we correctly wait for the promise to resolve, test fails because there are errors thrown when we resolve the promise chain.

src/__test__/redux/actions/cellSets/__snapshots__/runCellSetsClustering.test.js.snap
aerlaut: `runCellSetsClustering` used to have a callback in its dispatch to `fetchWork`: https://github.com/hms-dbmi-cellenics/ui/blob/master/src/redux/actions/cellSets/runCellSetsClustering.js#L41-L47. This callback runs `updateCellSetsClustering` which dispatches the action.

This PR removes that callback and move its implementation to: https://github.com/hms-dbmi-cellenics/ui/blob/0bcde96857db3fbb8a8f5798a8a17de8ca6a0002/src/utils/experimentUpdatesHandler.js#L70-L78. Because `updateCellSetsClustering` is not called as a callback anymore, the action is not logged in the store. Hence it is `undefined`.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
aerlaut: Because it is already being called inside the component. We do not need to preload it into the store, the component will do it as part of its initialization.

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
ivababukova: yes, you are right actually, somehow I got confused. Cool, let's leave it in the beforeEach

src/__test__/components/data-exploration/heatmap/HeatmapPlot.test.jsx
ivababukova: makes sense!

src/redux/actions/cellSets/runCellSetsClustering.js
```diff
@@ -33,19 +37,22 @@ const runCellSetsClustering = (experimentId, resolution) => async (dispatch, get
     type: CELL_SETS_CLUSTERING_UPDATING,
   });
 
+  const ETag = createObjectHash({
+    experimentId, body, qcPipelineStartDate,
+  });
+
```
ivababukova: for the rest of the actions that submit a work request, none of them calls `createObjectHash` and `dispatchWorkRequest`. They call `fetchWork` and the hash creation and all the details are handled there. Why do we need to break from this pattern and call `createObjectHash` and `dispatchWorkRequest` now for this action?

src/redux/actions/cellSets/updateCellSetsClustering.js
```diff
@@ -8,7 +8,7 @@ const updateCellSetsClustering = (experimentId) => async (dispatch) => {
   try {
     dispatch({ type: CELL_SETS_CLUSTERING_UPDATED });
 
-    dispatch(loadCellSets(experimentId));
+    dispatch(loadCellSets(experimentId, true));
```
ivababukova, line 11: why do we need to force reload?

src/redux/actions/cellSets/runCellSetsClustering.js
aerlaut: Yea, I have to rethink about this, I agree there should be a better way to implement this. This action `dispatchWorkRequest` but does not `fetchWorkResult` because it does not directly fetch the work result directly from S3 and use the returned work result. This action dispatches work request, which, when completed, will tell the browser to reload cell sets.

Arguably, this action indirectly **fetch work result from S3**, via a call to `loadCellSets` in https://github.com/hms-dbmi-cellenics/ui/blob/0bcde96857db3fbb8a8f5798a8a17de8ca6a0002/src/redux/actions/cellSets/updateCellSetsClustering.js#L11. This function gets the cell set from the `cellsets` bucket, using the `experimentId` as the object key, unlike work results which looks for `work-results` with `Etag` as the object key. So, if we use `fetchWork`, this call causes 404 errors.

src/redux/actions/cellSets/updateCellSetsClustering.js
aerlaut: `updateCellSetsClustering` is used in `configureEmbedding` to update cell sets when there is a reclustering. 

This is set to force reload to handle the case of multiple users, with different browsers, looking at the same experiment at the same time. If one of the users recluster the cell sets (i.e. via configure embedding), the cell sets in the other users' browsers will also be forced to update. Otherwise, they would be looking at different cell sets, which may mess up cell set operations.

If I don't set this to force reload, the cell sets will not update in the other users' browsers.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -159,11 +159,17 @@ const DotPlotPage = (props) => {
...
+    // If using custom genes, check that there are genes in the list
+    if (!config.useMarkerGenes && config.selectedGenes.length === 0) return;
+
+    // If using marker genes, check that the selected number is more than 0
+    if (config.useMarkerGenes && config.nMarkerGenes === 0) return;
 
     // Marker genes calculation needs that the cellIds in groupBy (refer to fn definition)
     // be represented by more than one groups in filterBy to enable comparison
-    if (config?.useMarkerGenes && !hasMoreThanTwoGroupsToCompare) {
+    if (config.useMarkerGenes && !hasMoreThanTwoGroupsToCompare) {
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: What is this change for? A bugfix on something else?

src/redux/actions/cellSets/runCellSetsClustering.js
cosa65: One option could be to make it so that the worker always sends the work finished notification to the socket for `Etag`, apart from hitting the `experimentUpdates` socket.

In this way we could still call `fetchWork` here so that we get the benefit of the timeout tracking on the client that triggered the clustering while still having clustering broadcast its result.

cosa65: To clarify: With this solution `updateCellSetsClustering` would basically just call `fetchWork` and if that results in error then send the redux error. 

It wouldn't do anything if the cellSets send the notification for success ,becuase that case is handled in `experimentUpdatesHandler`.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I made these changes when I was working on the tests and discovered actually`fetchPlotDataWork` was called multiple times when it's not supposed to. These checks prevent that.

src/redux/actions/cellSets/runCellSetsClustering.js
aerlaut: Yup sure, that can be done.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (ivababukova) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: could you add a structure to those checks by making them into a function named "shouldPlotDataBeFetched" or something like that, have those checks in this function, make the function return a boolean and call that function here?

src/redux/actions/cellSets/updateCellSetsClustering.js
ivababukova: so this is a bug that you are fixing with these changes?

src/redux/actions/cellSets/runCellSetsClustering.js
aerlaut: Done 👍 corresponding worker implementation in https://github.com/hms-dbmi-cellenics/worker/pull/253

src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx
Comment (aerlaut) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.test.jsx:76]: I think it can be worked on together with this ticket : https://biomage.atlassian.net/browse/BIOMAGE-1553. I've changed the ticket to a more exploratory one, that outlines how we should approach testing and factor out common functionalities.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: addressed in bd2d5c9

cosa65 (approved):

PR #621, opened 2022-01-27
[BIOMAGE-1584] Fix bugged embedding tooltip

StefanBabukov:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1584

https://ui-grasp6-ui621.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -5,10 +5,10 @@ import {
   useSelector, useDispatch,
 } from 'react-redux';
 import PropTypes from 'prop-types';
-import _ from 'lodash';
 import * as vega from 'vega';
 
 import 'vitessce/dist/es/production/static/css/index.css';
+import getCellClassProperties from 'utils/cellSets/getCellClassProperties';
```
cosa65: Minor nitpick, but could you group this import with the other utils import? As you can see all the other imports are grouped by components, redux actions, selectors ,etc

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -10,24 +10,27 @@ import _ from 'lodash';
  * @param {object} properties CellSets properties.
  */
 
-const getCellClassProperties = (cellId, cellSetClassKey, hierarchy, properties) => {
-  const childrenCellSets = _.find(hierarchy, ({ key }) => key === cellSetClassKey).children;
+const getCellClassProperties = (cellId, cellSetClassKey, cellSets) => {
```
cosa65: The behavior of this function changed ,so update the comment explaining what it does, also check out which part of this: "Returns the properties of the first of the cellSets that contains the cellId (by the order specified in hierarchy children)" should remain here and which shoudl go to the part where we now do `getCellClassProperties()[0]`

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -10,24 +10,27 @@ import _ from 'lodash';
...
 
-  childrenCellSets.every(({ key }) => {
-    if (properties[key].cellIds.has(cellId)) {
-      firstCellSetContainingCell = key;
-
-      // Break out of the loop
-      return false;
+  childrenCellSets.forEach(({ key }) => {
+    if (properties[key].cellIds.has(parseInt(cellId, 10))) {
+      cellSetsContainingCell.push(key);
```
cosa65: Why no keep the `return false` to break out of the loop? I feel like it is a little bit harder to follow now and understand that we are only getting the first cellSet that matches

cosa65: Or use `firstCellSetContainingCell` or something similar to keep making that evident, sicne it is a weirdness that we need to do, so let's make it as obvious a possible.

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -10,24 +10,27 @@ import _ from 'lodash';
...
     }
+  });
 
-    // Continue looping
-    return true;
+  const clusterProperties = cellSetsContainingCell.map((clusterKey) => {
+    const rootClusterName = hierarchy.filter(({ children }) => (
+      children.filter((child) => child.key === clusterKey).length > 0))[0].key;
+    return { ...properties[clusterKey], rootClusterName: _.capitalize(rootClusterName) };
```
cosa65: Why are we adding a property that doesn't exist in the original properties here?

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -165,7 +165,12 @@ const HeatmapPlot = (props) => {
   }, [louvainClusterCount]);
 
   useEffect(() => {
-    dispatch(updateCellInfo({ cellName: cellHighlight }));
+    if (cellHighlight) {
+      const containingClusters = getCellClassProperties(parseInt(cellHighlight, 10), ['louvain', 'scratchpad'], cellSets)
+        .map(({ name, rootClusterName }) => `${rootClusterName} : ${name}`);
+
+      dispatch(updateCellInfo({ cellName: cellHighlight, cellSets: containingClusters }));
```
cosa65: I am not sure I agree about this, I feel like the only information we need to share across components inside redux is which cell was highlighted, then each component can decide exactly what to display on their own tooltips.

For example in the embedding, when gene expressions for one particular gene is activated, it displays gene expression and other stuff, I feel like that doesn't need to be in redux, it can simply be stored in a `useRef` and recalculated when cellHighlight is updated.

I would probably not store `cellInfo` in redux, instead perhaps only `highlightedCell` or something like that.

src/utils/cellSets/getCellClassProperties.js
StefanBabukov: because the tooltip also includes the root cluster name e.g. 'Louvain: Cluster 0'

src/utils/cellSets/getCellClassProperties.js
StefanBabukov: we are not getting the first one that matches, instead in this PR i changed it to get every cluster that contains it. It is needed to display all the clusters in the tooltip that have it - could be multiple scratchpad ones and louvain.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
StefanBabukov: But when we hover on the heatmap we want the tooltip to appear on the embedding too so we need that cellInfo state to be checked in both components? How else would we update it?

StefanBabukov: We still need to save cellSets inside - when we hover on the heatmap we want the tooltip to appear on the embedding with the cluster information too. What are the drawbacks with the current version? You think its better to add more variables in the files and more logic instead of keeping everything in redux? I just dont like so many variables, used around, we will need to pass it also to other components can get confusing and slower to read

StefanBabukov: Each tooltip gets the things it needs from redux, ignoring the rest just like in many places across the platform (getting only hierarchy under cellSets and etc.), I don't see the problem here, maybe we can discuss?

src/utils/cellSets/getCellClassProperties.js
StefanBabukov: I changed it to include the whole thing instead of setting it in every component separately

src/utils/cellSets/getCellClassProperties.js
cosa65: Ok, I would double check with vicky about the tooltips showing all cellSets, in particular more than one per cellClass, but to me it actually makes a lot sense.

But if we go this way maybe one change I would consider to do would be to somehow be able to indicate which cellClass the preprties belong to. 
Like if I understood this correctly, right now we could end up returning `[{key: 'metadataCellSet1', ...}, {key: 'metadataCellSet3', ...}, {key: 'louvain-2', ...}]` with now way of knowing fast which of these cellSets is the one for louvain unless we go through each of them and check against `parentNodeKey` or something.

Perhaps we could return them grouped by `cellClass`? E.g.: `{metadata: [{key: 'metadataCellSet1', ...}, {key: 'metadataCellSet3', ...}, louvain: {}]`

src/utils/cellSets/getCellClassProperties.js
cosa65: I don't think we should add in this function custom properties that are used only for the tooltip.

I would extract that part of the code and do it when we are preparing the tooltip to display, not here.

This function says it returns the properties, not other properties that are useful for only some of the use cases of this function.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: > You think its better to add more variables in the files and more logic instead of keeping everything in redux?

The amount of logic is the same because we still need to calculate the same things, if it is well abstracted it's just a matter of having that code in the embedding instead of when dispatching to redux.


> Each tooltip gets the things it needs from redux, ignoring the rest just like in many places across the platform (getting only hierarchy under cellSets and etc.), I don't see the problem here, maybe we can discuss?

I would argue this isn't the same case as hierarchy, but we can dive into that later if you want.

I would encourage you to read these 2 points:
https://redux.js.org/style-guide/style-guide#keep-state-minimal-and-derive-additional-values
https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters

Apart from that, here I wrote down a more specific explanation on why I feel this pattern is better, regardless of whether redux suggests this or not.

**1:** The version where all the data each component needs gets sent to redux means that the following happens when cellId is updated in one of the components:
- on dispatch that component needs to calculate all of the extra data for each of the components which **might or might not be subscribed to the cellId updates**, so that logic is actually being executed whenever cellId changes.
- Each components picks up onlty what they need from redux.

**2:** The version where we only send to redux what we absolutely need:
- on dispatch the component sends [only the event specific data](https://redux.js.org/style-guide/style-guide#model-actions-as-events-not-setters): the cellId
- Each component listens to the cellId changing and processes that in whatever way they need to

The differences:
- With case **1** there is an unnecessary coupling between redux and each component that uses it, coupling is bad because:
    - if a component no longer needs something we need to adapt redux and that component and perhaps other components too, depending on how the code for calculating that outdated data that is dispatched works.
    - If a component is removed we need to adapt redux and the other components for similar reasons, while inthe other way we simply remove the component and be happy.

- With case **1**, if a component was closed by the user due to the remove panels functionality we have in data exploration, processing that data for display in the closed component is still going to be done even though we don't need to in reality.

- I think there are other reasons, but I think the redux style guide is pretty useful if you want more information too.

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -1,33 +1,29 @@
...
-
-  let firstCellSetContainingCell = null;
-
-  childrenCellSets.every(({ key }) => {
-    if (properties[key].cellIds.has(cellId)) {
-      firstCellSetContainingCell = key;
+const getCellClassProperties = (cellId, cellSetClassKeys, cellSets) => {
+  const { properties, hierarchy } = cellSets;
+  const cellClassProperties = {};
+  const cellIdInt = parseInt(cellId, 10);
```
cosa65: This `parseInt` shouldn't be done in here. Only vitessce gives us the ids as strings, it's better to just transform them to ints in vitessce's callbacks as soon as they are received instead of making many different functions have to deal with cellIds that may or may not be ints

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -1,33 +1,29 @@
...
 
-const getCellClassProperties = (cellId, cellSetClassKey, hierarchy, properties) => {
-  const childrenCellSets = _.find(hierarchy, ({ key }) => key === cellSetClassKey).children;
-
-  let firstCellSetContainingCell = null;
-
-  childrenCellSets.every(({ key }) => {
-    if (properties[key].cellIds.has(cellId)) {
-      firstCellSetContainingCell = key;
+const getCellClassProperties = (cellId, cellSetClassKeys, cellSets) => {
```
cosa65: Sorry to make you add this change since it doesn't have to do much with your ticket, but do you think this name fits well with with what the function does?

Perhaps `getContainingCellSetsProperties` fits better instead?

src/utils/cellSets/getCellClassProperties.js
```diff
@@ -1,33 +1,29 @@
...
-    }
+  cellSetClassKeys.forEach((rootCluster) => {
+    cellClassProperties[rootCluster] = [];
+    const childrenCellSets = hierarchy.filter(({ key }) => rootCluster === key)[0]?.children || [];
 
-    // Continue looping
-    return true;
+    childrenCellSets.forEach(({ key }) => {
+      if (properties[key].cellIds.has(cellIdInt)) {
+        cellClassProperties[rootCluster].push(properties[key]);
```
cosa65: `rootCluster` sounds louvain specific, I would go with `cellClass` (or `rootNode`)

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -165,7 +165,9 @@ const HeatmapPlot = (props) => {
   }, [louvainClusterCount]);
 
   useEffect(() => {
-    dispatch(updateCellInfo({ cellName: cellHighlight }));
+    if (cellHighlight) {
```
cosa65, line 168: What happens when `cellHighlight` turns to `null`? The tooltips for embedding and heatmap should disappear, does that happen if we have this `if (cellHighlight)`?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -337,6 +322,7 @@ const Embedding = (props) => {
                 <CellInfo
                   componentType={embeddingType}
                   coordinates={cellCoordintes}
+                  cellInfoRef={cellInfoTooltip.current}
```
cosa65: `cellInfoRef` should just be called `cellInfo`, you're not passing a `useRef`

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -132,23 +132,36 @@ const Embedding = (props) => {
...
+        });
+      });
 
-    if (cellSetProperties) {
-      setCellSetClusters(Object.entries(cellSetProperties).filter(([, cellSet]) => cellSet.type === 'cellSets'));
+      cellInfoTooltip.current = {
+        cellSets: prefixedCellSetNames,
+        cellName: selectedCell,
+        componentType: embeddingType,
+        ...extras,
```
cosa65: I would just pass `geneName` and `expression` here instead of wrapping it in `extras`

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -132,23 +132,36 @@ const Embedding = (props) => {
...
+      const rootClusterNodes = cellSetHierarchy.map(({ key }) => key)
+        .filter((key) => cellSetProperties[key].type === 'cellSets');
+
+      // getting the cluster properties for every cluster that has the cellId
+      const cellProperties = getCellClassProperties(selectedCell, rootClusterNodes, cellSets);
+
+      const prefixedCellSetNames = [];
+      Object.entries(cellProperties).forEach(([key, clusterProperties]) => {
+        clusterProperties.forEach(({ name, parentNodeKey }) => {
+          prefixedCellSetNames.push(`${_.capitalize(parentNodeKey)} : ${name}`);
```
cosa65: Using `capitalize` seems a little bit outdated. We have names for the parent nodes, so I would use that name from `properties` instead.

An example: With `capitalize` we will show `"Scratchpad"` which doesn't make sense to a user because it isn't called scratchpad anymore

src/components/data-exploration/heatmap/HeatmapPlot.jsx
StefanBabukov: yes, we check if cellHighlight is defined before rendering the heatmap cell info tooltip

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -60,22 +60,22 @@ const Embedding = (props) => {
 
   const focusData = useSelector((state) => state.cellInfo.focus);
   const focusedExpression = useSelector((state) => state.genes.expression.data[focusData.key]);
+  const cellSets = useSelector(getCellSets());
   const {
     properties: cellSetProperties,
     hierarchy: cellSetHierarchy,
     hidden: cellSetHidden,
-  } = useSelector(getCellSets());
+  } = cellSets;
```
aerlaut, line 68: Why not use the old style?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -60,22 +60,22 @@ const Embedding = (props) => {
...
 
   const selectedCell = useSelector((state) => state.cellInfo.cellName);
```
aerlaut: `cellName` is actually `cellId` right? Is it possible to change `cellName` to `cellId` it so that's it's easier to understand?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -132,23 +132,33 @@ const Embedding = (props) => {
...
-
-    if (cellSetProperties) {
-      setCellSetClusters(Object.entries(cellSetProperties).filter(([, cellSet]) => cellSet.type === 'cellSets'));
+    if (selectedCell) {
+      const expressionToDispatch = focusedExpression
+        ? focusedExpression.rawExpression.expression[selectedCell] : undefined;
+
+      // getting the root nodes which are of type cellSets
+      const rootClusterNodes = cellSetHierarchy.map(({ key }) => key)
+        .filter((key) => cellSetProperties[key].type === 'cellSets');
```
aerlaut: Can you use the selector `getCellSetsHierarchyByKeys` instead of this?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -132,23 +132,33 @@ const Embedding = (props) => {
...
+      const prefixedCellSetNames = [];
+      Object.entries(cellProperties).forEach(([key, clusterProperties]) => {
+        clusterProperties.forEach(({ name, parentNodeKey }) => {
+          prefixedCellSetNames.push(`${cellSetProperties[parentNodeKey].name} : ${name}`);
+        });
+      });
+
+      cellInfoTooltip.current = {
+        cellSets: prefixedCellSetNames,
+        cellName: selectedCell,
```
aerlaut: I think it would be good to change all occurrences of `cellName` to `cellId`. Is it possible to do that?

src/components/data-exploration/embedding/Embedding.jsx
StefanBabukov: Because I need to pass the whole cellSets to another function

src/utils/cellSets/getContainingCellSetsProperties.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * Returns the properties of the cellSets that contains the cellId
+ * (by the order specified in hierarchy children)
+ * @param {int} cellId The id of the cell we are searching for.
+ * @param {array} cellSetClassKeys The cellSetClass we are searching inside of
+ *  (e.g.: sample, louvain, scratchpad, etc..).
+ * @param {object} cellSets  cellSets for the experiment.
+ */
```
aerlaut, line 10: Great job in adding the function documentation! Can I ask you to complete it by adding a `* @return {[type]}      [description]`?

src/components/data-exploration/embedding/Embedding.jsx
StefanBabukov: `getCellSetsHierarchyByKeys` searches by keys, instead i need to get all the clusters that have `type: cellSets` (in this case louvain and scratchpad) for the tooltip

src/utils/cellSets/getContainingCellSetsProperties.js
StefanBabukov: Copilot actually generated the comment btw :D

src/__test__/utils/cellSets/getContainingCellSetsProperties.test.js
```diff
@@ -0,0 +1,13 @@
...
+import mockCellSets from 'utils/tests/mockStores/cellSets';
+
+describe('get cell class properties tests', () => {
+  const { cellSets } = mockCellSets();
+  it('returns information for a passed cellId', () => {
+    // doing it as a string to make sure it still works (sometimes a string is passed from vitessce)
+    const cellId = '3';
+    const cellSetClassKeys = ['test'];
+    const result = getContainingCellSetsProperties(cellId, cellSetClassKeys, cellSets);
+    expect(result).toMatchSnapshot();
```
aerlaut: Can you use a `cellId` that is contained in one of the cell sets? From the snapshot, The returned value doesn't contain anything.

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,25 +1,18 @@
...
-    const store = mockStore({
-      cellInfo: {
-        cellName: 1904,
-        componentType: 'heatmap',
-        expression: 0,
-        geneName: 'DOK3',
-      },
-    });
-
     const coordinates = {
```
aerlaut: Can you take out these `coordinates` and put them out of the test blocks so it can be defined once and used by all the tests?

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,25 +1,18 @@
 import React from 'react';
-import { Provider } from 'react-redux';
 import { mount } from 'enzyme';
-import configureMockStore from 'redux-mock-store';
-import thunk from 'redux-thunk';
 import { Card } from 'antd';
 import CellInfo from '../../../components/data-exploration/CellInfo';
 import '__test__/test-utils/setupTests';
```
aerlaut: Do you need to import this file? If not, then I think it can be removed.

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,25 +1,18 @@
 import React from 'react';
-import { Provider } from 'react-redux';
 import { mount } from 'enzyme';
-import configureMockStore from 'redux-mock-store';
-import thunk from 'redux-thunk';
 import { Card } from 'antd';
 import CellInfo from '../../../components/data-exploration/CellInfo';
```
aerlaut: Can you change this path into an absolute path?

src/components/data-exploration/CellInfo.jsx
```diff
@@ -1,19 +1,15 @@
...
 
-  const cellInfo = useSelector((state) => state.cellInfo);
   const [cellInfoVisible, setCellInfoVisible] = useState(false);
 
   useEffect(() => {
-    if (!cellInfo.cellName) {
+    if (!cellInfo) {
       return;
     }
     if (cellInfo.componentType !== componentType) {
```
aerlaut: Can you check if we need `componentType`? The value of `componentType` in `Embedding.jsx` is derived from `embedding.method`, which is the method of embedding chosen from configure embedding. The same value is also passed in `cellInfo`, so we don't seem to need this check

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,25 +1,18 @@
...
-const mockStore = configureMockStore([thunk]);
+const cellInfo = {
+  cellId: 1904,
+  componentType: 'heatmap',
+  expression: 0,
+  geneName: 'DOK3',
+};
 
 describe('CellInfo', () => {
   test('renders correctly when hovering over the same component', () => {
```
aerlaut: If we don't need the `componentType` param and component check, this test is not needed

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -30,24 +23,13 @@ describe('CellInfo', () => {
...
-      <Provider store={store}>
-        <CellInfo componentType='heatmap' coordinates={coordinates} />
-      </Provider>,
+      <CellInfo componentType='heatmap' coordinates={coordinates} cellInfo={cellInfo} />,
     );
 
     expect(component.find(Card).length).toEqual(1);
   });
 
   test('does not show when hovering over different component', () => {
```
aerlaut: If we don't need the `componentType` param and component check, this test is not needed

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -81,9 +57,7 @@ describe('CellInfo', () => {
     };
 
     const component = mount(
-      <Provider store={store}>
-        <CellInfo componentType='heatmap' coordinates={coordinates} />
-      </Provider>,
+      <CellInfo componentType='heatmap' coordinates={coordinates} cellInfo={{}} />,
     );
 
     expect(component.find(Card).length).toEqual(0);
```
aerlaut: Instead of doing this check, which only checks for the existence of the component, I think it would be better to check for the content of `cellInfo`, i.e. are all the information we are expecting available? It would be much easier to use react-testing-library then instead of enzyme.

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,25 +1,18 @@
 import React from 'react';
-import { Provider } from 'react-redux';
 import { mount } from 'enzyme';
```
aerlaut: Can we use RTL instead of enzyme for this test?

src/__test__/redux/reducers/cellInfoReducer.test.js
```diff
@@ -4,30 +4,6 @@ import {
   CELL_INFO_UPDATE, CELL_INFO_FOCUS, CELL_INFO_UNFOCUS,
 } from '../../../redux/actionTypes/cellInfo';
```
aerlaut: While you're editing this file, can you please change these references to use absolute path?

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: Can you use `getCellSetsHierarchyByType` then?

src/__test__/components/data-exploration/embedding/CrossHair.test.jsx
```diff
@@ -11,7 +11,7 @@ describe('CrossHair', () => {
   test('renders correctly', () => {
     const store = mockStore({
       cellInfo: {
-        cellName: 1904,
+        cellId: 1904,
```
aerlaut, line 14: While you're making this change, can you also fix the relative import paths to use absolute import paths in this file?

src/__test__/components/data-exploration/embedding/Embedding.test.jsx
```diff
@@ -63,7 +63,7 @@ describe('Embedding', () => {
       },
     },
     cellInfo: {
-      cellName: 2,
+      cellId: 2,
```
aerlaut, line 66: While you're making this change, can you also fix the relative import paths to use absolute import paths in this file?

src/__test__/redux/actions/cellInfo/updateCellInfo.test.js
```diff
@@ -10,7 +10,7 @@ const mockStore = configureStore([thunk]);
 
 describe('updateCellInfo action', () => {
   const updatePayload = {
-    cellName: 'C1',
+    cellId: 'C1',
```
aerlaut, line 13: While you're making this change, can you also fix the relative import paths to use absolute import paths in this file?

src/__test__/utils/cellSets/getContainingCellSetsProperties.test.js
StefanBabukov: oh yeah it had values before i must've broken it

src/utils/cellSets/getContainingCellSetsProperties.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * Returns the properties of the cellSets that contains the cellId
+ * (by the order specified in hierarchy children)
+ * @param {int} cellId The id of the cell we are searching for.
+ * @param {array} cellSetClassKeys The cellSetClass we are searching inside of
```
aerlaut, line 5: Btw, why do you set this parameter as an array? Why not just a string? Is there a case where we want to specify more than one `cellSetClassKey`?

src/components/data-exploration/CellInfo.jsx
StefanBabukov: nice catch!

src/utils/cellSets/getContainingCellSetsProperties.js
StefanBabukov: Yeah in the embedding we want to display all the cluster names that have the cellId - it can be both scratchpad and louvain

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,65 +1,30 @@
...
 
-    expect(component.find(Card).length).toEqual(0);
+describe('CellInfo', () => {
+  it('renders cell info card with properties', () => {
+    render(<CellInfo coordinates={coordinates} cellInfo={cellInfo} />);
+    expect(screen.getByText('Gene name: DOK3')).toBeInTheDocument();
+    expect(screen.getByText('Cell id: 1904')).toBeInTheDocument();
+    expect(screen.getByText('Expression Level: 0')).toBeInTheDocument();
+    expect(screen.getByText('Louvain : cluster1')).toBeInTheDocument();
+    expect(screen.getByText('anotherRootCluster : cluster2')).toBeInTheDocument();
```
aerlaut: Sorry for being picky, but can you replace the hard coded values here to use the values in the variable? e.g. `Gene name: ${cellInfo.geneName}`)

src/__test__/utils/cellSets/getContainingCellSetsProperties.test.js
```diff
@@ -4,10 +4,10 @@ import mockCellSets from 'utils/tests/mockStores/cellSets';
 describe('get cell class properties tests', () => {
   const { cellSets } = mockCellSets();
   it('returns information for a passed cellId', () => {
-    // doing it as a string to make sure it still works (sometimes a string is passed from vitessce)
```
aerlaut: So having `cellId` as a string doesn't work?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -157,6 +157,7 @@ const Embedding = (props) => {
         expression: expressionToDispatch,
         geneName: focusData?.key,
       };
+      console.log('SETTING THAT TOOLTIP ', cellInfoTooltip.current);
```
aerlaut: Don't forget to `UNSET THAT TOOLTIP`

aerlaut: Jk, don't forget to remove this `console.log` :D

src/components/data-exploration/CellInfo.jsx
```diff
@@ -57,7 +44,7 @@ const CellInfo = (props) => {
     </Card>
   );
 
-  if (cellInfoVisible && cellInfo.cellId && Object.keys(coordinates.current).length > 0) {
+  if (cellInfo.cellId && Object.keys(coordinates.current).length > 0) {
```
aerlaut: Is there a better way to check if `coordinates` exists instead of checking for `Object.keys(coordinates.current).length`? Does the `coordinates.current` not contain anything when we're not hovering?

src/__test__/components/data-exploration/CellInfo.test.jsx
```diff
@@ -1,65 +1,30 @@
...
+  current: {
+    x: 100,
+    y: 200,
+    width: 500,
+    height: 500,
+  },
+};
 
-    expect(component.find(Card).length).toEqual(0);
+describe('CellInfo', () => {
```
aerlaut, line 21: Can you add tests for the cases where : 
- CellInfo does not display anything when there is no cell info
- CellInfo does not display anything if coordinate is not given (e.g. coordinates.current is not set?)

src/__test__/utils/cellSets/getContainingCellSetsProperties.test.js
StefanBabukov: before the function would convert it to int - now its changed to expect an int so this string - int conversion is done only in the places that receive string outside the function

src/__test__/components/data-exploration/CellInfo.test.jsx
StefanBabukov: cellInfo is not called if there is no cell info - check Embedding.jsx

StefanBabukov: I added a check there

src/components/data-exploration/CellInfo.jsx
StefanBabukov: I removed the whole check because we don't need it anymore

src/components/data-exploration/CellInfo.jsx
```diff
@@ -37,9 +20,9 @@ const CellInfo = (props) => {
...
       }}
     >
-      {cellInfo.cellName ? (
+      {cellInfo.cellId ? (
         <div style={cellInfoStyle}>
-          {`Cell id: ${cellInfo.cellName}`}
+          {`Cell id: ${cellInfo.cellId}`}
         </div>
       ) : <></>}
       {cellInfo.geneName ? (
```
aerlaut: Sorry, just had another look. Can you see if we need these `cellInfo.cellId`, `cellInfo.expression`, `cellInfo....` checks? Are there cases where cellInfo exists but some of these properties do not / partially exist? Otherwise I don't think we need these checks,

StefanBabukov: Yeah sometimes we have  partial ones that's why i kept it. I guess cellId should always be there so i can remove that check

aerlaut (approved):

PR #626, opened 2022-02-04
Refactor data management loader

aerlaut:
# Description
Improve data management loader

- Move the loader into the `PageHeader` component so that it can be used on other pages if needed.
- Resize the loader element so that it does not cause layout shifts in the data management page (jerky Data Management panel) because of the loader appearing and disappearing.
- Add some padding to the icon of the "Create Project" button

(if you can not play the video below, see the video on this thread : https://this-is-biomage.slack.com/archives/C015CHSUDRU/p1643977867643889?thread_ts=1643977200.612799&cid=C015CHSUDRU)

https://user-images.githubusercontent.com/2862362/152527645-202b307d-11db-4658-b9d3-b241557e3de3.mov

# Details
#### URL to issue
N/A

N/A

#### Links to any PRs or resources
N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/__test__/components/Header.test.jsx
+++ b/src/__test__/components/Header.test.jsx
@@ -41,9 +41,11 @@ const renderHeader = async (store, props = {}) => render(
   </Provider>,
 );
 
+const loaderClass = "span[class='css-zog0u6']";
+
 describe('Page Header', () => {
-  it('renders properly', async () => {
-    await renderHeader(storeState);
+  it('Renders properly', async () => {
+    const { container } = await renderHeader(storeState);
 
     // It contains the page title
     expect(screen.getByText(pageTitle));
@@ -56,12 +58,25 @@ describe('Page Header', () => {
 
     // It contains the user button
     expect(screen.getByText(/User Button/));
+
+    // It should not have the loader by default
+    const loader = container.querySelector(loaderClass);
+    expect(loader).not.toBeInTheDocument();
   });
 
-  it('shows the loader bar if loading', async () => {
-    const { container } = await renderHeader(storeState);
+  it('Shows the loader bar if loading', async () => {
+    const { container } = await renderHeader(storeState, { loading: true });
 
     // It contains the loader SVG
-    expect(container.querySelector("span[class='css-zog0u6']")).not.toBeUndefined();
+    const loader = container.querySelector("span[class='css-zog0u6']");
+    expect(loader).toBeInTheDocument();
+  });
+
+  it('Should render extra elements passed into it', async () => {
+    const extraText = 'Extra element';
+    await renderHeader(storeState, { extra: <>{extraText}</> });
+
+    // It contains the loader SVG
+    expect(screen.getByText(extraText)).toBeInTheDocument();
   });
 });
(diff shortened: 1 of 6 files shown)


src/components/Header.jsx
```diff
@@ -51,12 +59,14 @@ Header.propTypes = {
...
   title: PropTypes.string.isRequired,
   extra: PropTypes.node,
+  loader: PropTypes.bool,
 };
 
 Header.defaultProps = {
   experimentId: null,
   experimentData: null,
   extra: <></>,
+  loader: false,
```
ivababukova: the name `loader` doesn't sound like a boolean. I suggest you rename it to `isLoading` or `loading` instead

src/components/Header.jsx
```diff
@@ -5,14 +5,15 @@ import {
...
+import UserButton from 'components/UserButton';
+import FeedbackButton from 'components/FeedbackButton';
+import ReferralButton from 'components/ReferralButton';
+import integrationTestConstants from 'utils/integrationTestConstants';
+import { ClipLoader } from 'react-spinners';
 
 const Header = (props) => {
   const {
-    experimentId, experimentData, title, extra,
+    experimentId, experimentData, title, extra, loader,
```
ivababukova: Are there any other components apart from `src/pages/data-management/index.jsx` that would benefit from using `loader`?

ivababukova: Here are all the usages of this component I could see:
Screenshot 2022-02-04 at 13 37 19

aerlaut: No, right now I don't think there are other pages that require loader. Usually each component has their own loader.

ivababukova: if the loader is not used and won't be used, then why adding it as part for the `Header` component?

src/components/Header.jsx
aerlaut: changed

src/components/Header.jsx
aerlaut: The consideration to put it here is that it's technically easier to add it to the header and it's clean. We can still keep it on Data Management, but I'd have to add a considerable amount of HTML and CSS to make it look nice and not jerk around. It also opens up an option for the loader to be used by other components in the UI.

src/__test__/components/Header.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+const storeState = mockStore({
+  ...initialExperimentState,
+  experiments: {
+    [mockExperimentId]: {
+      ...experimentTemplate,
+      name: mockExperimentName,
+    },
+  },
+});
```
ivababukova, line 28: I see this is a brand-new test. Why writing it with the fake store instead of using the real store directly?

aerlaut:
This small improvement seems to increase the page bundle size for quite a large amount. The feature is not even used by most of the pages that are affected. I think the improvement that I want to make here is better to be implemented via another approach. The value of the improvement vs page size increase brought by the current approach is not worth making.

PR #638, opened 2022-02-21
[BIOMAGE-1745] - Show error for failed sample uploads

aerlaut:
# Description

This PR and the accompanying API PR adds several fixes to make sample creation and upload more robust. Specifically this PR:
- Stops sample upload if the API fails to create a sample entry.
- Delegate updating the sampleIds in the `experiments` table to the API, so insertion of new sample id to the `experiments` table is done together when the sample id is inserted into the `samples` table in the database. Previously this was done before by updating Redux store first, and then saving the state. 
- Shows error information as the sample name if a sample fails to upload. This usually cause the UI to fail. To test for this case, log into the staging link using the admin account and look for experiment `fe96449d41b0bac2a9c48dbf9f10db8d`

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1745

https://ui-agi-ui638-api305.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/api/pull/305

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
   },
 ];
 
+// If a sample exists in projects but not in the samples file,
+// there is an error uploading the sample. We should show an error for that sample.
+const sampleUploadErrorTemplate = {
+  ...sampleTemplate,
+  name: 'UPLOAD ERROR: Please reupload sample',
```
ivababukova: I would still display the real same name here, because it will help the user identify quicker which sample is not uploaded correctly.

aerlaut: We can't get the real name, because the sample record is not there

aerlaut: I'll try to see how we can get the real name there,

ivababukova: If you want to make it more visible that there was an upload error, you can rephrase the message in `LaunchAnalysisButton` to say "Ensure that all samples are uploaded successfully and all relevant metadata is inserted."

src/__test__/components/data-management/SamplesTable.test.jsx
```diff
@@ -103,45 +108,47 @@ describe('Samples table', () => {
...
+    await renderSamplesTable(storeState);
 
     Object.values(samples).forEach((sample) => {
       expect(screen.getByText(sample.name)).toBeInTheDocument();
     });
   });
 
+  it('Should show an error if a sample fails to upload', async () => {
+    const missingSampleState = _.cloneDeep(storeState.getState());
+    const createMockStore = configureMockStore([thunk]);
```
aerlaut, line 127: I had to use the mock store to replicate the condition where the sample id exists in `projects` but not `samples` as there is no easy way to replicate that state using the real sore.

src/components/data-management/SamplesTable.jsx
aerlaut: Message in `LaunchAnalaysisButton` changed accordingly.

src/redux/actions/samples/createSample.js
```diff
@@ -88,10 +84,8 @@ const createSample = (
 
     pushNotificationMessage('error', message);
 
-    return Promise.reject(message);
```
aerlaut, line 91: This does not throw an error, so the subsequent line in the enclosing function continues to run

src/components/data-management/LaunchAnalysisButton.jsx
```diff
@@ -130,7 +132,7 @@ const LaunchAnalysisButton = () => {
     if (!canLaunchAnalysis()) {
       return (
         <Tooltip
-          title='Ensure all samples are uploaded and all metadata are inserted'
+          title='Ensure that all samples are uploaded successfully and all relevant metadata inserted.'
```
ivababukova: and all relevant metadata **is** inserted

src/components/data-management/SamplesTable.jsx
```diff
@@ -245,13 +254,13 @@ const SamplesTable = forwardRef((props, ref) => {
...
         key: idx,
-        name: samples[sampleUuid].name,
+        name: sampleData.name,
         uuid: sampleUuid,
         barcodes: barcodesData,
         genes: genesData,
         matrix: matrixData,
-        species: samples[sampleUuid].species || DEFAULT_NA,
-        ...samples[sampleUuid].metadata,
+        species: sampleData.species || DEFAULT_NA,
```
ivababukova: do we still use this species here?

src/components/data-management/SamplesTable.jsx
```diff
@@ -221,19 +235,14 @@ const SamplesTable = forwardRef((props, ref) => {
...
-    // Set table data
     const newData = activeProject.samples.map((sampleUuid, idx) => {
-      // upload problems sometimes lead to partial updates and incosistent states
-      // in this situation it's possible that the sampleUuid does not exist
-      // this a temporary fix so that the whole UI doesn't crash preventing the
-      // user from removing the dataset or uploading another one.
-      // this situation should be included into the SamplesTable.jsx tests
-      const sampleFiles = samples[sampleUuid]?.files || {};
+      const sampleData = samples[sampleUuid] || sampleUploadErrorTemplate;
+      const sampleFiles = sampleData.files;
```
ivababukova: if `sampleData` is equal to `sampleUploadErrorTemplate`, there will be no `files` in it. There will be `sampleFiles` instead

src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
   },
 ];
 
+// If a sample exists in projects but not in the samples file,
+// there is an error uploading the sample. We should show an error for that sample.
+const sampleUploadErrorTemplate = {
+  ...sampleTemplate,
```
ivababukova: what else is inside `sampleTemplate`?

src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
...
+  name: 'UPLOAD ERROR: Please reupload sample',
+  sampleFiles: {
```
ivababukova: why adding these under `sampleFiles`, when it looks like code will be reading them from `files`?

src/redux/actions/samples/createSample.js
```diff
@@ -66,17 +62,17 @@ const createSample = (
 
     throwIfRequestFailed(response, json, endUserMessages.ERROR_SAVING);
 
-    dispatch({
+    await dispatch({
       type: SAMPLES_CREATE,
       payload: { sample: newSample, experimentId },
     });
 
-    dispatch(saveExperiment(experimentId));
```
ivababukova, line 74: where did `saveExperiment` go?

src/utils/upload/processUpload.js
```diff
@@ -209,17 +209,28 @@ const processUpload = async (filesList, sampleType, samples, activeProjectUuid,
   }, {});
 
   Object.entries(samplesMap).forEach(async ([name, sample]) => {
-    // Create sample if not exists
-    sample.uuid ??= await dispatch(createSample(activeProjectUuid, name, sampleType));
+    // Create sample if not exists.
+    try {
+      sample.uuid ??= await dispatch(createSample(activeProjectUuid, name, sampleType));
+    } catch (e) {
+      // If sample creation fails, sample should not be created
```
ivababukova, line 216: log the error

ivababukova: you will want to return the error to the user, no?

src/components/data-management/SamplesTable.jsx
```diff
@@ -2,6 +2,7 @@
 import React, {
   useEffect, useState, forwardRef, useImperativeHandle,
 } from 'react';
+import Storage from '@aws-amplify/storage';
```
cosa65, line 5: Why are we importing storage? We replaced it with the signed urls and axios.

src/components/data-management/LaunchAnalysisButton.jsx
```diff
@@ -113,6 +113,8 @@ const LaunchAnalysisButton = () => {
     };
 
     const canLaunch = activeProject.samples.every((sampleUuid) => {
+      if (!samples[sampleUuid]) return false;
```
cosa65, line 116: In which cases is there a sampleId in activeProject but it doesn't show up in samples?

src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
   },
 ];
 
+// If a sample exists in projects but not in the samples file,
+// there is an error uploading the sample. We should show an error for that sample.
+const sampleUploadErrorTemplate = {
```
cosa65: We could just have this inside the initialState in samples as well

cosa65: Also I would call it `sampleCreateErrorTemplate` to be more specific.

src/components/data-management/SamplesTable.jsx
```diff
@@ -221,19 +235,14 @@ const SamplesTable = forwardRef((props, ref) => {
   };
 
   useEffect(() => {
-    if (!samples[activeProject.samples[0]]) {
+    if (activeProject.samples.length === 0) {
```
cosa65, line 225: This if is checking for something differnet than it was before, is that ok? 
Before: the first sample has sample data in redux.
After: the active project has sample ids

src/redux/actions/samples/createSample.js
cosa65: I think we don't need it anymore because now the sample creation endpoint is smarter and also updates the experiment in dynamo.

src/redux/actions/samples/createSample.js
```diff
@@ -66,17 +62,17 @@ const createSample = (
...
+    return newSampleUuid;
   } catch (e) {
     let { message } = e;
     if (!isServerError(e)) {
       console.error(`fetch ${url} error ${message}`);
-      message = endUserMessages.ERROR_SAVING;
+      message = `${endUserMessages.ERROR_SAVING}. Please reupload sample '${name}'`;
```
cosa65: Does it make sense to perhaps move this whole message to a new `endUserMessages` element called `ERROR_SAVING_SAMPLE_CREATION` or sometihng liek that?

src/utils/upload/processUpload.js
```diff
@@ -209,17 +209,28 @@ const processUpload = async (filesList, sampleType, samples, activeProjectUuid,
   }, {});
 
   Object.entries(samplesMap).forEach(async ([name, sample]) => {
-    // Create sample if not exists
-    sample.uuid ??= await dispatch(createSample(activeProjectUuid, name, sampleType));
+    // Create sample if not exists.
+    try {
+      sample.uuid ??= await dispatch(createSample(activeProjectUuid, name, sampleType));
+    } catch (e) {
```
cosa65, line 215: What happens if we don't catch the error here?

src/components/data-management/SamplesTable.jsx
aerlaut: Oh right, we've removed species. I'll remove this.

src/utils/upload/processUpload.js
aerlaut: It's already notified to the user by a `pushNotification` inside `createSample`.

src/redux/actions/samples/createSample.js
aerlaut: Hmmm... sure

src/components/data-management/LaunchAnalysisButton.jsx
aerlaut: Thanks for catching that :)

src/components/data-management/SamplesTable.jsx
aerlaut: yes, my bad. Fixed.

src/components/data-management/SamplesTable.jsx
aerlaut: We still use this to download the public dataset directly from S3:
https://github.com/hms-dbmi-cellenics/ui/blob/d120f71a700aa61c74d6bdc471a946c90126660f/src/components/data-management/SamplesTable.jsx#L268

src/components/data-management/SamplesTable.jsx
aerlaut:
```
  const sampleTemplate = {
    name: null,
    projectUuid: null,
    uuid: null,
    type: null,
    createdDate: null,
    lastModified: null,
    complete: false,
    error: false,
    fileNames: [],
    files: {},
    metadata: {},
  };
```

src/components/data-management/SamplesTable.jsx
aerlaut: Yes, it should be under files, my bad. I'll change it.

src/components/data-management/SamplesTable.jsx
aerlaut: I think this will only be used in this component, nowhere else, so I don't think putting it inside initial state is proper. Besides that,the values inside `files` will later on have to be generated dynamically depending on the technology chosen for the file upload once we support other techonlogies (e.g. BD Rhapsody)

src/components/data-management/SamplesTable.jsx
aerlaut: The previous if statement checks if the first `sampleUuid` in `projects.samples` exists in `samples`. That first `sampleUuid` might not exist in `samples`, but the remaining `sampleUuid`s might exist, and we should still show those samples. This if statement checks if there really is no sample registered to the project, and if that's the case, show an empty table.

src/utils/upload/processUpload.js
aerlaut: This function continues, and in the current implementation, the samples are still uploaded to S3, even though there is no record of it in DynamoDB.

src/components/data-management/LaunchAnalysisButton.jsx
aerlaut: In the case of experimentId `fe96449d41b0bac2a9c48dbf9f10db8d`. If you compare the requests that are sent to update the samples record here: https://biomage.atlassian.net/browse/BIOMAGE-1745?focusedCommentId=10813, you can see that in the later update, a `sampleId` is missing, although the ID exists in the `project.samples`, resulting in a condition where the `sampleUuid` exists in `project.samples` but not in `samples`.

src/redux/actions/samples/createSample.js
ivababukova: makes sense, thanks for clarifying

src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
   },
 ];
 
+// If a sample exists in projects but not in the samples file,
+// there is an error uploading the sample. We should show an error for that sample.
+const sampleCreateErrorTemplate = {
+  ...sampleTemplate,
+  name: 'UPLOAD ERROR: Please reupload sample',
```
ivababukova: now that you've made the experiments table add the sample name, you can display it directly rather and remove this `name`

src/components/data-management/SamplesTable.jsx
aerlaut: We might want a new endpoint for this, one without authentication, to download the public datasets. Should I create a new ticket?

src/components/data-management/SamplesTable.jsx
```diff
@@ -48,6 +50,18 @@ const exampleDatasets = [
   },
 ];
 
+// If a sample exists in projects but not in the samples file,
+// there is an error uploading the sample. We should show an error for that sample.
+const sampleCreateErrorTemplate = {
```
ivababukova: do you really need to create `sampleCreateErrorTemplate`? If I look at the code below on lines 243-262, you only really use `sampleCreateErrorTemplate` for:
1. Getting the different name, which will be fixed, because now with saving samples to experiments table (your api PR), you can just display the sample and `samples[sampleUuid]` will be defined.
2.  Displaying the `{ upload: { status: UploadStatus.FILE_NOT_FOUND } },` for the files. But you don't need these fields, because that will be handled in lines 246, 247 of the code.

So you can remove the `sampleCreateErrorTemplate`  and bringing back these changes:
```
      // upload problems sometimes lead to partial updates and incosistent states
      // in this situation it's possible that the sampleUuid does not exist
      // this a temporary fix so that the whole UI doesn't crash preventing the
      // user from removing the dataset or uploading another one.
      // this situation should be included into the SamplesTable.jsx tests
      const sampleFiles = samples[sampleUuid]?.files || {};
```

src/components/data-management/SamplesTable.jsx
ivababukova: I am not sure I understand -- why do you need to change this? What was not working?

src/redux/actions/samples/saveSamples.js
```diff
@@ -31,10 +31,7 @@ const saveSamples = (
     });
 
     payload = projectSamples;
-    payload = {
-      ...payload,
-      [newSample.uuid]: newSampleToUpload,
-    };
+    payload[newSample.uuid] = newSampleToUpload;
```
ivababukova, line 34: why removing the `...payload`?

src/utils/endUserMessages.js
```diff
@@ -18,6 +18,7 @@ export default {
   ERROR_FETCHING_BACKEND_STATUS: 'We couldn\'t get the status of processing the experiment\'s processing run.',
   ERROR_NOT_SIGNED_IN: 'You need to sign in first. Click Sign in on the top right to begin.',
   ERROR_SAVING: 'We couldn\'t save your data.',
+  ERROR_CREATING_SAMPLE: 'We couldn\'t save your data. Please reupload sample:',
```
ivababukova: Isn't it more correct to say "We couldn't create your sample. Please try uploading it again"?

src/utils/upload/processUpload.js
ivababukova: by logging the error, I meant for us -- we want to see it in the console when it happens for debugging

src/__test__/redux/actions/samples/createSample.test.js
```diff
@@ -94,21 +88,23 @@ describe('createSample action', () => {
...
     expect(actions[1].type).toEqual(SAMPLES_CREATE);
 
-    // Calls update experiment on success of fetch
-    expect(saveExperiment).toHaveBeenCalledWith(experimentId);
+    // Returns a new sampleUuid
+    expect(newUuid).toEqual(sampleUuid);
   });
 
-  it('Shows error message when there was a fetch error', async () => {
+  it('Shows error message and throw an error when there is a fetch error', async () => {
```
ivababukova: and throw**s** an error

src/components/data-management/SamplesTable.jsx
aerlaut: I discovered this when I was writing a test. In the test, I have a `projects` redux slice with 3 samples `A`, `B`, `C`. The `samples` redux slice initiallly have 3 samples too : A, B and C. In the test, I deleted A from the `samples` redux slice. This means that `samples` now only have data for 2 samples : `B` and `C`. In this case, we would still like to show the samples `B` and `C` in the table and show `A` as missing. However, with the previous if condition, the table returns empty, because it specifically checks for the existence of the first sample of the `projects` in `samples`, not if the projects have any samples.

src/components/data-management/SamplesTable.jsx
aerlaut: removed

src/components/data-management/SamplesTable.jsx
aerlaut: The sample name will only be defined for new samples. Old samples will still throw an error. Would that be ok?

aerlaut: I'll implement a simpler default to fill in the data, to cover the case for these old experiments without having to create `sampleCreateErrorTemplate`

src/redux/actions/samples/saveSamples.js
aerlaut: These lines insert the new sample uuid and sample object to the payload. In the previous implementation we create a new object by spreading the payload object and inserting the key. In this new implementation, we just insert the sample record to the payload object with a new key, without having to create a new object. So this new implementation is more efficient.

src/utils/endUserMessages.js
aerlaut: Yes, I think it's more accurate. I'll change it

src/__test__/redux/actions/samples/createSample.test.js
aerlaut: fixed 👍

src/components/data-management/SamplesTable.jsx
aerlaut: Resolved the variable is deleted

src/utils/upload/processUpload.js
aerlaut: Ah ok, I'll add the log inside `createSample`

aerlaut: Oh we already log it in these lines inside `createSample` : https://github.com/hms-dbmi-cellenics/ui/blob/7127b0dc9dcc9684807b63c61d6d4c818d55146e/src/redux/actions/samples/createSample.js#L72-L76

However, it logs the error only if it's not a server error. I don't know why this was implemented... maybe because if it's a server error (e.g. API error), we'd already have a log of it and not show it in the browser? Should we log this anyway?

src/components/data-management/SamplesTable.jsx
ivababukova: why would you want to download the datasets without authentication?

src/components/data-management/SamplesTable.jsx
ivababukova: `Old samples will still throw an error. Would that be ok?`I don't understand this, can you expand? Can't you just display this error name `'UPLOAD ERROR: Please reupload sample'` for sample name for the old samples that don't have a name?

src/components/data-management/SamplesTable.jsx
ivababukova: it is clear for me now, thank you for the example and explanation!

src/utils/upload/processUpload.js
ivababukova: can you not just do console.error(e) in this current piece of code? When debugging, we want to see the actual error

src/components/data-management/SamplesTable.jsx
```diff
@@ -221,19 +222,14 @@ const SamplesTable = forwardRef((props, ref) => {
...
 
-    // Set table data
     const newData = activeProject.samples.map((sampleUuid, idx) => {
-      // upload problems sometimes lead to partial updates and incosistent states
-      // in this situation it's possible that the sampleUuid does not exist
-      // this a temporary fix so that the whole UI doesn't crash preventing the
-      // user from removing the dataset or uploading another one.
-      // this situation should be included into the SamplesTable.jsx tests
-      const sampleFiles = samples[sampleUuid]?.files || {};
+      const sampleData = samples[sampleUuid];
```
ivababukova: for old projects with partially uploaded samples, wouldn't this fail?

src/utils/upload/processUpload.js
aerlaut: Ok, I'll remove the if statement so we log all errors

src/components/data-management/SamplesTable.jsx
aerlaut: Sorry - I mean without authorization, because these public datasets are not related to any experiment ids.

src/components/data-management/SamplesTable.jsx
aerlaut: No, `sampleData` would be assigned `undefined`. The line that causes the error and the bug in the previous case was line 257 in the old file, where we try to access the property `name` of `undefined`

src/components/data-management/SamplesTable.jsx
```diff
@@ -254,12 +241,12 @@ const SamplesTable = forwardRef((props, ref) => {
 
       return {
         key: idx,
-        name: sampleData.name,
```
aerlaut: This line was causing the error if `sampleUuid` doesn't exist in `samples`

ivababukova (approved):

src/components/data-management/SamplesTable.jsx
aerlaut: Ah I see your point. There's no need to have `sampleCreateErrorTemplate` because we only need the `name` and `samples`, and the cases where these are empty are already handled when we're generating the data for the table. I agree. I've removed the `sampleCreateErrorTemplate` and assign the default value of the file name to error if the sample record doesn't exist.

PR #647, opened 2022-03-03
[BIOMAGE-1631] Add UI component to share experiment

StefanBabukov:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1631

https://ui-grasp6-ui647-api313.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/api/pull/313

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


vickymorrison:
1. Can you change the button order to: Add samples, Add metadata, Download, Share, Go to data processing. I think this makes much more sense than having Share first!
2. What is this thing at the bottom for? It looks like a button so I tried to click it but it doesn't do anything. Is it just to tell the user the name of the project that is being shared? If so, it would make more sense to have it at the top e.g. where the red arrow is in the second screenshot. And make it look less like a button please.
Screenshot 2022-03-03 at 16 33 32
Screenshot 2022-03-03 at 16 36 30
3. The text description of an explorer needs to be updated because as far as I know the ability to clone a project is not implemented yet. I suggest updating to the following but would like @kafkasl 's input too:
"Explorer: the user will be able to use Data Exploration and Plots and Tables modules, but will not be able to make any changes to samples or metadata in Data Management or re-run the pipeline in the Data Processing module."

vickymorrison:
Also, I think 'share with collaborators' is better than 'share with people'.

vickymorrison:
I also can't get the sharing feature to work end to end - I do receive email notification that a project has been shared with me but then I get 'Analysis not found' when I try to access it.

src/components/EditableField.jsx
```diff
@@ -115,12 +126,21 @@ const EditableField = (props) => {
 
     return (
       <>
-        {renderBold ? <strong>{editedValue}</strong> : <span>{editedValue}</span>}
+        {renderBold ? <strong>{value}</strong> : <span>{value}</span>}
```
cosa65, line 129: This was one of the fixes for the problem: we were displaying the internal value `EditableField` uses for keeping track of user actions (`editedValue`) and of course if the request to modify this value fails that isn't really going to cause any re render because nothing changed (`value` is still the same) so `editedValue` wasn't updated.

As a consequence we were left with the latest value the user had put, which was the new changed label that had failed to be updated in the api.

With this change we are going to make sure that the value we are showing is the `value`, (we still need `editedValue` for internal onChange tracking of some edge use cases like when we have external button like save in the cellSets tool creation of new cellSets).

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+
+    const responses = await Promise.all(requests)
+      .then((res) => (
+        Promise.all(
+          res.map((data) => data.json()),
+        )));
+
+    responses.forEach((response, indx) => {
+      if (response?.data?.code === 200) {
+        pushNotificationMessage('success', `${addedUsers[indx]} added to ${experimentName}, they should have been notified.`);
```
kafkasl, line 74: actually they users aren't notified, they will just see their experiment so I would probably change this to:
`user added to experimentName, they should be able to see the experiment in Data Management.

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+  useEffect(() => {
+    loadRoles();
+  }, []);
+
+  const changeSelectedUsers = (selectedUsers) => {
+    const newUser = selectedUsers[selectedUsers.length - 1];
+    // check if the entry is in a valid email address format
+    const invalidEmail = newUser && !newUser?.toLowerCase()
```
aerlaut: This is a boolean right? Can you please change the variable name to better reflect that? E.g. `isEmailInvalid`

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+          role,
+          userEmail: user,
+        }),
+      })));
+
+    const responses = await Promise.all(requests)
+      .then((res) => (
+        Promise.all(
+          res.map((data) => data.json()),
+        )));
```
aerlaut, line 70: This code is already inside an `async` function, why not break down the promise chain and use `await`s?

src/components/data-management/ShareExperimentModal.jsx
aerlaut: Maybe a better success message? e.g. `User ${addedUsers[indx]} has been successfully invited to 
 view ${experimentName}`

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+
+    responses.forEach((response, indx) => {
+      if (response?.data?.code === 200) {
```
aerlaut, line 73: Can we do a better check for this than checking for the status? I'm afraid just checking for the HTTP status might no be enough. For example, what will happen if the user has previously been added to the experiment? What HTTP code will be shown? How will it be differentiated from a successful user addition?

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+          {experimentName}
+        </Text>
+        <Row gutter={10} style={{ width: '110%' }}>
+
+          <Col span={18}>
+            <Select
+              value={addedUsers}
+              style={{ width: '100%' }}
+              mode='tags'
+              placeholder='Input valid email addresses with enter'
```
aerlaut: Separate multiple email addressed with enter

src/components/data-management/ShareExperimentModal.jsx
```diff
@@ -0,0 +1,196 @@
...
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        userEmail,
+      }),
+    });
+
+    if (!response.ok) {
+      pushNotificationMessage('error', 'Error revoking access. You may not have permissions to do this.');
```
aerlaut: Error removing user. You may not have permissions to remove users.

src/redux/actions/experiments/updateExperiment.js
```diff
@@ -1,24 +1,82 @@
...
-  EXPERIMENTS_UPDATED,
-} from '../../actionTypes/experiments';
-import endUserMessages from '../../../utils/endUserMessages';
-import pushNotificationMessage from '../../../utils/pushNotificationMessage';
-import saveExperiment from './saveExperiment';
+  EXPERIMENTS_UPDATED, EXPERIMENTS_SAVING, EXPERIMENTS_SAVED, EXPERIMENTS_ERROR,
+} from 'redux/actionTypes/experiments';
+import endUserMessages from 'utils/endUserMessages';
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import { isServerError, throwIfRequestFailed } from '../../../utils/fetchErrors';
```
aerlaut: Make this into an absolut import

src/redux/actions/experiments/updateExperiment.js
```diff
@@ -1,24 +1,82 @@
...
+
+const convertedToApiModel = (experiment) => {
```
aerlaut: convertToAPIModel?

kafkasl (changes requested):
There are some more errors similar to the name remaining edited even if the user doesn't have permissions. 
1. Renaming a sample
2. Deleting a sample
3. Adding metadata

When trying to run the pipeline, instead of just getting a permission denied, the UI gets stuck trying to load but never does.

src/__test__/components/data-management/ProjectMenu.test.jsx
```diff
@@ -35,7 +47,18 @@ const defaultAPIResponse = generateDefaultMockAPIResponses(
...
+    name: 'Bob',
+    email: 'bob@bob.com',
+    role: 'explorer',
+  },
+  {
+    name: 'Mocked User',
+    email: 'mock@user.name',
+    role: 'owner',
+  }]))),
+});
```
aerlaut: Can you format this piece of code with better identation so it's easier to read?

kafkasl:
> 

@vickymorrison 
2. In the mockups that was not intended to be a clickable button, just a fancy way of showing the experiment name. I believe that styling using a button was the easiest way. We can either make it unclickable, or remove the experiment name altogether if we don't think it's necessary.
3. yes, we should remove the clone bit for now

src/__test__/components/data-management/ProjectMenu.test.jsx
```diff
@@ -91,4 +117,29 @@ describe('ProjectMenu', () => {
...
+      );
+    });
+
+    const shareButton = screen.getByText('Share').closest('button');
+
+    await act(async () => {
+      userEvent.click(shareButton);
+    });
+
+    await waitFor(() => expect(screen.getByText('Share with collaborators')).toBeInTheDocument());
```
aerlaut: Can you check if you need this `waitFor`? I think by wrapping the `act` above, we should already give time for the UI to finish its rendering.

kafkasl:
> 

@vickymorrison 
2. In the mockups that was not intended to be a clickable button, just a fancy way of showing the experiment name. I believe that styling using a button was the easiest way. We can either make it unclickable, or remove the experiment name altogether if we don't think it's necessary.
3. yes, we should remove the clone bit for now

kafkasl:
@StefanBabukov do you know why the bundle size has increased for `data-processing` ?

src/__test__/components/data-management/ShareExperimentModal.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+    // revoke button for the user viewing is disabled
+    expect(revokeButtons[1]).toHaveAttribute('disabled');
+  });
+
+  it('Inviting users works', async () => {
+    await renderShareExperimentModal();
+    const input = screen.getAllByRole('combobox');
+    userEvent.type(input[0], 'asd@asd.com{enter}');
+
+    await waitFor(() => expect(screen.getByText('Add')).toBeInTheDocument());
```
aerlaut, line 82: Check if these `waitFor`s are necessary.

src/__test__/components/data-management/ShareExperimentModal.test.jsx
```diff
@@ -0,0 +1,96 @@
...
+    {
+      name: 'Mocked User',
+      email: 'mock@user.name',
+      role: 'owner',
+    }]))),
+  };
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    enableFetchMocks();
```
aerlaut: You only need to do this once in the script, in the root (outside of thsi function) as its impact will be global.

src/__test__/redux/actions/samples/deleteSample.test.js
```diff
@@ -7,9 +7,9 @@ import initialSampleState, { sampleTemplate } from '../../../../redux/reducers/s
 import initialProjectState, { projectTemplate } from '../../../../redux/reducers/projects/initialState';
 
 import { saveProject } from '../../../../redux/actions/projects';
-import { SAMPLES_DELETE, SAMPLES_SAVED } from '../../../../redux/actionTypes/samples';
+import { SAMPLES_DELETE, SAMPLES_SAVED, SAMPLES_SAVING } from '../../../../redux/actionTypes/samples';
 import { PROJECTS_UPDATE } from '../../../../redux/actionTypes/projects';
-import { EXPERIMENTS_UPDATED } from '../../../../redux/actionTypes/experiments';
+import { EXPERIMENTS_SAVED, EXPERIMENTS_SAVING, EXPERIMENTS_UPDATED } from '../../../../redux/actionTypes/experiments';
```
aerlaut: Can you change these relative imports into absolute imports?

src/__test__/redux/actions/experiments/updateExperiment.test.js
```diff
@@ -1,18 +1,21 @@
...
+import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
+
 import {
+  EXPERIMENTS_SAVED,
+  EXPERIMENTS_SAVING,
   EXPERIMENTS_UPDATED,
 } from '../../../../redux/actionTypes/experiments';
-import { updateExperiment, saveExperiment } from '../../../../redux/actions/experiments';
+import { updateExperiment } from '../../../../redux/actions/experiments';
 import initialExperimentState, { experimentTemplate } from '../../../../redux/reducers/experiments/initialState';
```
aerlaut: Can you change these relative imports into absolute imports?

src/components/data-management/ShareExperimentModal.jsx
StefanBabukov: i wanted to convert every key into json() how do you propose i do this? have another variable that does that in another line?

src/components/data-management/ShareExperimentModal.jsx
StefanBabukov: if the user is already added we also receive 200, theres no duplicate entry added but email is sent again. I'm not sure if thats ok behaviour or no, maybe we need to not send the email again

vickymorrison:
My previous comments have been addressed - thanks.

A few additional things:

1. The messages could be improved when a new user is added:
Rather than "User does not exist" -> I suggest "This user does not have a Cellenics account."
Rather than "stefan@biomage.net added to release test uni, they should have been notified." -> I suggest "stefan@biomage.net added to 'release test uni'. They have been notified."

2. I see the revoke button is active already - is this already functional? I thought it wasn't coming until later.

3. The revoke access is enabled for the explorer to revoke the project owner - this should not be possible.

4. On a similar line, I don't think the explorer should be able to share the project further - I suggest disabling the 'Share' button for users with explorer access to a selected project.

5. As Pol mentioned in an earlier comment, when I try to run the pipeline as an explorer I don't get an error message but the UI gets stuck trying to load.

src/components/data-management/ShareExperimentModal.jsx
kafkasl: yes that's the default behavior and it's ok. We can rely on getting a 200. My question is why is needed the `?`. Shouldn't all the responses contain a `code`? If any does not contain a `code` then it's probably something we should fix.

StefanBabukov: yeah sometimes the response doesnt have those parts we need to work on the api and make the response have the same pattern every time. Sometimes the response is just {message: 'blabla';}

kafkasl (approved):

PR #648, opened 2022-03-04
[BIOMAGE-1739] - Show  information on heatmap when all cells are hidden

aerlaut:
# Description
#### Issue
Data exploration throws an error if all cell sets are hidden, causing Data Exploration to crash.

### Cause
The heatmap expects data to be available for display. Otherwise, it'll throw an error.

Follow these steps to reproduce the error
- Go to Data Exploration
- Hide all cell sets in the cell sets tree

### Solution
Add conditional check which checks for the existence of data and show an Empty message if there is no data. This prevents the Heatmap from being rendered if there is no cell sets to be show and prevet the error from being thrown.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1739

https://ui-agi-ui648.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/components/data-exploration/heatmap/HeatmapPlot.jsx
+++ b/src/components/data-exploration/heatmap/HeatmapPlot.jsx
@@ -58,8 +58,6 @@ const HeatmapPlot = (props) => {
   const [geneHighlight, setGeneHighlight] = useState(null);
   const [cellHighlight, setCellHighlight] = useState(null);
 
-  const [vitessceData, setVitessceData] = useState(null);
-
   const cellCoordinatesRef = useRef({ x: 200, y: 300 });
 
   const expressionData = useSelector((state) => state.genes.expression);
@@ -170,12 +168,6 @@ const HeatmapPlot = (props) => {
     }
   }, [cellHighlight]);
 
-  useEffect(() => {
-    if (!heatmapData) return;
-
-    setVitessceData(heatmapData);
-  }, [heatmapData]);
-
   if (isHeatmapGenesLoading || cellSetsLoading) {
     return (
       <center>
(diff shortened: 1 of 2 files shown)


src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -170,12 +165,6 @@ const HeatmapPlot = (props) => {
     }
   }, [cellHighlight]);
 
-  useEffect(() => {
-    if (!heatmapData) return;
-
-    setVitessceData(heatmapData);
-  }, [heatmapData]);
```
aerlaut, line 177: This is redundant, we can just directly use `heatmapData` in the component instead of setting it to `vitessceData`

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -206,33 +195,14 @@ const HeatmapPlot = (props) => {
     );
   }
 
-  if (selectedGenes?.length === 0) {
```
aerlaut, line 209: We can not remove all genes from the heatmap, so this code path is never reached.

When removing genes, we filter the currently displayed genes against the list that we want to remove, creating a new list of genes that we want to display ([ref](https://github.com/hms-dbmi-cellenics/ui/blob/7180879ee2c35cb6ecf8282237ff5954edde9353/src/components/data-exploration/generic-gene-table/ComponentActions.jsx#L30-L34)). Removing all genes will result in an empty array, which is passed to `loadGeneExpression`. However, there is a check in `loadGeneExpression` which displays genes that are already loaded, in case an empty ilst of genes is passed to `loadGeneExpression` ([ref](https://github.com/hms-dbmi-cellenics/ui/blob/417c9d5fa11fc79f772824f5bae7ef880493d74b/src/redux/actions/genes/loadGeneExpression.js#L49-L59)).

aerlaut: We can either :
- Modify `loadGeneExpression` to allow for no genes to be show in the heatmap (need a deeper exploration, maybe in another ticket)
- Or leave it just like this, if there's no problem.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
```diff
@@ -206,33 +195,14 @@ const HeatmapPlot = (props) => {
...
-    return (
-      <Empty
-        description={(
-          <Text>Add some genes to this heatmap to get started.</Text>
-        )}
-      />
-    );
-  }
-
-  if (cellSetsHierarchy.length === 0) {
```
aerlaut, line 219: This code is never reached, because this [conditional use effect](https://github.com/hms-dbmi-cellenics/ui/blob/aa71130be817de27e50e985c70ec7f3d86b143cc/src/components/data-exploration/heatmap/HeatmapPlot.jsx#L130-L142) prevents `heatmap` data from being set, causing [the conditional for platform error](https://github.com/hms-dbmi-cellenics/ui/blob/aa71130be817de27e50e985c70ec7f3d86b143cc/src/components/data-exploration/heatmap/HeatmapPlot.jsx#L176) to show up before this code is reached.

src/components/data-exploration/heatmap/HeatmapPlot.jsx
cosa65: We should try to find out the reasoning behind that check, but instinctively I'd say we need to have the ability to remove all genes because it's what a user would expect

cosa65 (approved):

PR #650, opened 2022-03-07
[BIOMAGE-1684] - Add DE warning

aerlaut:
# Description
#### Background
Running between-group DE comparison requires at least 2 samples to run. For comprehensive statistitics, we need the cells involved in the comparison to come from at least 3 different samples.

This PR provides a warning to the user to let them know if the group they have chosen to compare does not contain enough cells for complete statistical analysis.

#### To test
To test the check that is implemented here, you need to have a project with at least 4 samples.

- Go to Data Exploration, open the Differential Expression tab.
- Select comparison. between groups
- Choose a cluster located somewhere in the early middle of the cluster list (Cluster 0).
- Select one of the samples for the first group to compare
- Select "all" or "rest of samples".
- Change the cluster that is used as the basis.  You might be able to trigger the warning and the error alert.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1684

https://ui-agi-650.scp-staging.biomage.net

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -351,12 +362,31 @@ const DiffExprCompute = (props) => {
...
               showIcon
-            /> : <></>
+            /> :
+            isFormValid && canRunDiffExpr() === canRunDiffExprResults.WARNING ?
+              <Alert
+                message="Warning"
+                description={
+                  <>
+                    In the selected samples/groups, less than 3 samples contain enough cells to be compared.
+                    The differential expression analysis can be computed, but the adjusted p-value can not be calculated.
```
alexvpickering:
```suggestion
                    For the selected comparison, there are fewer than 3 samples with the recommended minimum number of cells (10). Only logFC values will be calculated and results should be used for exploratory purposes only.
```

will need to adjust the test as well if you agree

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -150,13 +159,22 @@ const DiffExprCompute = (props) => {
...
+    const numCellSetSampleWithEnoughCells = numSampleWithEnoughCells(filteredCellSetCellIds)
+    const numCompareWithSampleWithEnoughCells = numSampleWithEnoughCells(filteredCompareWithCellIds)
+
+
+    if (numCellSetSampleWithEnoughCells === 0 || numCompareWithSampleWithEnoughCells === 0) return canRunDiffExprResults.ERROR;
+
+    const sumComparedSamples = numCellSetSampleWithEnoughCells + numCompareWithSampleWithEnoughCells;
+
+    if (sumComparedSamples < MIN_NUM_SAMPLES_SHOW_WARNING) return canRunDiffExprResults.WARNING;
+    if (sumComparedSamples < MIN_NUM_SAMPLES_SHOW_ERROR) return canRunDiffExprResults.ERROR;
```
alexvpickering:
```suggestion
    if (sumComparedSamples < MIN_NUM_SAMPLES_SHOW_ERROR) return canRunDiffExprResults.ERROR;
    if (sumComparedSamples < MIN_NUM_SAMPLES_SHOW_WARNING) return canRunDiffExprResults.WARNING;
```
These are in the wrong order. The second check will never evaluate to true.

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -22,6 +22,15 @@ const getCellSetKey = (name) => (name?.split('/')[1] || name);
 const getRootKey = (name) => name?.split('/')[0];
 
 const MIN_NUM_CELLS_IN_GROUP = 10;
+const MIN_NUM_SAMPLES_SHOW_ERROR = 2;
+const MIN_NUM_SAMPLES_SHOW_WARNING = 3;
```
alexvpickering: I would call these e.g. `MIN_NUM_SAMPLES_OMIT_ERROR` or similar. Alternatively, decrement these by one and make the checks below `<=`.

I don't think it makes lexical sense right now as you only show the error if there are less than (but not equal to) the minimum number of samples to show an error (`MIN_NUM_SAMPLES_SHOW_ERROR`).

alexvpickering (changes requested):
I added a couple of comments/suggestions but overall looks top notch as always. Let me know when you address them and I'll approve

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: Righ! Thanks for catching it Alex!

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: No problme, added :+1:

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: Hmm.. yes, I think it's easier to understand.

alexvpickering (commented):
The changes you made look great Agi! I'm feeling off today and didn't carefully check the code so you may want to have an engineer take a look as well

alexvpickering (approved):

src/__test__/utils/cellSets/getAllCells.test.js
```diff
@@ -25,7 +25,7 @@ describe('Get all cells', () => {
 
     expect(allCellIds[0]).toMatchInlineSnapshot(`
       Object {
-        "cellId": 11,
+        "cellId": 1,
```
cosa65, line 28: Why was this changed?

src/__test__/utils/plotSpecs/generateEmbeddingCategoricalSpec.test.js
```diff
@@ -19,31 +19,18 @@ describe('filterCells', () => {
...
-      .find(({ key }) => key === 'louvain').children
-      .map(({ key, name, color }) => ({ key, name, color }));
-
-    // Contains all the louvain cellSets (and in the correct order)
-    expect(result.cellSetLegendsData).toEqual(expectedCellSetKeys);
-
-    // CellIds are filtered fine
-    expect(result).toMatchSnapshot();
-  });
-
```
cosa65, line 35: Why was this test removed?

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -150,13 +159,21 @@ const DiffExprCompute = (props) => {
           numCellsPerSampleInCellSet[sampleIdx] += 1;
         });
 
-      return numCellsPerSampleInCellSet.find(numCells => numCells > MIN_NUM_CELLS_IN_GROUP)
+      return numCellsPerSampleInCellSet.filter(numCells => numCells >= MIN_NUM_CELLS_IN_GROUP).length;
     }
 
-    const cellSetHasSampleWithEnoughCells = hasSampleWithEnoughCells(filteredCellSetCellIds)
-    const compareWithHasSampleWithEnoughCells = hasSampleWithEnoughCells(filteredCompareWithCellIds)
+    const numCellSetSampleWithEnoughCells = numSampleWithEnoughCells(filteredCellSetCellIds)
```
cosa65: `canRunDiffExpr` seems pretty long and complex, do you think it could be extracted to a function an imported here? And if there's also other functions like this maybe they should be changed too.

The jsx file could just have the logic more directly related to what to render and call functions for the processing, etc.

src/__test__/utils/plotSpecs/generateEmbeddingCategoricalSpec.test.js
aerlaut: This is a duplicated test with the one on line 111 in the old file

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: This logic is only used in this file. It is also tightly coupled with the variables and state of the function, so I think there's no need to extract it out.

Some of the cell set operations inside the function is repeated a lot throughout the code base though. For these operations, I agree we can factor it out into separate functions. However, I think it is a general refactor better left for another ticket / hackathon.

src/__test__/utils/cellSets/getAllCells.test.js
aerlaut: I added cells and made changes to the mock `__test__/data/cell_sets.json` file to be able to test the check introduced here. If you look at the diff for [cell_sets.json](https://github.com/hms-dbmi-cellenics/ui/pull/650/files#diff-4a1bbbf10ce496597b0ecff72b7426bc5942ba042d604319475c5110650c41d1R145-R153), the first cell id of sample `"b62028a1-ffa0-4f10-823d-93c9ddb88898"` was initiallly `11`.  I rearranged the cells in the sample and it was changed to 1.

cosa65: Woah, I would have though that changing cell_sets.json would have caused the need for way more changes in unit tests since it is being used in virtually all of the new unit tests.

Next time, however, if you need a new dataset I think I would recommend just creating a different `.json` for your use case instead of changing an existing one and having to update a lot of different tests.

cosa65: I think sometihng good about the previous `cell_sets.json` was that it was simple so it was easy to understand the tests we were making with it

aerlaut: this is also simple - there's just more cells

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
cosa65: I agree, we don't need to do a general refactor on the cell set operations here.

My reason for separating responsibilities was not mainly about reusability, more about readability and being able to have a clearer separation, an example of something similar could be `processUpload` or `calculateVolcanoDataPoints`, they are all about doing more involved calculations and they are extracted into functions even when we could easily just have them i one place in the `jsx` that uses them.

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -14,14 +14,20 @@ import { loadCellSets } from 'redux/actions/cellSets';
...
 const getCellSetKey = (name) => (name?.split('/')[1] || name);
 const getRootKey = (name) => name?.split('/')[0];
 
-const MIN_NUM_CELLS_IN_GROUP = 10;
+const canRunDiffExprResults = {
+  TRUE: 'TRUE',
+  FALSE: 'FALSE',
+  INSUFFICIENT_CELLS_WARNING: 'INSUFFICIENT_CELLS_WARNING',
+  INSUFFCIENT_CELLS_ERROR: 'INSUFFCIENT_CELLS_ERROR'
+}
```
cosa65: We could move this to checkCanRunDiffExpr

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -97,66 +103,23 @@ const DiffExprCompute = (props) => {
...
-        });
-
-      return numCellsPerSampleInCellSet.find(numCells => numCells > MIN_NUM_CELLS_IN_GROUP)
-    }
+  const canRunDiffExpr = useCallback(() => {
 
-    const cellSetHasSampleWithEnoughCells = hasSampleWithEnoughCells(filteredCellSetCellIds)
-    const compareWithHasSampleWithEnoughCells = hasSampleWithEnoughCells(filteredCompareWithCellIds)
+    if (selectedComparison === ComparisonType.WITHIN) return canRunDiffExprResults.TRUE;
+    if (!basis || !cellSet || !compareWith || !cellIdToSampleMap.length > 0) { return canRunDiffExprResults.FALSE; }
```
cosa65: We could also move this to checkCanRunDiffExpr

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: moved

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: moved

cosa65 (approved):

cosa65:
aprobado

PR #652, opened 2022-03-09
Add api v2 basic structure 

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1783

https://ui-martinfosco-ui652-api317.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/api/pull/317

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/utils/constants.js
```diff
@@ -33,9 +33,18 @@ const modules = {
   SETTINGS: 'SETTINGS',
 };
 
+const api = {
+  CURRENT_VERSION: 'v1',
```
cosa65: This is the way to set the api version we want to hit, I don't think this is the best option but wasn't sure about alternatives, so any help or ideas on this would be appreciated

aerlaut: This is more like a configuration right instead of constants? Would it be better if we store it inside a file that's used for configuration? e.g. `config.json`? constants is mostly used for values that we do not want to change right?

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,47 +1,99 @@
...
     createdDate,
   };
 
+  let url;
+  let experimentToSend;
+
   try {
-    dispatch(saveExperiment(newExperiment.id, newExperiment, false));
+    if (api.CURRENT_VERSION === api.possibleVersions.V1) {
+      experimentToSend = newExperiment || getState().experiments[experimentId];
```
aerlaut: `newExperiment` is always defined right? Do we still need `|| getState().experiments[experimentId]`?

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,47 +1,99 @@
...
+      experimentToSend = experimentConvertedToApiModel(experimentToSend);
+
+      url = `/v1/experiments/${experimentId}`;
+    } else if (api.CURRENT_VERSION === api.possibleVersions.V2) {
+      const { id, name, description } = newExperiment || getState().experiments[experimentId];
```
aerlaut: Likewise here, I don't think we need `|| getState().experiments[experimentId]`.

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,47 +1,99 @@
...
-    pushNotificationMessage('error', endUserMessages.ERROR_SAVING);
 
+    dispatch({
+      type: EXPERIMENTS_SAVED,
+    });
+  } catch (e) {
+    let { message } = e;
+    if (!isServerError(e)) {
+      console.error(`fetch ${url} error ${message}`);
+      message = endUserMessages.CONNECTION_ERROR;
```
aerlaut: This line will override the `message` that is stored on line 89. Do we want to store the error message if it's not server error? Or replace it with the `endUserMessage`?

src/utils/convertExperimentToApiModel.js
```diff
@@ -0,0 +1,20 @@
+// There are some differences between the property naming of elements
+// stored in the ui and in the api,
+// this is an attempt to deal with this in one single place (the ui)
+// We should try to converge to one single model to follow
+const experimentConvertedToApiModel = (experiment) => {
```
aerlaut: maybe `convertExperimentToApiModel`? Sounds easier to read and in active voice.

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,47 +1,99 @@
...
+    id: experimentId,
     name: newExperimentName,
     projectUuid,
     createdDate,
   };
 
+  let url;
+  let experimentToSend;
+
   try {
```
aerlaut: Why do we need the try block to begin here? Are we expecting errors during the conversion of the models? Otherwise I think the try block can begin much later in the code, just after dispatching experiments saving.

src/__test__/redux/actions/experiments/createExperiment.test.js
```diff
@@ -1,19 +1,18 @@
...
 
-import '__test__/test-utils/setupTests';
+import { EXPERIMENTS_SAVING, EXPERIMENTS_CREATED, EXPERIMENTS_SAVED } from 'redux/actionTypes/experiments';
+import { createExperiment } from 'redux/actions/experiments';
+import initialExperimentState from 'redux/reducers/experiments/initialState';
+import initialProjectState, { projectTemplate } from 'redux/reducers/projects/initialState';
 
-jest.mock('../../../../redux/actions/experiments/saveExperiment');
-saveExperiment.mockImplementation(() => async () => { });
+import '__test__/test-utils/setupTests';
```
aerlaut: Do we need to import this?

src/utils/convertExperimentToApiModel.js
ivababukova: can you put this new file into a new folder inside `src/utils`? I imagine that there will be functions for converting other data too, it will be good to give it some structure.

src/utils/convertExperimentToApiModel.js
```diff
@@ -0,0 +1,20 @@
+// There are some differences between the property naming of elements
+// stored in the ui and in the api,
+// this is an attempt to deal with this in one single place (the ui)
+// We should try to converge to one single model to follow
```
ivababukova: I agree. It will be best to not have to do such conversions. When do you think it will be a good time to do this conversion?

ivababukova: Nevermind, I actually see that for the v2 version, the api uses the same property names and there is no need for conversion there. Sounds great!

ivababukova: In this case, you can changes this comment to say that after we've completed the migration to api v2, the conversion will not be necessary. You can also update the function name to state that the conversion is to v1 model. For example `convertExperimentToApiV1Model`

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,47 +1,99 @@
...
       payload: {
         experiment: newExperiment,
       },
     });
-  } catch (e) {
-    pushNotificationMessage('error', endUserMessages.ERROR_SAVING);
 
+    dispatch({
+      type: EXPERIMENTS_SAVED,
+    });
```
ivababukova: why do we need to dispatch these two redux actions in a row? Is there any situation when we need to dispatch only one of them, but not the other? If there is no such situation, I suggest that we combine them into 1 action.

src/utils/constants.js
cosa65: Exactly, that's why I didn't like placing it here, where do you think this would fit better? Should I come up with a new `config.json` file?

src/utils/convertExperimentToApiModel.js
cosa65: based on [this comment](https://github.com/hms-dbmi-cellenics/ui/pull/652#discussion_r823550791), I think we can skip moving this into a folder sincre we are going to get rid of this function soon anyways.

src/redux/actions/experiments/createExperiment.js
```diff
@@ -1,31 +1,69 @@
...
     ...experimentTemplate,
-    id: hash.MD5(createdDate),
+    id: experimentId,
     name: newExperimentName,
     projectUuid,
     createdDate,
   };
 
+  let url;
+  let experimentToSend;
```
aerlaut, line 37: requestPayload?

src/redux/actions/projects/createProject.js
```diff
@@ -24,8 +25,12 @@ const createProject = (
 
   const newProjectUuid = uuidv4();
 
-  // Always create an experiment for a new project
-  // required because samples DynamoDB require experimentId
+  if (api.CURRENT_VERSION === api.possibleVersions.V2) {
+    await dispatch(createExperiment(newProjectUuid, newExperimentName));
+
+    return;
```
aerlaut, line 34: Are you returning here because there's no project creation anymore in the v2? Can you add comment about that here?

src/redux/actions/projects/createProject.js
```diff
@@ -24,8 +25,12 @@ const createProject = (
 
   const newProjectUuid = uuidv4();
 
-  // Always create an experiment for a new project
-  // required because samples DynamoDB require experimentId
+  if (api.CURRENT_VERSION === api.possibleVersions.V2) {
+    await dispatch(createExperiment(newProjectUuid, newExperimentName));
```
aerlaut, line 31: Do we need to await here?

src/utils/convertExperimentToApiV1Model.js
```diff
@@ -0,0 +1,20 @@
+// There are some differences between the property naming of elements
+// stored in the ui and in the api,
+// this is an attempt to deal with this in one single place (the ui)
+// We should try to converge to one single model to follow
+const convertExperimentToApiV1Model = (experiment) => {
```
aerlaut, line 5: Will there be any more models that need conversion for v1 to v2? If there are, maybe better make a folder to store these functions or export this function as a named export so people can extend easily

cosa65: No, this isn't converting v1 to v2

cosa65: It was added a couple of months ago already, it's about the differences between the experiment in redux vs the one being stored in dynamo

cosa65: Also it will no longer be necessary when we get rid of v1

src/redux/actions/projects/createProject.js
cosa65: When possible, it is better to link the async functions to its internal async processes when possible so that the async handling can be brought up to the containing function I think

src/redux/actions/experiments/createExperiment.js
cosa65: but we don't need to make this generic since we know it is an experiment to send, right?

aerlaut: yea... ok

aerlaut (approved):

PR #654, opened 2022-03-10
[BIOMAGE-1765] - Notify Slack on error

aerlaut:
# Description
#### Issue
We want to be able to monitor fatal error ("It's not you, it's us") errors thrown in production

#### Solution
[NextJS](https://nextjs.org/docs/advanced-features/error-handling#handling-client-errors) and [React](https://reactjs.org/docs/error-boundaries.html) recommends the use of React error boundaries to catch client-side errors. 

I use [a library](https://github.com/bvaughn/react-error-boundary) which implements this. The library exposes a wrapper component with an `onError` handler. Inside this handler, I used a Slack bot to post error logs to the #error-logs channel. The error log that is posted consists of : 
- The error message, without the stack trace. I omitted the stack trace because the stack trace does not accurately identify the position of the error, as the error originates from the compiled JS file, instead of the source file. I wanted to show the exact position of the error as we have in, however it turns out this is more complicated than it seems*.
- The component stack
- The Redux store dump

To correctly map the source of the error, we have to:
- Get the coordinate (line and column) of the error
- Get the source map file
- Remap the apparent coordinate to the source coordinate

However, looking at the error generated in production, the sourcemap does not map to source, but a compiled version of source. Even if I try to read the generated source map, it'll still not give the correct coordinate.

[Next.js version 11 comes with native sourcemap support](https://stackoverflow.com/questions/68505618/how-do-i-generate-sourcemaps-for-nextjs). Our Next.JS version is version 9.5.5. Upgrading to version 11 might allow us to have better source map support. However, we have to be careful when upgrading. 

I suggesst we upgrade first before revisiting this issue.

#### To test
One of the error that still persists is the single-sample DE "go-back" error : https://biomage.atlassian.net/browse/BIOMAGE-1764

To reproduce this : 
- Checkout this branch in your local environment
- Comment line 136 in `_app.jsx` so that logging works for development environment
- Create a new project and upload a single sample. Process the project.
- Go to Data Management, launch DE with any options
- Click "Go Back"
- This should throw an error, which will show an error screen in the UI.
- The error will also throw a log in the #error-logs channel in Slack.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1765

https://ui-agi-ui654.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/pages/_app.jsx
```diff
@@ -124,18 +129,28 @@ const WrappedApp = ({ Component, pageProps }) => {
...
-          routeExperimentId={experimentId}
-          experimentData={experimentData}
-        >
-          <Component
-            experimentId={experimentId}
+      <ErrorBoundary
+        FallbackComponent={Error}
+        onError={(error, info) => {
+          // Only log errors to Slack if not in production.
+          if (process.env.NODE_ENV !== 'production') return;
```
StefanBabukov, line 136: Why exclude production errors?

src/pages/_app.jsx
```diff
@@ -124,18 +129,28 @@ const WrappedApp = ({ Component, pageProps }) => {
...
-        <ContentWrapper
-          routeExperimentId={experimentId}
-          experimentData={experimentData}
-        >
-          <Component
-            experimentId={experimentId}
+      <ErrorBoundary
+        FallbackComponent={Error}
+        onError={(error, info) => {
+          // Only log errors to Slack if not in production.
```
StefanBabukov: if in*

src/utils/crypt.js
```diff
@@ -36,8 +36,18 @@ const getWebhookUrl = () => {
...
+  };
+
+  return decrypt(botToken);
+};
+
 export {
   decrypt,
   encrypt,
   getWebhookUrl,
+  getLoggerBotToken,
```
StefanBabukov: Maybe change `getLoggerBotToken` to getLoggerWebhookUrl or something like that - the idea is that the similar to the feedback hook - `getWebhookUrl` just different channel, so they should have similar names?

src/pages/experiments/[experimentId]/data-exploration/index.jsx
```diff
@@ -120,11 +122,12 @@ const ExplorationViewPage = ({
     'Data Management': {
       toolbarControls: <MosaicCloseButton key='remove-button-data-management' />,
       component: (width, height) => (
-        <CellSetsTool
-          experimentId={experimentId}
-          width={width}
-          height={height}
-        />
+        <></>
+        // <CellSetsTool
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/data-exploration/index.jsx]: Why comment all the plots?

src/utils/AppRouteProvider.jsx
```diff
@@ -30,6 +30,7 @@ const PATH_STUBS = {
   [modules.DATA_EXPLORATION]: '/data-exploration',
   [modules.PLOTS_AND_TABLES]: '/plots-and-tables',
   [modules.SETTINGS]: '/settings',
+  [modules.DEFAULT]: '/',
```
StefanBabukov, line 33: Do we need this endpoint?

src/utils/postErrorToSlack.js
```diff
@@ -0,0 +1,98 @@
...
+    user, timestamp, experimentId, url,
+  } = context;
+
+  const message = `
+  \u26A0  Uncaught UI Error - ExpID ${experimentId} - ${timestamp}
+  URL: ${url}
+
+  User: ${user.attributes.name}
+  Email: ${user.attributes.email}
+  User ID: ${user.username}
```
StefanBabukov: Can you not get the message from `buildErrorMessage` and just pass it here without creating a new message variable with those fields again? I see there are repeating  values in both places

src/utils/crypt.js
aerlaut: It's different - we're not using a webhook in this case, but using a bot to post on our behalf.

src/pages/experiments/[experimentId]/data-exploration/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-exploration/index.jsx]: Aah sorry, I was commenting this out to make development faster. Forgot to uncomment it. Uncommented.

src/utils/AppRouteProvider.jsx
aerlaut: It's a fallback case so that the application does not error out when visiting pagest that are not in the module, e.g. `_error.jsx`. I discover this when I'm trying to modify the error page.

src/utils/postErrorToSlack.js
aerlaut: `buildErrorMessage` builds the string that is going to be posted as the error log, which appears in the collapsible part of the error. These lines build the text that appears in the post message - it's primary purpose is to help distinguish one error message from another. So they're for different purposes, although they might look the same.

src/pages/_app.jsx
```diff
@@ -124,18 +129,28 @@ const WrappedApp = ({ Component, pageProps }) => {
...
-          routeExperimentId={experimentId}
-          experimentData={experimentData}
-        >
-          <Component
-            experimentId={experimentId}
+      <ErrorBoundary
+        FallbackComponent={Error}
+        onError={(error, info) => {
+          // Only log errors to Slack if in production.
+          if (process.env.NODE_ENV !== 'production') return;
```
kafkasl: could you put this logic into `postErrorToSlack`? I think putting it here makes it harder to read the `app.jsx` file

src/pages/_app.jsx
```diff
@@ -124,18 +129,28 @@ const WrappedApp = ({ Component, pageProps }) => {
...
     return (
-      <AppRouteProvider>
-        <ContentWrapper
-          routeExperimentId={experimentId}
-          experimentData={experimentData}
-        >
-          <Component
-            experimentId={experimentId}
+      <ErrorBoundary
+        FallbackComponent={Error}
```
kafkasl: what does the fallback component do?

src/utils/AppRouteProvider.jsx
kafkasl: doesn't the error have a specific path like `/error` or something? so we can use that instead of having default?

src/utils/crypt.js
kafkasl: as I mentioned it'd be grate if you add some documentation regarding that, because it was unclear to me and also for stefan, so it will probably be too for the rest of people

src/utils/postErrorToSlack.js
kafkasl: I would probably collapse the user info into a single line
`User: ${user.attributes.name} (${user.attributes.email}) ${user.username}`

src/utils/postErrorToSlack.js
```diff
@@ -0,0 +1,98 @@
...
+  });
+
+  try {
+    const res = await fetch('https://slack.com/api/files.upload', {
+      method: 'POST',
+      body: formData,
+    });
+
+    if (!res.ok) {
+      console.error('Failed sending error to slack');
```
kafkasl: would be great to know more details about the why this failed; at least add the status code of the response

src/pages/_app.jsx
aerlaut: This is the component that will be rendered if there is an error.

src/pages/_app.jsx
aerlaut: moved

src/utils/crypt.js
aerlaut: added

src/utils/AppRouteProvider.jsx
aerlaut: No, the error does not have any specific path. The path `/_error`  is actually just a service path that helps us in developing the error view for development. It is not a path that should be accessible to users in production. When there is an error, NextJS renders the error in into view, not redirect the browser to this URL.

kafkasl:
I am concerned that this PR is increasing a lot the size of the main bundle and that will slow down and degrade all the UI. Could you evaluate alternatives? see if you can maybe place it in another place, or see how to decrease the size. Or maybe use a lighter or custom implementation. 

Looking at this https://bundlephobia.com/package/react-error-boundary@3.1.4 it seems that the library should be only 2.4Kb, could you investigate?

I've seen it can be done with a custom component too https://reactjs.org/docs/error-boundaries.html

src/utils/postErrorToSlack.js
aerlaut: error logged

src/utils/crypt.js
```diff
@@ -1,43 +1,19 @@
...
-};
-
-const decrypt = (hash) => {
-  const decipher = crypto.createDecipheriv(ENCRYPTION_ALGORITHM, NON_SECURE_SECRET_KEY, Buffer.from(hash.iv, 'hex'));
-
-  const decrpyted = Buffer.concat([decipher.update(Buffer.from(hash.content, 'hex')), decipher.final()]);
-
-  return decrpyted.toString();
-};
+const decrypt = (hash) => window.atob(escape(hash));
```
kafkasl: 1. any reason to use `window.atob` instead of this? https://stackoverflow.com/a/6182519 . I am concerned that `window` is only available in the browser so it might give errors when compiling for SSR or if it's called by accident.
2. Considering that this file now only exports functions for slack, it no longer makes sense to call it `crypt.js`, I would probably rename it to `slack.js`

src/utils/crypt.js
```diff
@@ -1,43 +1,19 @@
...
 
+// Token for Logger bot to post errors as file
+// Required becuase webhook doesn't suport file uploads
+const getLoggerBotToken = () => {
+  const botToken = 'eG94Yi0xMTc1NDc1NTM5NzE0LTMyMjE3MTYwNDg4OTgtUWFXNmlwTEVSY1RJRXQycVU2RnJ3RGp1';
+  return decrypt(botToken);
+};
+
 export {
   decrypt,
```
kafkasl: I think it is no longer needed to export `decrypt`.

src/utils/postErrorToSlack.js
```diff
@@ -0,0 +1,126 @@
...
+  const match = url.match(/experiments\/([^/]+)/i);
+  return match ? match[1] : null;
+};
+
+const takeNumLines = (string, numLines) => string.split('\n').slice(0, numLines).join('\n');
+
+// Truncates data arrays so that it doesn't produce large logs
+const ARRAY_TRUNCATE_LENGTH = 21;
+const truncateCollection = (arr) => {
+  if (arr.length < ARRAY_TRUNCATE_LENGTH) return arr;
```
kafkasl: what is this exactly truncating? the error message to make it collapsible?

src/components/ErrorBoundary.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import postErrorToSlack from 'utils/postErrorToSlack';
+import { connect } from 'react-redux';
+import Error from 'pages/_error';
+
+class ErrorBoundary extends React.Component {
```
kafkasl, line 8: now that we know that the bundle size increase comes from the crypto libraries, do you think it's worth using our custom component instead of the library? do you expect much overhead having to maintain it? do you think there are going to be new features in the library that could be beneficial?

src/__test__/utils/postErrorToSlack.test.js
```diff
@@ -0,0 +1,93 @@
...
+    },
+    username: '5152fake-eb52-474c-user-mocke8c8user',
+  }),
+}));
+
+describe('PostErrorToSlack', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    process.env.NODE_ENV = 'production';
```
kafkasl: Having functions use environment variables makes it trickier to test them because they need global variables to be set and it's much harder to reason about the global state than just the local state. In this situations it's more convenient to have the state passed into the function ie:

`const postErrorToSlack = async (error, info, reduxDump, env) => {`
 
so then you can test them by simply:
```
await postErrorToSlack(mockError, mockInfo, mockReduxDump, 'staging');
await postErrorToSlack(mockError, mockInfo, mockReduxDump, 'production');
```

src/__test__/utils/__snapshots__/postErrorToSlack.test.js.snap
```diff
@@ -0,0 +1,67 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`PostErrorToSlack Posts requests correctly 1`] = `
+Array [
+  "https://slack.com/api/files.upload",
```
kafkasl, line 5: make sure that none of the snapshots will inadvertently post the token / URL

src/components/ErrorBoundary.jsx
aerlaut: Hmm... I don't expect much changes to this component. The only place where I foresee modification is inside the "componentDidCatch" block, which is the handler for the raised error. Besides that, this component basically acts like a try-catch block.

Under the hood, the library is also actually using the same code, but just abstracting the functionality into props.

The equivalent piece of code in this component vs that provided by the library are : 
- Display an error message (`FallbackComponent` in the library's API)
This is implemented by the if statement in the `render()` method. Equivalent to [https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L129-L130](https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L129-L130)

- Catch and act on erorr (`onError` in the library's API)
This is implemented by the code block under `componentDidCatch`. Equivalent to [https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L87-L89](https://github.com/bvaughn/react-error-boundary/blob/master/src/index.tsx#L87-L89)

I don't think the other APIs provided by the library (even `useErrorHandler`) will be used in our use case.

Therefore, I'd say that this custom component suits our use case, with less baggage.

src/utils/crypt.js
aerlaut: This code is run on the browser-side, not the server-side. Meanwhile, the answer in [https://stackoverflow.com/a/6182519](https://stackoverflow.com/a/6182519) is for Node.js / server-side. The `Buffer` object doesn't exist in the browser. so the solution doesn't work. The code passes the CI, so it builds fine.

aerlaut: Moved to `slack/creds.js`

src/utils/crypt.js
aerlaut: removed

src/utils/postErrorToSlack.js
aerlaut: No, this truncates large arrays and sets such as cellIds or work results data, so that it doesn't overwhelm the logs. If we don't truncate this, the size of the logs will be very big because of data and cell ids.

src/__test__/utils/postErrorToSlack.test.js
aerlaut: Actually I'd rather have the calling function decide whether the to call the function or not, rather than passing in the environment variable to the function. This means,

    const ErrorBoundary = () => {
        ...

        if(process.env.NODE_ENV !== 'production') return;
        postErrorToSlack(error, errorInfo);

    }

Therefore, the only concern of `postErrorToSlack` is posting the error to Slack - it doesn't have to know to decide based on the environment.

src/__test__/utils/__snapshots__/postErrorToSlack.test.js.snap
aerlaut: Yes, it's written out here :

https://github.com/hms-dbmi-cellenics/ui/blob/ae1cbb19ac9a4c0610026b3f417a5673151af5b2/src/__test__/utils/postErrorToSlack.test.js#L76-L81

src/components/ErrorBoundary.jsx
kafkasl: sounds good

src/__test__/components/ErrorBoundary.test.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+const content = 'Some content';
+
+// ErrorBoundary renders the `_error.jsx` component if an error occurs.
+// This is a string contained in the `_error.jsx component to test if the component is rendered.
+const errorText = /Sorry, something went wrong on our end/i;
+
+jest.mock('utils/slack/postErrorToSlack');
+
+delete window.location;
```
kafkasl, line 18: why is this delete needed? isn't the assignment in next line enough? what is it used for?

kafkasl, line 18: why is this delete needed? isn't the assignment in next line enough? what is it used for?

src/__test__/utils/postErrorToSlack.test.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+    username: '5152fake-eb52-474c-user-mocke8c8user',
+  }),
+}));
+
+describe('PostErrorToSlack', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    fetchMock.resetMocks();
+    fetchMock.doMock();
```
kafkasl: what does `doMock`?

kafkasl: what does `doMock`?

src/utils/slack/creds.js
```diff
@@ -0,0 +1,18 @@
+const decrypt = (hash) => decodeURIComponent(window.atob(hash));
```
kafkasl: I think creating a an extra folder, and a file for only this was not needed, a single `slack.js` file would have been enough. I don't expect to have much more code related to slack

kafkasl: I think creating a an extra folder, and a file for only this was not needed, a single `slack.js` file would have been enough. I don't expect to have much more code related to slack

kafkasl (approved):

src/__test__/components/ErrorBoundary.test.jsx
aerlaut: I think this has to first be deleted because the `location` object under `window` is not [writable](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#description). This happens in JSDOM version >= 14, according to this blog post: [https://remarkablemark.org/blog/2018/11/17/mock-window-location/#update-for-jsdom-14](https://remarkablemark.org/blog/2018/11/17/mock-window-location/#update-for-jsdom-14)

src/__test__/utils/postErrorToSlack.test.js
aerlaut: `doMock()` enables `fetchMock`. Actually `resetMock()` only clears out the mock data and does not disable/unmock `fetchMock`, so `doMock` is not needed here. I'll remove this.

src/utils/slack/creds.js
aerlaut: Hmm... yea... ok. I put `postErrorToSlack` together with this cred file because they're related to Slack-related functionality. However, it can be argued that most of the functionality in `postErrorToSlack` is not Slack specific. I'll remove it out.

PR #655, opened 2022-03-10
[BIOMAGE-1764] - Fix Error when going back from single-sample DE

aerlaut:
# Description
#### Issue
Clicking "Go Back" from the DE table causes an error to happen. This happens in the piece of code to automatically select the only sample available for the dataset.  The piece of code does dispatch action to update the selected sample, but uses direct assignment, which causes the error.

#### Solution
Dispatch action to update the selected sample correctly.

#### To test
To reproduce the bug :

- Launch a project with a single sample
- Go to Data Management, launch DE with any options
- Click "Go Back"
- This should throw an error, causing an "it's not you, it's us" error page to show up.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1764

https://ui-agi-ui655.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


cosa65:
>  I'll verify with Vicky about this
This you get the chance to check with @vickymorrison on this?

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -30,7 +30,7 @@ const DiffExprCompute = (props) => {
   const dispatch = useDispatch();
   const { properties, hierarchy } = useSelector(getCellSets());
   const [isFormValid, setIsFormValid] = useState(false);
-  const [numSamples, setNumSamples] = useState(1)
+  const [numSamples, setNumSamples] = useState(0);
```
cosa65, line 32: Why was this changed? (and do you know why was it 1 before?)

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
```diff
@@ -46,40 +46,30 @@ const DiffExprCompute = (props) => {
...
-
-        dispatch(
-          setComparisonGroup({
-            type,
-            ...comparisonGroup[type],
-            ...deleteKeys,
-          }),
-        );
-      }
+    Object.keys(comparisonGroup).forEach(
```
cosa65: I don't understand too much, but I am guessing that this is dealing with what happens when one of the selected cellSets is deleted, right?

If that is the case, I don't think this should be dealt with here. It should be dealt with the tools we have in redux directly instead of having this go through a different component.

This is effectively dealing with the consequences of a cellSet being deleted, the tool to use for that is a reducer on the `CELL_SETS_DELETE` action. Right now this `useEffect` is performing the work of a reducer in an indirect way.

We trigger the action `CELL_SETS_DELETE`, as a consequence the reducer of each slice should perform its necessary changes.
In this case, I would add a `cellSetsDelete` reducer in the `differentialExpression` slice of the store that checks whether the selected option shoudlb e unselected, and all of this `useEffect` just goes away.

src/redux/actions/differentialExpression/loadDifferentialExpression.js
```diff
@@ -5,14 +5,14 @@ import {
...
-const getCellSetName = (name) => (name?.split('/')[1] || name);
+import { getCellSetKey } from 'utils/cellSets';
 
 const generateDiffExprBody = (experimentId, comparisonGroup, comparisonType, extras) => ({
   name: 'DifferentialExpression',
   experimentId,
-  cellSet: getCellSetName(comparisonGroup.cellSet),
-  compareWith: getCellSetName(comparisonGroup.compareWith),
-  basis: getCellSetName(comparisonGroup.basis),
+  cellSet: getCellSetKey(comparisonGroup.cellSet),
```
cosa65, line 13: nicee

src/utils/cellSets/getCellSetNames.js
```diff
@@ -0,0 +1,7 @@
+const getCellSetKey = (name) => (name?.split('/')[1] || name);
+const getRootKey = (name) => name?.split('/')[0];
+
+export {
+  getCellSetKey,
+  getRootKey,
+};
```
cosa65: This file is called `getCellSetNames`, where is this format of cell set name defined? Is it used only in differential expression? Do you know why does this format exist?

aerlaut: Hmm... yea, maybe the file name should be `getCellSetKeys`. I'll change it.

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: It was set to 1 before because we assume that there will always be at least 1 sample. That is a correct assumption of course, but we have some behaviors which will trigger specifically when the number of samples is 1, even while the cell set is still loading. After the cell set is loaded, the number of samples might change. This means we'll always trigger the behaviour specific for 1 sample, even though in reality there are more than 1 samples. So, I think the better option is to set the initial number of cellsets to 0, which will then be corrected after the cell sets load.

aerlaut:
> > I'll verify with Vicky about this
> > This you get the chance to check with @vickymorrison on this?

Yes I did. We do not want to remove the behaviour, so I implemented the fix.

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: Great idea! I'll implement that.

src/utils/cellSets/getCellSetNames.js
aerlaut: The format of the cell set name is 'cellClassKey/cellSetKey'. I've adjusted the function name accordingly and put in comment too.

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
aerlaut: implemented

src/redux/reducers/differentialExpression/index.js
```diff
@@ -39,6 +41,9 @@ const differentialExpressionReducer = (state = initialState, action) => {
     case EXPERIMENT_SETTINGS_PIPELINE_START: {
       return initialState;
     }
+    case CELL_SETS_DELETE: {
+      return differentialExpressionUnselectDeletedOption(state, action);
```
cosa65: The naming we are using for reducers mean this should just be `cellSetsDelete`

src/__test__/components/data-exploration/differential-expression-tool/DiffExprCompute.test.jsx
```diff
@@ -334,4 +334,66 @@ describe('DiffExprCompute', () => {
       expect(screen.getByText(/Compute/i).closest('button')).not.toBeDisabled();
     });
   });
+
+  it('Unselect deleted option', async () => {
```
cosa65: Does this test make sense now that we have removed the code handling deletion on the react component?

src/redux/reducers/differentialExpression/differentialExpressionUnselectDeletedOption.js
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable no-param-reassign */
+import produce from 'immer';
+import { getCellSetKey } from 'utils/cellSets';
+
+const differentialExpressionUnselectDeletedOption = produce((draft, action) => {
+  const { key: deletedKey } = action.payload;
+
+  Object.keys(draft.comparison.group).forEach((comparisonType) => {
+    const comparisonGroup = draft.comparison.group[comparisonType];
```
cosa65: Does this work? usually I can't pick up stuff from `draft` without first doing a `current` or `original`

src/redux/reducers/differentialExpression/index.js
cosa65: Think of it as one single reducer being split up across slices

aerlaut: Ahh... ok

src/__test__/components/data-exploration/differential-expression-tool/DiffExprCompute.test.jsx
aerlaut: Yes, the test still stand, because it's testing for the expected outcome, not the functionality.

src/redux/reducers/differentialExpression/differentialExpressionUnselectDeletedOption.js
cosa65: This could be replaced with 
`Object.entries(draft.comparison.group).forEach(([comparisonType, comparisonGroup]) => {`

src/redux/reducers/differentialExpression/differentialExpressionUnselectDeletedOption.js
```diff
@@ -0,0 +1,18 @@
...
+
+    Object.keys(comparisonGroup).forEach((comparisonKey) => {
+      if (getCellSetKey(comparisonGroup[comparisonKey]) !== deletedKey) return;
```
cosa65: Here there's the same pattern of using `Object.keys` but then getting the value instead of using `Object.entries` directly

src/redux/reducers/differentialExpression/differentialExpressionUnselectDeletedOption.js
aerlaut: Yea, it seems to work... I see the intended outcome. I'll check if there is a difference between using `current()` or direct access like this.

src/__test__/components/data-exploration/differential-expression-tool/DiffExprCompute.test.jsx
cosa65: So it is testing the redux action?

src/components/data-exploration/differential-expression-tool/DiffExprCompute.jsx
cosa65: nice

src/__test__/components/data-exploration/differential-expression-tool/DiffExprCompute.test.jsx
aerlaut: Yes

src/redux/reducers/differentialExpression/differentialExpressionUnselectDeletedOption.js
aerlaut: So it seems that `current()` has to be used on a `draft`, which is a proxy to an object. When you directly access a value that is not an object, accessing draft directly returns the value, so it's comparable. If you try to access `current()` on a primtive value (e.g. not-object), it'll throw an error.

In this case, I'm accessing a property which is a primitive(string), so it returns the real value which can then be used to compare.

aerlaut: Ultimately used `Object.values` because we don't need `comparisonType`.

src/redux/reducers/differentialExpression/index.js
aerlaut: reducer renamed

src/redux/reducers/differentialExpression/cellSetsDelete.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable no-param-reassign */
+import produce from 'immer';
+import { getCellSetKey } from 'utils/cellSets';
+
+const differentialExpressionUnselectDeletedOption = produce((draft, action) => {
+  const { key: deletedKey } = action.payload;
+
+  Object.values(draft.comparison.group).forEach((comparisonGroup) => {
+    Object.keys(comparisonGroup).forEach((comparisonKey) => {
+      if (getCellSetKey(comparisonGroup[comparisonKey]) !== deletedKey) return;
```
cosa65: This one wasn't changed for this comment yet, right? https://github.com/hms-dbmi-cellenics/ui/pull/655#discussion_r829051914

aerlaut: I changed it in a378b88 to use

    Object.values(draft.comparison.group).forEach((comparisonGroup) => { ...

cosa65 (approved):

PR #660, opened 2022-03-18
[BIOMAGE-1807] - Continue showing the UI if it doesn't blow up

aerlaut:
# Description
#### Problem
- The QC pipeline throws an error when first reading the config. This was caught by the error boundary element which causes the UI to show an error page. This is fixed.
- The error boundary element shouldn't show an error page if the UI doesn't blow up.

#### Solution
- Continue rendering the child components inside 
- Fix the errors

#### To test
- Go to Data Processing
- Re-run the pipeline
- In staging (ui-default), this shows the error page (unwanted behaviour)
- In this PR, this doesn't show the error page (wanted behaviour)

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1807

https://ui-agi-ui660.scp-staging.biomage.net/

#654 

#### Integration test branch
master

# Merge checklist
N/A

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional
- [X] Staging environment is unstaged before merging.
- [X] Photo of a cute animal attached to this PR.


src/components/ErrorBoundary.jsx
```diff
@@ -27,8 +16,6 @@ class ErrorBoundary extends React.Component {
   }
 
   render() {
-    if (this.state.hasError) return <Error />;
```
aerlaut, line 30: This is removed so that the error boundary still shows the UI in the case of an error.

src/components/data-processing/StatusIndicator.jsx
```diff
@@ -29,13 +28,11 @@ const StatusIndicator = (props) => {
     experimentId, allSteps, currentStep, completedSteps,
   } = props;
 
-  const {
-    status: { pipeline },
```
aerlaut, line 33: The `initialValue` for status is `null`, so this throws an error because we can't read `pipeline` from `undefined`. We do not want to hard code the `initialValue` of this in the UI because the `status` object is returned from the API.

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -77,7 +77,7 @@ const DataProcessingPage = ({ experimentId, experimentData }) => {
   const pipelineHadErrors = pipelineErrors.includes(pipelineStatusKey);
   const pipelineNotFinished = pipelineRunning || pipelineHadErrors;
 
-  const completedSteps = pipelineStatus?.completedSteps;
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx:80]: `pipelineStatus` is `null` if `backendStatus` is still loading. This evaluates to `null` which throws an error when `completedSteps` is used inside `StatusIndicator.jsx`.

src/__test__/components/ErrorBoundary.test.jsx
```diff
@@ -19,16 +20,28 @@ jest.mock('utils/postErrorToSlack');
...
+};
+
+const TestComponent = () => {
+  const [hasError, setHasError] = useState(false);
+
+  if (hasError) throw new Error('Error');
+
+  return (
+    <>
+      <button type='button' onClick={() => setHasError(true)}>Throw</button>
```
kafkasl, line 34: is there any reason to use this `setHasError(true)` instead of directly throwing the error like this:
`<button type='button' onClick={() => throw new Error('Error')}>Throw</button>`

src/components/ErrorBoundary.jsx
```diff
@@ -3,14 +3,14 @@ import React from 'react';
...
 import { connect } from 'react-redux';
-import Error from 'pages/_error';
 
 // Implementation of https://reactjs.org/docs/error-boundaries.html
 // Using React class because it's not yet supported for functional components
 // according to https://stackoverflow.com/a/68075800/1940886
 class ErrorBoundary extends React.Component {
   constructor(props) {
     super(props);
+    // eslint-disable-next-line react/no-unused-state
```
kafkasl: why disabling the linter instead of removing the state? afaict `this.state` is no longer used

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (kafkasl) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:80]: this is an error that's happening already in production and that you are fixing as part of the PR right? not directly related to the error boundary stuff

src/components/data-processing/StatusIndicator.jsx
```diff
@@ -118,15 +117,15 @@ const StatusIndicator = (props) => {
         <Text strong>
           Your data processing is
           {' '}
-          {statusIndicators[status].title}
+          {statusIndicators[status]?.title || 'loading'}
```
kafkasl: 1. Why are you adding `loading` as default here? the backend status default is set in `/Users/ahriman/repos/github.com/biomage-ltd/ui/src/redux/selectors/backendStatus/getBackendStatus.js`. If the status is `null` it might be because it has not run right, instead of loading?

2. Did you have a look at the status indicator with loading? the other status have specific formatting (like coloring the status name, etc... Wouldn't it be better to add another `statusIndicator` case instead of doing `?` conditional chaining?

src/utils/experimentUpdatesHandler.js
```diff
@@ -42,7 +42,7 @@ const experimentUpdatesHandler = (dispatch) => (experimentId, update) => {
 const onQCUpdate = (update, dispatch, experimentId) => {
   const { input, output } = update;
 
-  const processingConfigUpdate = output.config;
+  const processingConfigUpdate = output?.config;
```
kafkasl, line 45: when is the `output` null? is it a valid state or an error we should fix instead of using the conditional chaining?

src/__test__/components/ErrorBoundary.test.jsx
aerlaut: That's the difference between errors that can be caught by React error boundary and those that can't. As stated in [the docs](https://reactjs.org/docs/error-boundaries.html), React error boundary does not catch errors that originate in error handlers. It seems to only catch errors that originate during the generation of React views - i.e. errors arising in the body of functional components.

src/components/ErrorBoundary.jsx
aerlaut: If I remove this linter disable, the linter will complain that the constructor is useless and that it should be removed. We still need to call the parent's constructor to be able to set up the required functionalities.

I've removed this linter and the state, but need to add another linter command to stop the linter complaining of empty constructor.

src/utils/experimentUpdatesHandler.js
aerlaut: The output is `null` when moving from GEM2S to QC in `staging` or `production` due to a call to `assignPodToPipeline` returning a message to the UI. This can not be reproduced in local because we emulate step functions using lambdas instead of step functions. I had a discussion about this with Iva about a month ago. A screenshot of the message that causes the error can be seen [here](https://files.slack.com/files-pri/T0155DZFVM0-F033B39AWLX/image.png).

I can open a ticket to fix this.

src/components/data-processing/StatusIndicator.jsx
aerlaut: The value of `status` is `null` because the backend status data has not been loaded. We do not use a `switch` statement to give the status, but use an object keyed with status to access the details. I just noticed that we can use `null` as the key. Should I do this?

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx:80]: Yes, that's correct.

src/__test__/pages/_error.test.jsx
```diff
@@ -0,0 +1,47 @@
...
+delete window.location;
+window.location = { reload: jest.fn() };
+
+describe('ErrorPage', () => {
+  it('Renders properly without props', () => {
+    renderErrorPage();
+
+    expect(screen.getByText(/Sorry, something went wrong on our end./i)).toBeInTheDocument();
+
+    // There should be a feedback buttona and a reload button
```
kafkasl: typo in `buttona`

src/components/ErrorBoundary.jsx
```diff
@@ -3,21 +3,21 @@ import React from 'react';
...
 import { connect } from 'react-redux';
-import Error from 'pages/_error';
 
 // Implementation of https://reactjs.org/docs/error-boundaries.html
 // Using React class because it's not yet supported for functional components
 // according to https://stackoverflow.com/a/68075800/1940886
 class ErrorBoundary extends React.Component {
+  // The parent constructor needs to be called so that the
+  // proper functions are setup and views can be rendered
+  // eslint-disable-next-line no-useless-constructor
```
kafkasl, line 13: afaik https://eslint.org/docs/rules/no-useless-constructor this means that by default, when a class inherits another one, it will call `super(props)` if there is no constructor, so the linter seems to be correct and the call is unneeded. For reference the following works:

```
class A {
    constructor (value) {
        this.value = value
    }

    print() { 
        console.log(this.value)
    }
}

class B extends A {
}
const b = new B('a')

b.print() # prints 'a'
```

src/utils/experimentUpdatesHandler.js
kafkasl: I think it's ok to fix it like this. When we decided to notify (or rather not filter out) the `assignPodToPipeline` call, the reasoning was that maybe it would be useful for the UI to know that a pod was already assigned and display it to the user as another step or something like that. I don't think we will do that in the short-term but I don't think it's necessary to add extra code to filter them out. This looks like the best solution

src/components/data-processing/StatusIndicator.jsx
```diff
@@ -112,24 +113,57 @@ const StatusIndicator = (props) => {
...
-      </Paragraph>
-      <Paragraph>
-        {statusIndicators[status].description}
-      </Paragraph>
-    </Card>
-  );
+  const renderOverlay = () => {
+    const renderOverlayContent = () => {
+      if (loadingBackendStatus) { return (<Text>Loading run status...</Text>); }
+      if (errorLoadingBackendStatus) {
```
kafkasl, line 119: not sure how to test this in the staging like

kafkasl (approved):

src/components/ErrorBoundary.jsx
aerlaut: I tried removing the constructor, but it doesn't work. I found this blog which talks about this: https://overreacted.io/why-do-we-write-super-props/. Calling `super(props)` is required to set variables associated with `this`, which seems to be used internally by React to function.

src/components/data-processing/StatusIndicator.jsx
aerlaut: It can't be tested in staging - we have to setup our own conditions in the UI

PR #663, opened 2022-03-22
Patch and get endpoints experiment v2

cosa65:


src/redux/actions/experiments/reorderSamples.js
```diff
@@ -0,0 +1,74 @@
...
+        error: message,
+      },
+    });
+
+    pushNotificationMessage(
+      'error',
+      message,
+    );
+  }
+};
```
ivababukova, line 72: keep in mind that Pol is refactoring the bits around handling errors in actions and also what is returned by fetchAPI. If he merges first, these will need to be changed according to the new error handling implementation, as well as the error handling in `updateExperiment`.

src/redux/actions/experiments/updateExperiment.js
```diff
@@ -47,38 +52,55 @@ const updateExperiment = (
 
       const json = await response.json();
       throwIfRequestFailed(response, json, endUserMessages.ERROR_SAVING);
+    } else if (config.currentApiVersion === api.V2) {
+      if (_.isUndefined(experimentDiff.sampleIds)) {
```
ivababukova: why is this check `isUndefined` necessary?

ivababukova: I see that if `_.isUndefined` is false, `fetchAPI` won't be called, but `EXPERIMENTS_UPDATED` and `EXPERIMENTS_SAVED` actions will still be dispatched. Can you explain why is that?

cosa65: This handling wasn't well written, we discussed with iva and decided to throw an error if a dev tries to set `sampleIds`/`samplesOrder` through this action in v2 because we want this json array to be handled through specific endpoints (for reordering samples for example) or by the api directly as a side effect to whatever happens to the samples

src/redux/actions/experiments/reorderSamples.js
cosa65: I'll see when I try to merge and ask pol if I run into problems

src/redux/actions/projects/updateProject.js
```diff
@@ -22,7 +25,11 @@ const updateProject = (
   const newProject = mergeObjectWithArrays(currentProject, diff);
 
   try {
-    await dispatch(saveProject(projectUuid, newProject));
+    if (config.currentApiVersion === api.V1) {
+      await dispatch(saveProject(projectUuid, newProject));
+    } else if (config.currentApiVersion === api.V2) {
+      // Dont do any fetch, updating the experiment is enough
+    }
```
ivababukova, line 32: this if statement with empty body looks weird. Why did you add it? Is it just to make the code clearer?

ivababukova (approved):

PR #665, opened 2022-03-24
[BIOMAGE-1634] Refactored error handling for fetchAPI

kafkasl:
# Description

This PR refactors the way errors are thrown by the `fetchAPI` function and how they should be handled in the UI.

**How things were:**

* `fetchAPI` raises exceptions only for fetch operations
* HTTP status codes (4xx, 5xx) are set in the response but do not raise exceptions
* if an HTTP error code is present `throwIfRequestFailed` will throw a custom `ServerError` exception
* in the catch section, if the error **is not** a server error, assume it's a fetch error and display a console log
* after that, normal error handling usually including notifications to the user
Example of previous workflow
```
  try {
    const response = await fetchAPI(url);
    const status = await response.json();
    throwIfRequestFailed(response, status, endUserMessages.ERROR_FETCHING_STATUS);
    ...
  } catch (e) {
    if (!isServerError(e)) {
      console.error(`fetch ${url} error ${e.message}`);
    }
    // other error handling
    pushNotification(....)
```

**Issues**
* A lot of boilerplate code duplicated everywhere to parse response into json, throw if HTTP error code, and exception handling.
* Impossible to have a centralized place with all possible exceptions. E.g. show always the same message for a given status code like `UNAUTHORIZED` instead of the API error which won't be helpful to the user

**Solution**
* Group all error throwing and `json` parsing into `fetchAPI`
* `fetchAPI` throws a custom exception for HTTP status codes. If needed, it also parses the response into `json` format.
* Group all error handling into `handleError`
* `handleError` takes care of formatting the error message, notifying the user (if needed) and logging
* `handleError` returns the error message it generated so it can be stored in redux when needed.

Example:
```
  try {
    const data = await fetchAPI(URL);
    ...
  } catch (e) {
    const errorMessage = handleError(e, endUserMessages.ERROR_FETCHING_CELL_SETS);
    dispatch({
      type: CELL_SETS_ERROR,
      payload: { error: errorMessage },
    });
  }
```

**Import things to check**

* No duplicated error notifications
* Check the actual error text in error notifications / fields (coming from redux)

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1634

https://ui-randy-red-cuscus.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/__test__/components/GEM2SLoadingScreen.test.jsx
```diff
@@ -7,14 +7,14 @@ import {
 } from 'antd';
 import configureMockStore from 'redux-mock-store';
 import GEM2SLoadingScreen from '../../components/GEM2SLoadingScreen';
-import fetchAPI from '../../utils/fetchAPI';
+import fetchAPI from '../../utils/http/fetchAPI';
```
StefanBabukov: use absolute paths for these

src/__test__/pages/experiments/[experimentId]/plots-and-tables/frequency/index.test.jsx
```diff
@@ -32,15 +32,15 @@ describe('Frequency plots and tables index page', () => {
...
+  beforeEach(async () => {
     enableFetchMocks();
     fetchMock.resetMocks();
     fetchMock.doMock();
     fetchMock.mockIf(/.*/, mockAPI(mockApiResponses));
     storeState = makeStore();
 
     // getting the experiment info which is otherwise done by the SSR
-    storeState.dispatch({
+    await storeState.dispatch({
```
Comment (StefanBabukov) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/frequency/index.test.jsx:43]: I think you dont need to await dispatches

src/components/data-management/ProjectCard.jsx
```diff
@@ -58,10 +58,13 @@ const ProjectCard = (props) => {
   };
 
   const updateProjectName = (newName) => {
-    dispatch(updateProject(project.uuid, { name: newName.trim() }));
-
-    // Before multiple experiment support, use project name for analysis name
-    dispatch(updateExperiment(projectExperiment, { name: newName.trim() }));
+    dispatch(updateProject(project.uuid, { name: newName.trim() })).then((e) => {
```
StefanBabukov: If you not use .then and updateProject crashes will it still do updateExperiment too?

src/components/plots/DotPlot.jsx
```diff
@@ -28,7 +28,7 @@ const DotPlot = (props) => {
       return (
         <PlatformError
           error={cellSetsError}
-          reason={cellSetsError.message}
+          reason={cellSetsError}
```
StefanBabukov, line 31: Error and reason are the same?

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -460,11 +460,14 @@ const DataProcessingPage = ({ experimentId, experimentData }) => {
...
                         disabled={prefiltered}
                         data-testid='enableFilterButton'
-                        onClick={() => {
-                          dispatch(setQCStepEnabled(
-                            currentStep.key, !stepEnabled,
-                          ));
-                          dispatch(saveProcessingSettings(experimentId, currentStep.key));
+                        onClick={async () => {
+                          dispatch(saveProcessingSettings(experimentId, currentStep.key)).then(() => {
+                            if (processingConfig.meta.saveSettingsError === false) {
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx]: in this if instead of checking `processingConfig.meta.saveSettingsError` can you not do that check from what `then` returns - `blabla.then((res)=>if(!res.ok))` or something like that?

src/pages/experiments/[experimentId]/plots-and-tables/frequency/index.jsx
```diff
@@ -105,6 +105,7 @@ const FrequencyPlotPage = ({ experimentId }) => {
   ];
 
   if (!config) {
+    console.log('no config lcs');
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/frequency/index.jsx:108]: remove log

src/utils/http/errors/NotFoundApiError.js
```diff
@@ -0,0 +1,15 @@
+const httpStatusCodes = require('../httpStatusCodes');
+const BaseError = require('./ApiError');
+
+class NotFoundApiError extends BaseError {
```
StefanBabukov: Why does this extend BaseError not just Error?

ivababukova:
I opened data management while being in the admin account and got a couple of experiment undefined errors displayed as notifications, see this screenshot:
Screenshot 2022-04-05 at 08 52 20

ivababukova:
When I tried to access an experiment for which I don't have permissions, I get a 500 error:
Screenshot 2022-04-05 at 08 57 30
If I try to access the same experiment that I don't have access to (`c51cc377a855c7fa61b3608f26d5a881`) in ui-default, I correctly get `Analysis not found` error page.

src/utils/http/errors/NotFoundApiError.js
ivababukova: Why renaming `ApiError` to `BaseError`?

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
+import APIError from 'utils/http/errors/ApiError';
```
ivababukova: better be consistent with naming -- since the component is called `ApiError`, better change `APIError` to `ApiError`

src/utils/http/handleError.js
```diff
@@ -0,0 +1,64 @@
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from '../endUserMessages';
+
+const httpStatusCodes = require('./httpStatusCodes');
+
+const handleCodedErrors = (error, message, notifyUser) => {
```
ivababukova, line 10: this `handleError.js` file looks specific for errors -- why don't you move it under `http/errors`?

src/components/plots/DotPlot.jsx
kafkasl: original cellSetsError was an object. However, everywhere else in the code the errors stored in redux are the message (strings) so I unified them to all be string.

src/utils/http/fetchAPI.js
kafkasl: I renamed it the other way around

src/utils/http/errors/NotFoundApiError.js
kafkasl: not intentional, fixed

src/__test__/pages/experiments/[experimentId]/plots-and-tables/frequency/index.test.jsx
Comment (kafkasl) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/frequency/index.test.jsx:43]: yes you have to in the `beforeEach` if it's async because it was causing a race condition where this was run after the actual test

src/components/data-management/ProjectCard.jsx
kafkasl: actually the .then is completely unnecessary, without the if both would be triggered even when the first fails

kafkasl:
> c51cc377a855c7fa61b3608f26d5a881

fixed

src/utils/http/handleError.js
kafkasl: because `fetchAPI` & `handleError` should be used together so it makes sense for them to be at the same folder level IMO. Moreover, the `/errors` folder only contains actual errors types, not code

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (kafkasl) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx]: yeah that's a great idea I did that for the `ProjectCard` too

src/pages/experiments/[experimentId]/data-processing/index.jsx
```diff
@@ -460,11 +460,15 @@ const DataProcessingPage = ({ experimentId, experimentData }) => {
...
                         disabled={prefiltered}
                         data-testid='enableFilterButton'
-                        onClick={() => {
-                          dispatch(setQCStepEnabled(
-                            currentStep.key, !stepEnabled,
-                          ));
-                          dispatch(saveProcessingSettings(experimentId, currentStep.key));
+                        onClick={async () => {
+                          dispatch(saveProcessingSettings(experimentId, currentStep.key)).then((res) => {
+                            console.log('returning res: ', res);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/data-processing/index.jsx]: remove console.log

src/pages/experiments/[experimentId]/plots-and-tables/frequency/index.jsx
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/frequency/index.jsx]: remove console log

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -259,6 +259,7 @@ const MarkerHeatmap = ({ experimentId }) => {
     }
 
     if (errorMarkerGenes) {
+      console.log('error maker genes lcs ', errorMarkerGenes);
```
Comment (ivababukova) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: remove console log

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -22,24 +22,22 @@ const loadCellSets = (experimentId, forceReload = false) => async (dispatch, get
 
   const url = `/v1/experiments/${experimentId}/cellSets`;
   try {
-    const response = await fetchAPI(url);
-    const json = await response.json();
-
-    throwIfRequestFailed(response, json, endUserMessages.ERROR_FETCHING_CELL_SETS);
+    const data = await fetchAPI(url);
+    console.log('cell sets loaded lcs');
```
ivababukova: remove console log

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -22,24 +22,22 @@ const loadCellSets = (experimentId, forceReload = false) => async (dispatch, get
...
-        data: json.cellSets,
+        data: data.cellSets,
       },
     });
   } catch (e) {
-    if (!isServerError(e)) {
-      console.error(`fetch ${url} error ${e.message}`);
-    }
+    const errorMessage = handleError(e, endUserMessages.ERROR_FETCHING_CELL_SETS);
+    console.log(`cell sets failed to load lcs message ${errorMessage}`);
```
ivababukova: remove console log

src/redux/actions/componentConfig/fetchPlotDataWork.js
```diff
@@ -19,9 +19,12 @@ const fetchPlotDataWork = (experimentId, plotUuid, plotType) => async (dispatch,
       payload: { plotUuid },
     });
 
+    console.log('fetching work lcs');
```
ivababukova: remove console logs

src/redux/actions/componentConfig/loadPlotConfig.js
```diff
@@ -1,31 +1,32 @@
...
+        plotType,
+        plotData: data.plotData,
+        config,
+      },
+    });
+  } catch (e) {
+    console.log('error lcs ', e);
+    // load default plot config if it not found
+    if (e.statusCode === httpStatusCodes.NOT_FOUND) {
+      console.log('returning defaul initial config lcs');
```
ivababukova: remove console log

src/redux/actions/genes/loadGeneExpression.js
```diff
@@ -66,11 +66,14 @@ const loadGeneExpression = (
...
   try {
+    console.log('0 lcs fetch work');
     const data = await fetchWork(
       experimentId, body, getState, { timeout },
     );
+    console.log('0.1 lcs fech work');
 
     if (data[genesToFetch[0]]?.error) {
+      console.log('1 lcs error genes laready loaded');
       pushNotificationMessage('error', data[genesToFetch[0]].message);
```
ivababukova: remove console logs

src/redux/actions/pipeline/runGem2s.js
```diff
@@ -31,33 +31,29 @@ const runGem2s = (experimentId, paramsHash) => async (dispatch) => {
...
     dispatch(loadBackendStatus(experimentId));
   } catch (e) {
-    let { message } = e;
-    if (!isServerError(e)) {
-      console.error(`fetch ${url} error ${message}`);
-      message = endUserMessages.CONNECTION_ERROR;
-    }
+    let errorMessage = handleError(e, endUserMessages.ERROR_STARTING_PIPLELINE);
+
     console.log(`error run gem2s ${e}`);
```
ivababukova: remove console log

src/redux/actions/pipeline/runPipeline.js
```diff
@@ -81,31 +81,29 @@ const runPipeline = (experimentId) => async (dispatch, getState) => {
...
   } catch (e) {
-    let { message } = e;
-    if (!isServerError(e)) {
-      console.error(`fetch ${url} error ${message}`);
-      message = endUserMessages.CONNECTION_ERROR;
-    }
+    let errorMessage = handleError(e, endUserMessages.ERROR_STARTING_PIPLELINE);
 
-    console.log(`error run pipeline ${e}`);
+    console.log(`error run gem2s ${e}`);
```
ivababukova: remove console log

src/redux/actions/projects/updateMetadataTrack.js
```diff
@@ -40,8 +40,12 @@ const updateMetadataTrack = (
...
 
   try {
-    await dispatch(saveProject(projectUuid, newProject, false));
-    await dispatch(saveSamples(projectUuid, newSamples, false, false));
+    const notifyUser = false;
+    console.log('calling save project');
+    await dispatch(saveProject(projectUuid, newProject, false, notifyUser));
+    console.log('calling save samples');
+    await dispatch(saveSamples(projectUuid, newSamples, false, false, notifyUser));
+    console.log('call after save samples');
```
ivababukova: remove console log

src/redux/actions/samples/createSample.js
```diff
@@ -57,30 +56,26 @@ const createSample = (
...
       payload: { sample: newSample, experimentId },
     });
 
     return newSampleUuid;
   } catch (e) {
-    const { message } = e;
-    console.error(e);
+    const errorMessage = handleError(e, endUserMessages.ERROR_CREATING_SAMPLE);
 
+    console.log('failed to create sample');
```
ivababukova: remove console log

src/redux/actions/samples/saveSamples.js
```diff
@@ -60,30 +60,31 @@ const saveSamples = (
...
-    let { message } = e;
-    if (!isServerError(e)) {
-      console.error(`fetch ${url} error ${message}`);
-      message = endUserMessages.ERROR_SAVING;
-    }
+    // REVIEW is the payload in SAMPLES_ERROR needed?
+    // the exception handling & user notification will
+    // be called in the calling function
+    // ideally this next line should be removed
+    console.log(`catching exception and notify? ${notifyUser}`);
```
ivababukova: remove console log

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
...
+  // console.log(response);
+  // if the response is not OK, parse json an return API error message
+  if (!response.ok) {
+    console.log('reponse type ', response.headers.get('content-type'));
+    console.log('response not ok, ', response, typeof (response));
+    const data = isJson ? await response.json() : null;
+    // data.message & data.errors follow error formatting defined in:
+    // HTTPError.v1.yaml
+    console.log('response not ok, raising APIError, ', data);
+    throw new APIError(response.status, data?.message, data?.errors);
```
ivababukova: remove console logs

src/utils/http/handleError.js
```diff
@@ -0,0 +1,64 @@
...
+    pushNotificationMessage('error', `${message}`);
+  }
+
+  console.log(`${message} ${error}`);
+  return message;
+};
+
+const handleError = (error, message, notifyUser = true) => {
+  console.log('error lcs notify? :', notifyUser, error);
+  console.log();
```
ivababukova: remove console logs

src/utils/upload/processUpload.js
```diff
@@ -208,7 +206,10 @@ const processUpload = async (filesList, sampleType, samples, activeProjectUuid,
     };
   }, {});
 
+  console.log('samplesMap', samplesMap);
+  console.log(samplesMap);
   Object.entries(samplesMap).forEach(async ([name, sample]) => {
+    console.log(`lcs sample ${name}`, sample);
     // Create sample if not exists.
```
ivababukova: remove console log

src/utils/work/fetchWork.js
```diff
@@ -67,8 +67,10 @@ const fetchGeneExpressionWork = async (
     experimentId, missingGenesBody, qcPipelineStartDate, extras, cacheUniquenessKey,
   });
 
+  console.log('lcs fetchGeneExpressionWork 1');
   // Then, we may be able to find this in S3.
   let response = await seekFromS3(ETag, experimentId);
+  console.log('lcs fetchGeneExpressionWork 2');
```
ivababukova: remove console logs

src/__test__/components/data-exploration/cell-sets-tool/HideButton.test.jsx
```diff
@@ -4,9 +4,9 @@ import { Provider } from 'react-redux';
 import configureMockStore from 'redux-mock-store';
 import thunk from 'redux-thunk';
 
+import { CELL_SETS_HIDE, CELL_SETS_UNHIDE } from 'redux/actionTypes/cellSets';
+import initialState from 'redux/reducers/cellSets/initialState';
 import HideButton from '../../../../components/data-exploration/cell-sets-tool/HideButton';
```
ivababukova: Can you also fix the import of `HideButton`? It's the last one left in this file

src/__test__/components/data-exploration/gene-list-tool/GeneListTool.test.jsx
```diff
@@ -8,20 +8,20 @@ import thunk from 'redux-thunk';
 import _ from 'lodash';
 import { Empty } from 'antd';
 import waitForActions from 'redux-mock-store-await-actions';
+import { GENES_PROPERTIES_LOADING, GENES_PROPERTIES_LOADED_PAGINATED } from 'redux/actionTypes/genes';
+import { fetchWork } from 'utils/work/fetchWork';
 import GeneListTool from '../../../../components/data-exploration/gene-list-tool/GeneListTool';
-import { fetchWork } from '../../../../utils/work/fetchWork';
 
 import Loader from '../../../../components/Loader';
```
ivababukova: can you also fix those 2 imports? They are the last ones left in this file

src/__test__/components/data-exploration/generic-gene-table/ComponentActions.test.jsx
```diff
@@ -6,17 +6,16 @@ import waitForActions from 'redux-mock-store-await-actions';
 import thunk from 'redux-thunk';
 import preloadAll from 'jest-next-dynamic';
 import configureMockStore from 'redux-mock-store';
+import { GENES_EXPRESSION_LOADING, GENES_EXPRESSION_LOADED } from 'redux/actionTypes/genes';
+import { fetchWork } from 'utils/work/fetchWork';
 import ComponentActions from '../../../../components/data-exploration/generic-gene-table/ComponentActions';
```
ivababukova: can you also fix the import of `ComponentActions`?

src/__test__/components/data-management/ProjectDetails.test.jsx
```diff
@@ -8,16 +8,16 @@ import userEvent from '@testing-library/user-event';
...
+import rootReducer from 'redux/reducers/index';
+import * as createMetadataTrack from 'redux/actions/projects/createMetadataTrack';
+import initialProjectState, { projectTemplate } from 'redux/reducers/projects/initialState';
+import initialSamplesState, { sampleTemplate } from 'redux/reducers/samples/initialState';
+import initialExperimentsState from 'redux/reducers/experiments/initialState';
+import initialExperimentSettingsState from 'redux/reducers/experimentSettings/initialState';
+import { initialExperimentBackendStatus } from 'redux/reducers/backendStatus/initialState';
+import PipelineStatus from 'utils/pipelineStatusValues';
+import UploadStatus from 'utils/upload/UploadStatus';
 import ProjectDetails from '../../../components/data-management/ProjectDetails';
```
ivababukova: could you fix the import for `ProjectDetails` too?

src/__test__/components/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -140,14 +146,17 @@ describe('Marker heatmap plot', () => {
   });
 
   it('loads marker genes on specifying new nunmber of genes per cluster', async () => {
+    console.log('loads test');
```
ivababukova: remove this

src/__test__/components/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -176,9 +187,11 @@ describe('Marker heatmap plot', () => {
       .mockImplementationOnce((ETag) => mockWorkerResponses[ETag])
       // 2nd load
       .mockImplementationOnce(() => null)
-      .mockImplementationOnce((ETag) => mockWorkerResponses[ETag]);
+      .mockImplementation((ETag) => mockWorkerResponses[ETag]);
 
     await renderHeatmapPage(storeState);
+    // await renderHeatmapPage(storeState);
+    // await renderHeatmapPage(storeState);
```
ivababukova: can you remove this commented code?

src/__test__/redux/reducers/componentConfig/fetchPlotDataWork.test.js
```diff
@@ -33,14 +33,13 @@ const plotUuid = 'DotPlotMain';
...
 describe('fetchPlotDataWork', () => {
-  beforeEach(() => {
+  beforeEach(async () => {
     jest.clearAllMocks();
     fetchMock.resetMocks();
 
     fetchMock.mockResponse(() => Promise.resolve({ status: 404, body: JSON.stringify('Plot config not found') }));
 
     testStore = makeStore();
-    testStore.dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
```
ivababukova, line 43: why do we not need to load the plot config anymore?

src/components/data-management/ProjectCard.jsx
```diff
@@ -58,10 +57,14 @@ const ProjectCard = (props) => {
   };
 
   const updateProjectName = (newName) => {
-    dispatch(updateProject(project.uuid, { name: newName.trim() }));
-
-    // Before multiple experiment support, use project name for analysis name
-    dispatch(updateExperiment(projectExperiment, { name: newName.trim() }));
+    // dispatch(updateProject(project.uuid, { name: newName.trim() }))
+    dispatch(updateProject(project.uuid, { name: newName.trim() })).then((res) => {
+      // console.log('Almost finished ', e);
```
ivababukova: can you remove the commented out code

src/components/data-management/ProjectCard.jsx
```diff
@@ -58,10 +57,14 @@ const ProjectCard = (props) => {
...
+      if (res) {
+      // Before multiple experiment support, use project name for analysis name
```
ivababukova: you can remove `Before multiple experiment support` from the comment, that's no longer relevant. Just keep `Use project name for analysis name`

src/pages/_app.jsx
```diff
@@ -199,31 +199,21 @@ WrappedApp.getInitialProps = async ({ Component, ctx }) => {
...
-      if (res && e.payload.status) {
-        res.statusCode = e.payload.status;
-      } else {
-        res.statusCode = 500;
-      }
-
-      return { pageProps: { ...pageProps, ...results, httpError: e.payload.status || true } };
-    }
-    console.error('Error in WrappedApp.getInitialProps', e);
+    res.statusCode = e.statusCode;
```
ivababukova: what happened with the code that handles "User is not authenticated" error?

ivababukova: Right now, if I try to access the platform while I am not logged in, it gives "it's not you it's us" error

src/redux/actions/experimentSettings/processingConfig/saveProcessingSettings.js
```diff
@@ -27,32 +24,27 @@ const saveProcessingSettings = (experimentId, settingName) => async (dispatch, g
...
       type: EXPERIMENT_SETTINGS_PROCESSING_ERROR,
       payload: {
-        error: endUserMessages.ERROR_SAVING,
-        errorType: errorTypes.SAVE_PROCESSING_SETTINGS,
+        error: errorMessage,
       },
     });
+
+    // throw e;
+    Promise.reject(errorMessage);
```
ivababukova: do we need to do a promise reject here? I notice that none of the other actions do `Promise.reject` if an error happens.

ivababukova: or is this done to enable the React components from which this `saveProcessingSettings` action is dispatched to detect if there were errors (the `.then((e) => if(e) {don't continue with second action})` block)?

src/redux/actions/genes/loadMarkerGenes.js
```diff
@@ -24,9 +26,11 @@ const loadMarkerGenes = (
   try {
     const timeout = getTimeoutForWorkerTask(getState(), 'MarkerHeatmap');
 
+    console.log('1. lcs load marker genes');
     const data = await fetchWork(experimentId, body, getState, { timeout });
-
+    console.log('marker genes data loaded ', data);
     const { data: markerGeneExpressions, order } = data;
+    console.log('marker genes data unst');
```
ivababukova: these console logs need removing

src/redux/actions/pipeline/runGem2s.js
```diff
@@ -31,33 +31,29 @@ const runGem2s = (experimentId, paramsHash) => async (dispatch) => {
...
     // temporarily give the user more info if the error is permission denied
-    if (message.includes('does not have access to experiment')) {
-      message += ' Refresh the page to continue with your analysis.';
+    if (errorMessage.includes('does not have access to experiment')) {
+      errorMessage += ' Refresh the page to continue with your analysis.';
```
ivababukova, line 47: do you know why we need to do ask the user to refresh the page?

ivababukova: It's also in the `runPipeline.js` action

src/redux/actions/projects/createProject.js
```diff
@@ -63,28 +61,20 @@ const createProject = (
...
-          error: message,
+          error: errorMessage,
         },
       });
 
-      pushNotificationMessage('error', message);
-      return Promise.reject(message);
+      // REVIEW I assume that we need to reject this so the UI component
+      // NewProjectModal will stop the execution and won't run next lin
+      return Promise.reject(errorMessage);
```
ivababukova: My understanding is that it works this way indeed, but we need to handle the case of rejected promise in `NewProjectModal`. Also, you probably want to remove the `return`, just keep `Promise.reject` - based on my understanding, we shouldn't be returning from within Redux actions.

ivababukova: Now I am trying to find out why we shouldn't be returning from within Redux actions, but I can't find anything.

src/redux/actions/projects/saveProject.js
```diff
@@ -36,30 +36,28 @@ const saveProject = (
...
+        error: errorMessage,
       },
     });
-    pushNotificationMessage('error', message);
-    return Promise.reject(message);
+
+    // TODO I think this will lead to duplicated error messages but it might be needed.
+    // return Promise.reject(errorMessage);
+    throw e;
+    // throw new Error(errorMessage);
```
ivababukova: why don't you just do Promise.reject?

src/redux/actions/samples/createSample.js
```diff
@@ -57,30 +56,26 @@ const createSample = (
...
+        error: errorMessage,
       },
     });
 
-    pushNotificationMessage('error', endUserMessages.ERROR_CREATING_SAMPLE);
-
-    throw new Error(message);
+    // throw again the error so `processUpload` won't upload the sample
+    // Promise.reject(errorMessage);
+    throw new Error(errorMessage);
```
ivababukova: This seems to be popping up a lot - error handling in Redux. I read more about the recommended way to do error handling in Redux action creators. In both here: https://stackoverflow.com/questions/58266418/correct-way-of-error-handling-in-react-redux and https://exportknowledge.com/article/error-handling-with-react-and-redux people actually use the error field in redux state. 
So when an error happens, all that this file does is to dispatch an "ERROR" action type and it doesn't raise and return anything. Then all components that care about that error state, just check the error in the store. Now I remember that we followed this flow with Data Exploration in the very beginning and it worked very well. I am going to ask someone that knows a lot about front end about this too and then I will tell you what they recommend.

src/redux/actions/pipeline/runGem2s.js
kafkasl: if the user tries to run the pipeline without permissions it will show the "It's not you it's us" page. In that page there's no way to go back to data processing (which they can indeed explore but not run). It's bad UX as it might not be immediately clear what to do. Ideally, we would prevent the user from trying to run it, but for the time being Vicky accepted telling the user to refresh the page (so they can once again explore data processing)

src/__test__/redux/reducers/componentConfig/fetchPlotDataWork.test.js
kafkasl: this was never needed. It just load some plot config, but the test actually overwrite that config so the initial state is never used. Moreover, as it lacks an `await` before it, this was running out of order sometimes, and the initialization was happening after the test so the final state was wrong. It was a pain to debug this and it's unnecessary code that probably people copies over from other tests

src/components/data-management/ProjectCard.jsx
kafkasl: I realized why, the action will always succeed but it's synchronous, meaning that the if will always run after the action which is ok

src/pages/experiments/[experimentId]/data-processing/index.jsx
Comment (kafkasl) [reply src/pages/experiments/[experimentId]/data-processing/index.jsx]: I reviewed this with Iva and it turns out returning things in actions is not the correct way to go. I added back the `if` way because it's the recommended way of handling things in react/redux https://stackoverflow.com/a/58299220 https://exportknowledge.com/article/error-handling-with-react

src/__test__/components/ReferralButton.test.jsx
```diff
@@ -6,9 +6,9 @@ import userEvent from '@testing-library/user-event';
 import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
 import '@aws-amplify/auth';
 
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from 'utils/endUserMessages';
 import ReferralButton from '../../components/ReferralButton';
```
aerlaut: Can you also fix the import of the `ReferralButton`?

src/redux/actions/projects/createProject.js
kafkasl: refactored so that it uses the proper way

src/redux/actions/samples/createSample.js
kafkasl: agreed, but this one should be done in another refactor

src/redux/actions/projects/saveProject.js
kafkasl: because this action is called from another action which is the one handling the exception, and for that it needs the exception to be re-thrown as it is

src/__test__/components/ReferralButton.test.jsx
kafkasl: ok, for the record I just used global search and replace for `/redux` and `/utils` so I didn't go file by file, that's why it's not changed

src/__test__/components/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -176,9 +187,11 @@ describe('Marker heatmap plot', () => {
       .mockImplementationOnce((ETag) => mockWorkerResponses[ETag])
       // 2nd load
       .mockImplementationOnce(() => null)
-      .mockImplementationOnce((ETag) => mockWorkerResponses[ETag]);
+      .mockImplementation((ETag) => mockWorkerResponses[ETag]);
```
cosa65, line 189: Why is the last a `mockImplementation` instead of `mockImplementationOnce`?

src/__test__/redux/actions/projects/createProject.test.js
```diff
@@ -110,16 +111,14 @@ describe('createProject action', () => {
   it('Shows error message when there was a fetch error', async () => {
     config.currentApiVersion = api.V1;
 
-    const fetchErrorMessage = 'someFetchError';
+    const fetchErrorMessage = endUserMessages.ERROR_CREATING_PROJECT;
 
     fetchMock.mockResponse(JSON.stringify({ message: fetchErrorMessage }), { url: 'mockedUrl', status: 400 });
```
cosa65, line 114: `endUserMessages.ERROR_CREATING_PROJECT` should be what is being generated in the UI, not what is being returned from the api, right?

src/__test__/redux/actions/projects/saveProject.test.js
```diff
@@ -71,7 +71,7 @@ describe('saveProject action', () => {
     try {
       await store.dispatch(saveProject(mockProject.uuid, mockProject));
     } catch (e) {
-      expect(e).toEqual(errorMsg);
+      expect(e.statusCode).toEqual(400);
     }
```
aerlaut, line 75: I think this should be

```
await expect(
    store.dispatch(saveProject(mockProject.uuid, mockProject))
).rejects.toThrow(endUserMessages.ERROR_SAVING)
```

src/__test__/redux/actions/samples/saveSamples.test.js
```diff
@@ -151,7 +151,7 @@ describe('saveSamples action', () => {
     try {
       await store.dispatch(saveSamples(mockprojectUuid, updatedSample1));
     } catch (e) {
-      expect(e).toEqual(errorMsg);
+      expect(e.statusCode).toEqual(400);
     }
```
aerlaut, line 155: I think this should be
```
await expect(
   store.dispatch(saveSamples(mockprojectUuid, updatedSample1))
).rejects.toThrow(endUserMessages.ERROR_SAVING)
```

src/redux/actions/cellSets/createCellSet.js
```diff
@@ -78,18 +78,12 @@ const createCellSet = (experimentId, name, color, cellIdsSet) => async (dispatch
           createCellSetJsonMerger(dataForUpload, 'scratchpad'),
         ),
       },
+      false,
```
cosa65: This false is being passed as `extras` (the 3rd parameter for `fetchApi` is `extras`, an object)

src/components/data-management/ProjectCard.jsx
```diff
@@ -59,9 +58,11 @@ const ProjectCard = (props) => {
 
   const updateProjectName = (newName) => {
     dispatch(updateProject(project.uuid, { name: newName.trim() }));
-
-    // Before multiple experiment support, use project name for analysis name
-    dispatch(updateExperiment(projectExperiment, { name: newName.trim() }));
+    console.log('projects.meta: ', projects.meta);
```
aerlaut: Don't forget to remove this console.log

src/components/data-management/ProjectCard.jsx
```diff
@@ -59,9 +58,11 @@ const ProjectCard = (props) => {
...
+    if (projects.meta.error === false) {
+      console.log('go for it');
```
aerlaut: Don't forget to remove this console.log

src/components/plots/ContinuousEmbeddingPlot.jsx
```diff
@@ -2,14 +2,14 @@ import React, { useState, useEffect } from 'react';
...
+import { loadProcessingSettings } from 'redux/actions/experimentSettings';
+import { getCellSets } from 'redux/selectors';
+import { generateSpec, generateData } from 'utils/plotSpecs/generateEmbeddingContinuousSpec';
 import PlatformError from '../PlatformError';
-import { generateSpec, generateData } from '../../utils/plotSpecs/generateEmbeddingContinuousSpec';
-import { loadCellSets } from '../../redux/actions/cellSets';
-import { loadEmbedding } from '../../redux/actions/embedding';
-import { loadProcessingSettings } from '../../redux/actions/experimentSettings';
-import { getCellSets } from '../../redux/selectors';
+import Loader from '../Loader';
```
aerlaut, line 12: Can you change these 2 imports to use the import format that we use

src/components/plots/DoubletScoreHistogram.jsx
```diff
@@ -2,8 +2,8 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Vega } from 'react-vega';
 
+import generateSpec from 'utils/plotSpecs/generateDoubletScoreHistogram';
 import EmptyPlot from './helpers/EmptyPlot';
```
aerlaut, line 6: Change this import to use the format that we use

src/components/plots/ElbowPlot.jsx
```diff
@@ -1,8 +1,8 @@
 import React, { useState, useEffect } from 'react';
 import { Vega } from 'react-vega';
 import PropTypes from 'prop-types';
+import { generateSpec } from 'utils/plotSpecs/generateElbowSpec';
 import EmptyPlot from './helpers/EmptyPlot';
```
aerlaut, line 5: Change this import to use the format that we use

src/redux/store.js
```diff
@@ -11,15 +11,15 @@ enableMapSet();
...
-  }
+  // if (process.env.K8S_ENV !== 'test') {
+  //   // eslint-disable-next-line import/no-extraneous-dependencies
+  //   const { createLogger } = require('redux-logger');
+
+  //   // do not log server-side redux actions
+  //   middleware.push(createLogger({
+  //     predicate: () => typeof window !== 'undefined',
+  //   }));
+  // }
```
cosa65: What is this code? Is it adding the logger of redux actions?

src/components/plots/FeaturesVsUMIsScatterplot.jsx
```diff
@@ -2,9 +2,9 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Vega } from 'react-vega';
 
+import generateSpec from 'utils/plotSpecs/generateFeaturesVsUMIsScatterplot';
 import EmptyPlot from './helpers/EmptyPlot';
 import transformOldFeaturesVsUMIsPlotData from './helpers/transformOldFeaturesVsUMIsPlotData';
```
aerlaut, line 7: Change this import to use the format that we use

src/components/plots/MitochondrialFractionHistogram.jsx
```diff
@@ -2,8 +2,8 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Vega } from 'react-vega';
 
+import generateSpec from 'utils/plotSpecs/generateMitochondrialFractionHistogram';
 import EmptyPlot from './helpers/EmptyPlot';
```
aerlaut, line 6: Change this import to use the format that we use

src/components/plots/MitochondrialFractionScatterplot.jsx
```diff
@@ -2,8 +2,8 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Vega } from 'react-vega';
 
+import generateSpec from 'utils/plotSpecs/generateMitochondrialFractionScatterplot';
 import EmptyPlot from './helpers/EmptyPlot';
```
aerlaut, line 6: Change this import to use the format that we use

src/utils/http/errors/FetchError.js
```diff
@@ -0,0 +1,12 @@
+class FetchError extends Error {
+  constructor(error) {
+    super(error);
+
+    Object.setPrototypeOf(this, new.target.prototype);
+    this.name = 'Fetch Error';
+    this.isOperational = true;
```
cosa65: What does `isOperational` mean?

src/utils/http/errors/FetchError.js
```diff
@@ -0,0 +1,12 @@
+class FetchError extends Error {
+  constructor(error) {
+    super(error);
+
+    Object.setPrototypeOf(this, new.target.prototype);
```
cosa65: What does this do?

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
...
+    throw new FetchError(e);
+  }
+
+  const isJson = response.headers.get('content-type')?.includes('application/json');
+
+  if (response.ok) {
+    // only parse response into JSON if needed because
+    // most patch / put reqs do not need a response
+    if (returnValue) {
+      return response.json();
```
aerlaut, line 34: This assumes all returned values are JSON right?

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
...
+  try {
+    response = await fetch(url, parameters);
+  } catch (e) {
+    // wrap fetch errors in custom error
+    throw new FetchError(e);
+  }
+
+  const isJson = response.headers.get('content-type')?.includes('application/json');
+
+  if (response.ok) {
```
aerlaut: Do we still need this clause? At the end of the function you also return `response.json()`

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
...
+    // only parse response into JSON if needed because
+    // most patch / put reqs do not need a response
+    if (returnValue) {
+      return response.json();
+    }
+    return;
+  }
```
cosa65: Remove this, there is the same handling right before this function ends

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
...
+    const data = isJson ? await response.json() : null;
+    // data.message & data.errors follow error formatting defined in:
+    // HTTPError.v1.yaml
+    console.log('response not ok, raising APIError, ', data);
+    throw new APIError(response.status, data?.message, data?.errors);
+  }
+
+  // only parse response into JSON if needed because
+  // most patch / put reqs do not need a response
+  if (returnValue) return response.json();
```
cosa65: The previous fetchApi didn't perform the parsing, why is it better to parse inside fetchApi instead of leaving the parsing to the caller? Removing this .json() also means we can get rid of the returnValue parameter.

Also, I assume we are returning json always from fetchApi, but should we consider the possibility of us having to parse different types of files?

src/utils/http/handleError.js
```diff
@@ -0,0 +1,64 @@
...
+
+const handleCodedErrors = (error, message, notifyUser) => {
+  console.log(`elcs ${message}, ${error.userMessage}`);
+  let errorMessage = [message, error.userMessage].filter((x) => x).join(' ');
+
+  // we might want to override some errors to make them more user friendly
+  // like the UNAUTHORIZED
+  switch (error.statusCode) {
+    case httpStatusCodes.BAD_REQUEST:
+    // message = message
```
cosa65: Should these commented out lines be removed?

src/utils/http/errors/NotFoundApiError.js
```diff
@@ -0,0 +1,15 @@
+const httpStatusCodes = require('../httpStatusCodes');
```
aerlaut: Why are you using `require` instead of `import`?

src/utils/http/errors/APIError.js
```diff
@@ -0,0 +1,16 @@
+const http = require('http');
```
aerlaut: Why are you using require instead of import?

src/utils/http/fetchAPI.js
```diff
@@ -0,0 +1,56 @@
+import APIError from 'utils/http/errors/APIError';
+import getApiEndpoint from '../apiEndpoint';
+import getAuthJWT from '../getAuthJWT';
+import FetchError from './errors/FetchError';
```
aerlaut: Please change these to use the common import path format

src/utils/work/seekWorkResponse.js
```diff
@@ -1,24 +1,35 @@
...
+    response = await fetchAPI(`/v1/workResults/${experimentId}/${ETag}`);
+  } catch (e) {
+    console.log('raised error lcs', e);
+    if (e.statusCode === httpStatusCodes.NOT_FOUND) {
+      return null;
+    }
+    // REVIEW: do we want to show a notification to the user
+    // when we can't fetch from s3? (and it's not a 404)
+    // handleError(e);
+    // throwResponseError(response);
```
cosa65: although I think some kind of feedback on an error could come back, I would leave it to the action creators that indirectly triggered this function. (I wouldn't handle it here if that is the question)

src/utils/http/handleError.js
```diff
@@ -0,0 +1,64 @@
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from '../endUserMessages';
+
+const httpStatusCodes = require('./httpStatusCodes');
```
aerlaut: Why are you using `require` here?

src/utils/http/handleError.js
```diff
@@ -0,0 +1,64 @@
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from '../endUserMessages';
```
aerlaut: Please change the path to our normal `import` path format

src/utils/ssr/getEnvironmentInfo.js
```diff
@@ -1,5 +1,5 @@
+import loadEnvironment from 'redux/actions/networkResources/loadEnvironment';
 import { ssrGetCurrentEnvironment } from '../environment';
```
aerlaut: can you also change the path of this one to the import path format that we use

src/utils/work/seekWorkResponse.js
```diff
@@ -1,24 +1,35 @@
...
-    throwResponseError(response);
+  let response;
+  try {
+    response = await fetchAPI(`/v1/workResults/${experimentId}/${ETag}`);
+  } catch (e) {
+    console.log('raised error lcs', e);
+    if (e.statusCode === httpStatusCodes.NOT_FOUND) {
+      return null;
+    }
+    // REVIEW: do we want to show a notification to the user
```
aerlaut: I think this is an internal process, so we shouldn't show the 404 to the user.

src/__test__/redux/actions/projects/saveProject.test.js
kafkasl: no it should not, because all the tests rely on checking the actual error message which is pretty bad because it makes the code very hard to change. In this test I preferred to check the error thrown

src/components/plots/DoubletScoreHistogram.jsx
kafkasl: what format?

src/utils/http/fetchAPI.js
kafkasl: it's used only in the case of an error. Ideally, it should be used for both cases, however most of the tests do not set that field so using the `fetchAPI` with those will fail. In the general situation, when we want `fetchAPI` to return something it's safe to assume it's JSON. But when there's an error there might not be json which would lead to `response.json()` to fail, and throw an exception hiding the real exception. That's why in that case I only deserialize json when I'm sure it won't fail.

src/utils/http/errors/APIError.js
kafkasl: just copied from somewhere, changed it

src/__test__/components/plots-and-tables/marker-heatmap/index.test.jsx
kafkasl: wrong change, reverted it

src/__test__/redux/actions/projects/createProject.test.js
kafkasl: it doesn't really matter as it won't be checked. I've changed it back though

src/redux/actions/cellSets/createCellSet.js
kafkasl: That's a very good point. What do you think it's best:
1. Add an empty object to each of the function calls that need to set the 4th parameter
`fetchAPI(URL, params, {}, false)`
2. Change `fetchApi` so that it receives just an object and can do restructuring of the parameters
`fetchAPI({url, params, returnValue: false})`

src/redux/store.js
kafkasl: forgot to uncomment it. it's the code that prints all actions and redux store states for debugging. It's just too annoying when trying to debug the tests

src/utils/http/fetchAPI.js
kafkasl: it's better to do inside because **all** our calls do it so we avoid a lot of code. Moreover, I haven't done it here but it's possible to wrap those `json` calls in a `try/catch` to return a custom error like `JSONError` which then will also be handled by `handleError` abstracting away all the changes / handling of exceptions from the caller.

I don't think we should consider different types of files if we are not using them.

src/utils/http/errors/FetchError.js
kafkasl: https://www.joyent.com/node-js/production/design/errors#:~:text=two%20broad%20categories%3A-,3,-Operational%20errors%20represent

I just added some parameters from examples I found, but it's not used.

src/utils/http/errors/FetchError.js
kafkasl: I have no idea and I haven't found the link where I copied it from. Removed

src/utils/http/handleError.js
kafkasl: changed

src/utils/work/seekWorkResponse.js
kafkasl: ok

src/redux/actions/cellSets/createCellSet.js
cosa65: An empty object is probably more coherent with how we work with default parameters usually

cosa65: I wish we could name the optional parameters we are defining like in python :(

src/redux/store.js
cosa65: there was a change from !== 'production' to !== 'test', why was that?

Also unrelated to your work, but I am kind of confused, we are getting the redux logs in production too so the `!== 'production'` didn't work?

src/utils/http/errors/FetchError.js
cosa65: if its not used could we remove it?

kafkasl: done

src/redux/store.js
kafkasl: the change was unintentional. It didn't work because there's no `process.env.K8S_ENV` in the browser. I've fixed it to use `ssrGetCurrentEnvironment`

src/redux/actions/cellSets/createCellSet.js
kafkasl: well the nodejs way of doing named parameters is passing the object that's why I suggested it. I feel having to add empty parameters is much more error-prone in addition to make it easier to introduce bugs

src/__test__/redux/actions/samples/saveSamples.test.js
kafkasl: checking the status code is more reliable, and easier to change than error messages

src/utils/http/fetchAPI.js
cosa65: I think perhaps the difference here is that I considered `fetchApi` as a wrapper for `fetch` that has some handling added, the final result when eveything works well is the same one as for `fetch`.

That's why having a `response.json()` inside kind of seemed weird to me, especially since that addition causes us to have to add a `returnValue` parameter to check whether to do it or not (because we are actually not doing `.json()` in all our calls after your refactoring, for example in `createCellSets`).

That's the main points I had, but I think if we keep it like this it should be ok as well. 

If we keep doing `response.json()` the only thing I would recommend perhaps is doing `return await reponse.json()` because of [this](https://stackoverflow.com/questions/44806135/why-no-return-await-vs-const-x-await/44806230#44806230).
Also disable the no-return-await rule from eslint, we already did that for the api.

src/pages/_app.jsx
kafkasl: fixed it. now the `getExperimentInfo` throws an `APIError` 401 so it's hanlded correctly

src/utils/http/handleError.js
```diff
@@ -0,0 +1,60 @@
...
+import pushNotificationMessage from 'utils/pushNotificationMessage';
+import endUserMessages from 'utils/endUserMessages';
+
+import httpStatusCodes from 'utils/http/httpStatusCodes';
+
+const env = ssrGetCurrentEnvironment();
+
+const handleCodedErrors = (error, message, notifyUser) => {
+  let errorMessage = message;
+  // let errorMessage = [message, error.userMessage].filter((x) => x).join(' ');
```
cosa65, line 12: Remove commented out code

cosa65 (dismissed):

cosa65 (approved):

PR #671, opened 2022-03-30
[BIOMAGE-1568] - Refactor Plot Container

aerlaut:
# Description
#### Problem
Currently, the "reset" button for plots is located in the header beside the log-out button. This is not good UI as it is confusing for the user - is the "reset" associated with the experiment? or the plot? This PR solves that issue.

#### Solution
- Move the reset button from the plot header into the plot container, so it's clearer that "Reset" means the plot.
- Remove the ability of the plot to be minimized.
- Better keys of plot styling components so some options are open by default.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1568

https://ui-agi-snazzy-havanese.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


ivababukova:
Hi @aerlaut, I took a look, a couple of things to fix in terms of UX:
1. On the top right side of the plot box, you have the information icon and the Reset button is on its right side: 
Screenshot 2022-04-05 at 08 37 49
Better swap them, have the reset button be before the information icon -- right now this order looks odd to me.
2. My opinion is that having the `accordion` is actually better - otherwise people have to go and close manually all these open dropdowns.
3. I noticed that after clicking the `Reset` button, settings under `Select data` don't get reset - both the data displayed on the plot and the menu settings stay the same. Can you fix it?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -244,7 +244,7 @@ const DotPlotPage = (props) => {
 
   const renderExtraPanels = () => (
     <>
-      <Panel header='Gene selection' key='gene-selection'>
+      <Panel header='Gene selection' key='Gene selection'>
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: For these `key=` it seems we are either using `-` separted words syntax or camelCase, if we are removing the `-` syntax I would go for camelCase instead, not this new different space-separated words syntax

src/components/plots/PlotContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+  <Button
+    key='reset'
+    type='primary'
+    size='small'
+    onClick={onClickReset}
+    disabled={disabled}
+  >
+    Reset
+  </Button>
+);
```
cosa65: Do we need this to be a new component? It doesn't seem to have enough custom functionality to warrant a new component

src/components/plots/PlotContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+  const checkConfigEquality = (currentConfig, initialConfig) => {
+    const ignoredFields = {
+      // config fields that are set dynamically on component render
+      // should not be compared to their initial values
+      embeddingContinuous: ['shownGene'],
+      violin: ['shownGene', 'title'],
+      markerHeatmap: ['selectedGenes'],
+      DotPlot: ['selectedGenes'],
+    };
```
cosa65, line 35: I think this has the potential to break very fast whenever we add more config, we'll forget we need to update this, maybe we could just add another property (called `customFields` or something like that), ignore only that property and store all of these cusotm properties within that one?

cosa65: Or any option that deals with this in a more automatic way would probably be good

src/components/plots/PlotContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+    const areAllValuesTheSame = Object.keys(initialConfig).every((key) => {
+      // By pass plot data because we want to compare settings not data
+      if (key === 'plotData') return true;
+      if (ignoredFields[plotType]?.includes(key)) return true;
+      if (typeof currentConfig[key] === 'object') {
```
cosa65, line 41: Do you whether it is better to use `instanceOf` or `typeOf`?

src/components/plots/PlotContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+    const ignoredFields = {
+      // config fields that are set dynamically on component render
+      // should not be compared to their initial values
+      embeddingContinuous: ['shownGene'],
+      violin: ['shownGene', 'title'],
+      markerHeatmap: ['selectedGenes'],
+      DotPlot: ['selectedGenes'],
+    };
+
+    const areAllValuesTheSame = Object.keys(initialConfig).every((key) => {
```
cosa65, line 37: I think it might be better to look into using `_.isEqualWith()` instead of `Object.keys`. 
It will be way more readable and less code

src/components/plots/PlotContainer.jsx
```diff
@@ -0,0 +1,147 @@
...
+    dispatch({
+      type: LOAD_CONFIG,
+      payload: {
+        experimentId,
+        plotUuid,
+        plotType,
+        config: _.cloneDeep(initialPlotConfigStates[plotType]),
+      },
+    });
+    dispatch(savePlotConfig(experimentId, plotUuid));
```
cosa65: I think a nicer way to do this would be to call a special action creator that does the changes, something like `resetPlotConfig` that dispatches one single action `RESET_PLOT_CONFIG` and also performs the api request

src/components/plots/styling/PlotStyling.jsx
```diff
@@ -86,7 +86,7 @@ const PlotStyling = (props) => {
   });
 
   return (
-    <Collapse accordion defaultActiveKey={defaultActiveKey}>
+    <Collapse accordion defaultActiveKey={defaultActivePanel}>
```
cosa65: `defaultActiveKey` is the name given by antd, I think it will be less confusing to preserve that name instead of changing to `defaultActivePanel` and using 2 different names for the same thing.

src/components/plots/styling/violin/ViolinControls.jsx
```diff
@@ -15,13 +15,13 @@ const ViolinControls = (props) => {
 
   return (
     <Collapse>
-      <Panel header='Gene selection' key='1'>
+      <Panel header='Gene selection' key='Gene selection'>
```
cosa65: These to camelcase as well, there's also other palces with this change required as well I think

src/components/plots/styling/volcano/ThresholdsGuidesEditor.jsx
```diff
@@ -78,13 +81,10 @@ const ThresholdsGuidesEditor = (props) => {
             <Space>
               <InputNumber
                 min={0}
-                defaultValue={config.negLogpValueThreshold}
+                value={config.negLogpValueThreshold}
```
cosa65, line 84: Does this work well? It updates the render fast? Or does it lag waiting for the dispatch and redux to update?

cosa65: If it lags I think I would look into adding an internal value for this that causes the `onUpdate` to be triggered like with `SliderWithInput` for example

src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
```diff
@@ -135,49 +132,35 @@ const EmbeddingCategoricalPage = ({ experimentId }) => {
...
-            </Space>
+            </PlotContainer>
+          </Col>
+          <Col span={8}>
+            <PlotStyling
+              formConfig={plotStylingControlsConfig}
+              config={config}
+              onUpdate={updatePlotWithChanges}
+              renderExtraPanels={renderExtraPanels}
+              defaultActivePanel='Group by'
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx]: This to camelcase too

src/components/plots/PlotContainer.jsx
aerlaut: Good catch! I'll just use the `<Button>` component

src/components/plots/PlotContainer.jsx
aerlaut: I think `instanceof` can be used in this case, as all objects are derived from the `Object` object. However, semantically, I think `typeof` is better in this case, because we don't really care which object the value is derived from.

src/components/plots/PlotContainer.jsx
aerlaut: I think it's basically doing the same thing, because inside `_.isEqualWith`, we'd also have to iterate over every keys of one of the objects, and compre it to the 2nd object. In effect, we'd be wrapping this inside the callback to `_.isEqualWith`

src/components/plots/styling/PlotStyling.jsx
aerlaut: I think semantically `defaultActivePanel` is more meaningful, as it's easier to understand that we want the `defaultActivePanel` to be the one that's opened first. or `defaultActivePanelKey`?

src/components/plots/styling/violin/ViolinControls.jsx
aerlaut: changed

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I think it's better to go with the `kebab-case` to differentiate it from variable names, where we use camel case

src/components/plots/PlotContainer.jsx
aerlaut: action added

src/components/plots/PlotContainer.jsx
aerlaut: Hmm... I'm not sure how to make this more automatic. As commented, these are fields which are dynamically changed. Most of them are obvious (e.g. `shownGene`, `selectedGenes`), but one example (`title` on violin plot) isn't. Putting them inside a property like you mention is a good start,  but whoever is creating the plot would also have to be aware. Maybe We can have an array of keys in each plot's config that we want to skip when we reset the config  (e.g. `keepValuesOnReset`). That would be a start I think.

cosa65: between an array `keepValuesOnReset` with keys apart from the actual keys and an object `customFields` where the keys/values are and nowhere else I think I prefer the object, why would the array be better?

aerlaut: If we use the object, then we'd have to be aware that for some plots, keys that are usually under config (e.g. `title`) would be under the object key (e.g. if the object is `customFields` then the value would be in `customFields.title`). 

It would also make it difficult if we want to make changes to the field which makes it not dynamic anymore. For example, if in the previous case, if we don't want `title` to not be dynamic anymore, we'd have to move `title` out and make changes to the corresponding `generatePlotConfig/generatePlotData` functions.

If we use the array, we can just insert/remove keys which we want to ignore.

cosa65: Ah ok, good points, array sounds good then

aerlaut:
Moved to #679 as changes due to merging of other PRs are causing too many merged conflicts

src/components/plots/styling/volcano/ThresholdsGuidesEditor.jsx
aerlaut: Yea, it's quite good after I put some debounce on it. You can try in staging in PR #679

PR #672, opened 2022-03-30
[BIOMAGE-1567] - Move plots into a common container

aerlaut:
# Description
#### Problem
There are inconsistencies of the visual between pages. We'd like this to be uniform so our platform will look more polished.

#### Solution
Refactor page containers into separate component. Created 2 types of components: 
- MultiTileContainer which allows a page to have multi-tile. This is just a refactor of the current tiling functionality that we have so things are much cleaner.
- SingleTileContainer which use divs with the correct styling to provide a visual of having a tile but strips the required imports to keep page size small.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1567

https://ui-agi-lanky-frog.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


vickymorrison (approved):

kafkasl:
@aerlaut  what is this new page `/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer` reported by bundle size ?

I don't understand what's the difference between `MultiTileContainer` & `SingleTileContainer`.

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
```diff
@@ -117,39 +15,9 @@ const PlotsTablesHome = ({ experimentId, experimentData }) => {
...
-                  >
-                    <CardItem item={item} />
-                  </Link>
-                </Card>
-              </Col>
-            ))}
-          </Row>
-        ))}
-      </Space>
+      <SingleTileContainer title='Select Plot'>
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/index.jsx]: I think the `Select Plot` title is unnecessary. It's just obvious that you have to select a plot and it makes the screen more cluttered IMO. You already have the title `Plots & Tables` at the top

src/pages/data-management/index.jsx
```diff
@@ -112,7 +104,7 @@ const DataManagementPage = () => {
           );
         }
 
-        return (<ProjectDetails width={width} height={height} />);
```
kafkasl, line 115: why are these removed?

src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx
```diff
@@ -0,0 +1,169 @@
...
+];
+
+const PlotsTablesContainer = (props) => {
+  const { width, height, experimentId } = props;
+
+  return (
+    <div
+      style={{
+        width: width - (layout.PANEL_PADDING * 2),
+        height: height - (layout.PANEL_HEADING_HEIGHT * 2),
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx]: I think the plot container should not have a heading at all. It just feels unnecessary to have an extra line saying "Plot". We were trying to remove stuff so I think that just having the "Reset" and "Info" buttons floating on the top right without a header would look much cleaner.

src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx]: having this in pages seems to add a new route `/plots-and-tables/PlotsTablesContainer` which I assume it's not what we want. Shouldn't this be placed in `components` ?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -316,12 +311,12 @@ const DotPlotPage = (props) => {
         <center>
           <Empty description={(
             <>
-              <Paragraph>
```
Comment (kafkasl) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:319]: have you checked that `Paragraph` styling is the same to `p` in all browsers? I'm concerned you checked that it looked good in your browser but `<Paragraph>` sets the styling for older or different browsers to make it look consistent everywhere.

kafkasl:
in order to try to make PRs smaller maybe you could've split this one and first propose the new tiling components (without doing the migration) so people can check whether they make sense and review quickly, and then just proceed with migrating the other components

aerlaut:
> @aerlaut what is this new page `/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer` reported by bundle size ?
> 
> I don't understand what's the difference between `MultiTileContainer` & `SingleTileContainer`.

Oh yea, this should be under `components`. I'll move it there.

`MultipleTileContainer` is a component that shows multiple tiles in a single page. This is like what we have in Data Management and Data Processing, where we have different panels for different components. We use the `ReactMosaic` library to have this UI. This library also provides additional capabilities such as resizing and re-arranging the position of tiles.

`SingleTileContainer` provides the same "tiling" view like multiple tile, but for a single tile. Because there is only a single component/panel that has to be rendered, interactions provided by the `ReactMosaic` package that is useful when we have multiple panels (e.g. resizing, rearranging) is useless for single tile (we don't resize because the panel will always be 100% or rearrange because that's the only panel). Looking at the generated HTML, the tiling effect of `ReactMosaic` is rendered by a series of `divs` with some styles, so I wrote `SingleTileContainer` to emulate what `ReactMosaic` would render for single tile without importing the libray.

src/pages/experiments/[experimentId]/plots-and-tables/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/index.jsx]: Removed

src/pages/data-management/index.jsx
aerlaut: The `width` and `height` properties were used in this component to calculate the height of project details component. I replaced this with css changes to the `ProjectDetails.jsx` and `MultiTIleContainer.jsx` to provide the same effect.

src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx]: We're trying to make the look more consistent throughout the platform. In my opinion, these additions make for a better UI and UX.

src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/PlotsTablesContainer.jsx]: Yep, my bad. Moved under `components`.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx:319]: The antd `Paragraph` (https://github.com/ant-design/ant-design/blob/master/components/typography/Paragraph.tsx) replaces paragraphs with `div`s and adds a 1-em margin styling, which is also added for normal (non-antd) paragraph elements. It doesn't have any other internal functionality. I've checked that replacing this tag with paragraphs doesn't change anything.

aerlaut:
> in order to try to make PRs smaller maybe you could've split this one and first propose the new tiling components (without doing the migration) so people can check whether they make sense and review quickly, and then just proceed with migrating the other components

Yea I should've done that. The idea for the tiling came up when I was in the middle of doing the PR. Still, I should have separated it though.

ivababukova:
From UIX perspective, what is the point of saying "Select Plot" at the top of plots and tables?
Screenshot 2022-04-18 at 11 12 38

aerlaut:
> From UIX perspective, what is the point of saying "Select Plot" at the top of plots and tables? Screenshot 2022-04-18 at 11 12 38

That's removed with commit [f0b5434](https://github.com/hms-dbmi-cellenics/ui/pull/672/commits/f0b5434a7ad6328537222d9372aff4bf4aa05e76). The staging link is not yet updated.

kafkasl:
> From UIX perspective, what is the point of saying "Select Plot" at the top of plots and tables? Screenshot 2022-04-18 at 11 12 38

@ivababukova I think it's the same point I raised here for the "Plot" text which just feels unnecessary https://github.com/hms-dbmi-cellenics/ui/pull/672#discussion_r851333093. I would even say that both "Plot" and "Controls" are completely unnecessary. Pretty obvious what they are

image

aerlaut:
> > From UIX perspective, what is the point of saying "Select Plot" at the top of plots and tables? Screenshot 2022-04-18 at 11 12 38
> 
> @ivababukova I think it's the same point I raised here for the "Plot" text which just feels unnecessary [#672 (comment)](https://github.com/hms-dbmi-cellenics/ui/pull/672#discussion_r851333093). I would even say that both "Plot" and "Controls" are completely unnecessary. Pretty obvious what they are
> 
> image

The headers for these components are provided by the `ReactMosaicComponent`. I'll see if I can remove these.

aerlaut:
Looking at the ReactMosaic source code, there is no way to display or not display the "header" bar: https://github.com/nomcopter/react-mosaic/blob/master/src/MosaicWindow.tsx#L70-L72

Using the current implementation, the best that can be done is to leave the headers without text:
image

What do you think @kafkasl  @ivababukova?

ivababukova:
> Looking at the ReactMosaic source code, there is no way to display or not display the "header" bar: https://github.com/nomcopter/react-mosaic/blob/master/src/MosaicWindow.tsx#L70-L72
> 
> Using the current implementation, the best that can be done is to leave the headers without text: image
> 
> What do you think @kafkasl @ivababukova?

I personally don't think it is a big deal. It just looked strange to me, if it is hard to remove it, we can keep it.

kafkasl:
> 

 The controls one empty looks pretty bad. I'd say then to keep "Controls" and remove the "Plot". But I don't have any strong preference anymore if the header can't be removed which would be the best for me.

aerlaut:
> > 
> 
> The controls one empty looks pretty bad. I'd say then to keep "Controls" and remove the "Plot". But I don't have any strong preference anymore if the header can't be removed which would be the best for me.

I think it looks even weirder if one of the tile has title, and the other doesn't have one. So, although it's unnecessary, I think it's better to just keep both.

src/components/MultiTileContainer.jsx
```diff
@@ -0,0 +1,57 @@
...
+import { layout } from 'utils/constants';
+
+const renderWindow = (tile, width, height, style) => {
+  if (!tile) return <></>;
+
+  return (
+    <div style={{
+      padding: layout.PANEL_PADDING,
+      overflow: 'auto',
+      ...style,
```
StefanBabukov, line 15: why do you need to spread the style at the end?

src/components/plots/PlotContainer.jsx
```diff
@@ -36,42 +52,51 @@ const PlotContainer = (props) => {
...
+      setResetDisabled(true);
       return;
     }
 
-    setEnableReset(true);
+    setResetDisabled(false);
   }, [config]);
 
   const onClickReset = (event) => {
     event.stopPropagation();
```
StefanBabukov: can you check if we need to stop event propagation here? I think it might not be needed?

src/components/plots/PlotContainer.jsx
```diff
@@ -84,12 +109,40 @@ const PlotContainer = (props) => {
     </Space>
   );
 
+  const TILE_MAP = {
+    [PLOT]: {
+      toolbarControls: renderPlotToolbarControls(),
+      component: () => <center>{children}</center>,
```
StefanBabukov: Do we need this `center` here? Also I read that center is removed from HTML5 and instead you should use `text-align: center` css property if you actually need it - https://www.techonthenet.com/html/elements/center_tag.php#:~:text=Description,as%20the%20element.

src/components/plots/PlotResetButton.jsx
```diff
@@ -0,0 +1,105 @@
...
+  const checkConfigEquality = (currentConfig, initialConfig) => {
+    const ignoredFields = {
+      // config fields that are set dynamically on component render
+      // should not be compared to their initial values
+      frequency: ['proportionGrouping', 'xAxisGrouping'],
+      embeddingCategorical: ['axes'],
+      embeddingContinuous: ['shownGene', 'axes'],
+      violin: ['shownGene'],
+      markerHeatmap: ['selectedGenes', 'groupedTracks'],
+      DotPlot: ['selectedGenes'],
```
StefanBabukov: If we change the plotUuid of a plot this will not work, can you check if you can add the plot names dynamically? Do we have a file with the plot names?

src/components/plots/PlotContainer.jsx
```diff
@@ -36,42 +52,51 @@ const PlotContainer = (props) => {
...
+      {showReset ? (
         <Button
           key='reset'
           type='primary'
           size='small'
           onClick={onClickReset}
-          disabled={!enableReset}
+          disabled={resetDisabled}
         >
           Reset
```
StefanBabukov, line 100: Can you use the ResetButton component you created here and remove all this code including the functions and variables working with it?

src/components/plots/PlotsTablesContainer.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+import { plotNames, layout } from 'utils/constants';
+import heatmap from '../../../public/static/media/heatmap.png';
+import embeddingContinuous from '../../../public/static/media/embeddingContinuous.png';
+import embeddingCategorical from '../../../public/static/media/embeddingCategorical.png';
+import violin from '../../../public/static/media/violin.png';
+import dotPlot from '../../../public/static/media/dotplot.png';
+import volcano from '../../../public/static/media/volcano.png';
+import frequency from '../../../public/static/media/frequency.png';
+import markerHeatmap from '../../../public/static/media/marker_heatmap.png';
```
StefanBabukov, line 16: change paths to absolute?

src/components/plots/styling/violin/ViolinControls.jsx
```diff
@@ -59,9 +59,14 @@ const ViolinControls = (props) => {
...
 ViolinControls.propTypes = {
-  config: PropTypes.object.isRequired,
+  config: PropTypes.object,
   onUpdate: PropTypes.func.isRequired,
   setSearchedGene: PropTypes.func.isRequired,
   cellSets: PropTypes.object.isRequired,
 };
+
+ViolinControls.defaultProps = {
+  config: null,
```
StefanBabukov: What is the use of the controls if config can be null? Why dont we enforce it like before?

src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx
```diff
@@ -27,40 +24,38 @@ import { plotNames } from 'utils/constants';
...
-    },
-    geneNamesFilter: null,
-    sorter: { field: PROPERTIES[0], columnKey: PROPERTIES[0], order: 'descend' },
-  };
-
   const [searchedGene, setSearchedGene] = useState();
 
   useEffect(() => {
-    if (!_.isNil(config?.shownGene) && !searchedGene) {
+    dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx]: I think you should load this only if redux is empty, right? Otherwise we do redundant calls

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -214,106 +209,37 @@ const MarkerHeatmap = ({ experimentId }) => {
...
+    spec.description = 'Marker heatmap';
 
-    const newVegaSpec = {
-      ...spec,
-      axes: [...spec.axes, ...displayLabels()],
-      data: spec.data.map((datum) => ({
-        ...datum,
-        values: data[datum.name],
-      })),
+    const extraMarks = {
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:211]: I feel like this stuff should happen inside the generateSpec not here? Also in the other heatmap i saw a similar thing for the axes

src/pages/experiments/[experimentId]/plots-and-tables/violin/index.jsx
```diff
@@ -90,40 +86,39 @@ const ViolinIndex = ({ experimentId }) => {
     />
   );
 
+  const renderPlot = () => {
+    if (!config) {
+      return (
+        <center>
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/violin/index.jsx:92]: We dont need this center I think?

src/components/plots/PlotResetButton.jsx
aerlaut: This component is a carry over from a previous commit. This is not used anymore in favor of a direct button as implemented in `PlotContainer.jsx`. I'm deleting this,

src/components/plots/PlotContainer.jsx
aerlaut: Replaced this by adding styles to the containing div.

ivababukova:
> > in order to try to make PRs smaller maybe you could've split this one and first propose the new tiling components (without doing the migration) so people can check whether they make sense and review quickly, and then just proceed with migrating the other components
> 
> Yea I should've done that. The idea for the tiling came up when I was in the middle of doing the PR. Still, I should have separated it though.

Separating this PR out would have been amazing

src/components/plots/PlotsTablesContainer.jsx
aerlaut: These can't be changed to the absolute path because they lay outside the `src` folder. We'd have to add a new path for these files. I don't think it's worth it because we don't use it a lot, so I'd leave this.

src/components/plots/PlotContainer.jsx
aerlaut: The `ResetButton` was a carry over from a previous commit, I'm using this instead of that

src/components/plots/styling/violin/ViolinControls.jsx
aerlaut: changed to be required

src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx]: hmm... yea. I'll put guard clauses around these `loadPlotConfig` calls.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:211]: The generateHeatmapSpec functionality is used for both the marker heatmap and the heatmap plot. What differentiates them is just these markings. I don't think passing the plot type to `generateHeatmapSpec` and using it inside the function to decided what to add is a good way to differentiate between the two. That's why I take them out and put them here.

What I can do to make it look better is maybe to wrap it inside a wrapper function that adds these settings to the output of `generateHeatmapSpec`?

src/pages/experiments/[experimentId]/plots-and-tables/violin/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/violin/index.jsx:92]: removed

src/components/plots/PlotContainer.jsx
aerlaut: Yep, it's not needed. Removed. I think this was also a carry over from a previous PR,

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (StefanBabukov) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:211]: I think that just complicates things more. Maybe just keep it like this?

src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
```diff
@@ -31,8 +31,8 @@ const EmbeddingCategoricalPage = ({ experimentId }) => {
   const hierarchy = useSelector(getCellSetsHierarchy());
 
   useEffect(() => {
+    if (!config) dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
     dispatch(loadCellSets(experimentId));
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx:35]: Add a check for the cellsets too?

src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx
```diff
@@ -50,7 +50,7 @@ const ContinuousEmbeddingPage = ({ experimentId }) => {
   const [searchedGene, setSearchedGene] = useState();
 
   useEffect(() => {
-    dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
+    if (!config) dispatch(loadPlotConfig(experimentId, plotUuid, plotType));
     dispatch(loadCellSets(experimentId));
```
Comment (StefanBabukov) [inline src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx:54]: Add a check for the cellsets too?

src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/embedding-categorical/index.jsx:35]: `loadCellSets` has [an internal check](https://github.com/hms-dbmi-cellenics/ui/blob/master/src/redux/actions/cellSets/loadCellSets.js#L13-L15) which would not fire the request if cell set exists, so we don't have to explicitly guard this

src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/embedding-continuous/index.jsx:54]: `loadCellSets` has [an internal check](https://github.com/hms-dbmi-cellenics/ui/blob/master/src/redux/actions/cellSets/loadCellSets.js#L13-L15) which would not fire the request if cell set exists, so we don't have to explicitly guard this

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:211]: I'll create the wrapper function to make it easier to read.

StefanBabukov (approved):
lgtm

PR #676, opened 2022-04-07
Sample create delete api v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1805

https://ui-martinfosco-ui676-api330.scp-staging.biomage.net/
N/A

https://github.com/hms-dbmi-cellenics/api/pull/330

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/redux/actions/projects/loadProjects.js
```diff
@@ -63,9 +63,9 @@ const loadProjects = () => async (dispatch) => {
       // This section commented out because we going to use it
       // when samples loading is implemented in api v2
```
ivababukova, line 65: can you remove this comment?

src/redux/actions/projects/loadProjects.js
```diff
@@ -53,7 +53,7 @@ const loadProjects = () => async (dispatch) => {
 
       await Promise.all(data
         .filter((entry) => entry.samples.length)
-        .map((entry) => dispatch(loadSamples(false, entry.uuid))));
+        .map((entry) => dispatch(loadSamples(null, entry.uuid))));
```
ivababukova, line 56: why passing null here?

src/redux/actions/samples/createSample.js
```diff
@@ -35,14 +49,22 @@ const createSample = (
...
 
-  const url = `/v1/projects/${projectUuid}/${experimentId}/samples`;
+    let sampleTechnology;
+    if (type === '10X Chromium') {
+      sampleTechnology = '10x';
+    } else {
+      throw new Error(`Sample type ${type} is not implemented`);
+    }
+
+    body = { name, sampleTechnology };
```
ivababukova, line 66: Why are we not adding the rest of the items in `sampleTemplate` to the body? I see that we are adding them only for v1

src/redux/actions/samples/deleteSamples.js
```diff
@@ -33,6 +37,20 @@ const sendDeleteSamplesRequest = async (projectUuid, experimentId, sampleUuids)
   );
 };
 
+const sendDeleteSamplesRequestApiV2 = async (experimentId, sampleUuids) => {
```
ivababukova, line 40: I see that sometimes the code naming assumes that no version means v1 (for example, in `sendDeleteSamplesRequest` function name) and sometimes, no version name means v2 (for example, the `SAMPLES_DELETE` and `SAMPLES_DELETE_API_V1` action names). Can you be consistent with this? This difference creates bigger scope for errors and more complexity (every time the code reader has to check both versions of a function/variable and its implementations to see what version it is, unless is has v1 or v2 appended to its name). I suggest for both v1- and v2- specific things, append v1 and v2 to the name respectively

src/redux/actions/samples/deleteSamples.js
```diff
@@ -87,26 +105,36 @@ const deleteSamples = (
...
+          dispatch({
+            type: PROJECTS_UPDATE,
+            payload: {
+              projectUuid,
+              project: {
+                samples: newSamples,
+              },
+            },
+          });
+
```
ivababukova, line 123: why do you not need to dispatch `PROJECTS_UPDATE` in v2?

src/redux/actions/projects/loadProjects.js
cosa65: this was a minor fix, this line was using `false` in the same way as `null` relying on the fact that we check for nulls by doing `!experimentId`. `null` is way clearer about the fact that this is not being deinfed on purpose and doesn't vary the type of what we are passing between `string` and `bool`.

src/redux/actions/samples/deleteSamples.js
```diff
@@ -87,26 +105,36 @@ const deleteSamples = (
...
+          dispatch({
+            type: SAMPLES_DELETE_API_V1,
```
ivababukova, line 125: it seems that both SAMPLES_DELETE_API_V1 and SAMPLES_DELETE use the same reducer. Then why we need to separate action types?

src/redux/actions/samples/createSample.js
cosa65: Because we don't need any of them for api v2:
- `experimentId`/`projectUuid` is being passed in the url
- `sampleId` is being passed in the url
- createdDate (createdAt) and lastModified (updatedAt) are managed by the api
- metadata is managed by the api in `MetadataTrack`'s method `createNewSampleValues`

src/redux/reducers/projects/index.js
```diff
@@ -87,6 +88,10 @@ const projectsReducer = (state = initialState, action) => {
       return samplesCreate(state, action);
     }
 
+    case SAMPLES_DELETE: {
+      return samplesDelete(state, action);
```
ivababukova, line 92: There is `samplesDelete` in reducers/samples too, so I am confused. Are they doing different things? Why do we need samplesDelete in all these places?

src/redux/actions/samples/createSample.js
cosa65: The ui was sending exactly what to store in dynamo, now we send only the information that the api needs

src/redux/actions/samples/deleteSamples.js
cosa65: becuase I added instead a reducer in the `projects` slice for `SAMPLES_DELETE`, same thing for `experiments` slice and `samples` slice

src/redux/actions/projects/loadProjects.js
ivababukova: Yes, I see what you mean. Null refers to `experimentId`. It is indeed better to pass null here instead.

src/redux/actions/samples/createSample.js
ivababukova: makes sense, thanks for the clarification!

src/redux/actions/samples/deleteSamples.js
cosa65: related to [this](https://github.com/hms-dbmi-cellenics/ui/pull/676#discussion_r847313292), because the difference is that in v1 we send 3 different redux actions to update each of the slices, so `SAMPLES_DELETE_API_V1` is ignored by expeirment and projects slices, but `SAMPLES_DELETE` is not ignored.

I'll update `SAMPLES_DELETE` to `SAMPLES_DELETE_API_V2` due to your other comment about consistent naming separation.

src/redux/reducers/projects/index.js
cosa65: related to [this](https://github.com/hms-dbmi-cellenics/ui/pull/676#discussion_r847323018)

src/redux/actions/samples/deleteSamples.js
cosa65: Updated

ivababukova (approved):

src/redux/actions/projects/loadProjects.js
cosa65: I think it should be left because we are going to use it one we implement samples loading

PR #686, opened 2022-04-14
[BIOMAGE-1835] Implement samples PATCH for api v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1835

https://ui-martinfosco-ui686-api334.scp-staging.biomage.net/
N/A

https://github.com/hms-dbmi-cellenics/api/pull/334

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/__test__/redux/actions/samples/updateSample.test.js
+++ b/src/__test__/redux/actions/samples/updateSample.test.js
@@ -1,14 +1,20 @@
 import _ from 'lodash';
 import configureStore from 'redux-mock-store';
+import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
 import thunk from 'redux-thunk';
+
 import updateSample from 'redux/actions/samples/updateSample';
 import initialState, { sampleTemplate } from 'redux/reducers/samples/initialState';
 import { saveSamples } from 'redux/actions/samples';
 
-import { SAMPLES_UPDATE } from 'redux/actionTypes/samples';
+import {
+  SAMPLES_ERROR, SAMPLES_SAVED, SAMPLES_SAVING, SAMPLES_UPDATE,
+} from 'redux/actionTypes/samples';
+
+import config from 'config';
+import { api } from 'utils/constants';
 
 jest.mock('redux/actions/samples/saveSamples');
-saveSamples.mockImplementation(() => async () => { });
 
 const mockStore = configureStore([thunk]);
 
@@ -33,6 +39,16 @@ describe('updateSample action', () => {
     },
   };
 
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    saveSamples.mockImplementation(() => async () => { });
+
+    enableFetchMocks();
+    fetchMock.resetMocks();
+    fetchMock.doMock();
+  });
+
   it('Dispatches event correctly', async () => {
     const store = mockStore(mockState);
     await store.dispatch(updateSample(mockUuid, updatedSample));
@@ -56,5 +72,49 @@ describe('updateSample action', () => {
     await store.dispatch(updateSample(mockUuid, updatedSample));
 
     expect(saveSamples).toHaveBeenCalled();
+  });
+
+  it('Dispatches events correctly for api v2', async () => {
+    config.currentApiVersion = api.V2;
+    fetchMock.mockResponseOnce(() => Promise.resolve(JSON.stringify({})));
+
+    const sampleDiff = {
+      name: 'updated name',
+    };
+
+    const store = mockStore(mockState);
+    await store.dispatch(updateSample(mockUuid, sampleDiff));
+
+    const actions = store.getActions();
+    expect(_.map(actions, 'type')).toEqual([SAMPLES_SAVING, SAMPLES_SAVED, SAMPLES_UPDATE]);
+    expect(_.map(actions, 'payload')).toMatchSnapshot();
+
+    expect(fetchMock).toHaveBeenCalledWith(
+      'http://localhost:3000/v2/experiments/null/samples/asd123',
+      {
+        body: JSON.stringify(sampleDiff),
+        headers: { 'Content-Type': 'application/json' },
+        method: 'PATCH',
+      },
+    );
+
+    expect(saveSamples).not.toHaveBeenCalled();
+  });
+
+  it('Error handling for api v2 works', async () => {
+    config.currentApiVersion = api.V2;
+
+    fetchMock.mockRejectOnce(() => Promise.reject(new Error('Api error')));
+
+    const sampleDiff = {
+      name: 'updated name',
+    };
+
+    const store = mockStore(mockState);
+    await store.dispatch(updateSample(mockUuid, sampleDiff));
+
+    const actions = store.getActions();
+    expect(_.map(actions, 'type')).toEqual([SAMPLES_SAVING, SAMPLES_ERROR]);
+    expect(_.map(actions, 'payload')).toMatchSnapshot();
   });
 });
--- a/src/redux/actions/samples/updateSample.js
+++ b/src/redux/actions/samples/updateSample.js
@@ -3,12 +3,52 @@ import _ from 'lodash';
 
 import endUserMessages from 'utils/endUserMessages';
 
+import {
+  SAMPLES_UPDATE, SAMPLES_SAVING, SAMPLES_SAVED,
+} from 'redux/actionTypes/samples';
+import saveSamples from 'redux/actions/samples/saveSamples';
+
 import mergeObjectWithArrays from 'utils/mergeObjectWithArrays';
 import handleError from 'utils/http/handleError';
-import {
-  SAMPLES_UPDATE,
-} from '../../actionTypes/samples';
-import saveSamples from './saveSamples';
+import fetchAPI from 'utils/http/fetchAPI';
+
+import config from 'config';
+import { api } from 'utils/constants';
+
+const sendApiV2Request = async (experimentId, sampleId, diff, dispatch) => {
+  const url = `/v2/experiments/${experimentId}/samples/${sampleId}`;
+  const body = diff;
+
+  dispatch({
+    type: SAMPLES_SAVING,
+    payload: {
+      message: endUserMessages.SAVING_SAMPLE,
+    },
+  });
+
+  await fetchAPI(
+    url,
+    {
+      method: 'PATCH',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(body),
+    },
+  );
+
+  dispatch({
+    type: SAMPLES_SAVED,
+  });
+
+  dispatch({
+    type: SAMPLES_UPDATE,
+    payload: {
+      sampleId,
+      sample: diff,
+    },
+  });
+};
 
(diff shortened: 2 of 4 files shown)


src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,84 @@ import _ from 'lodash';
...
+  });
+
+  dispatch({
+    type: SAMPLES_UPDATE,
+    payload: {
+      sampleId,
+      sample: diff,
+    },
+  });
+};
```
ivababukova: if the version of the api is v2, then the `SAMPLES_UPDATE` action will be dispatched twice: once here and once in lines 77-83

src/__test__/redux/actions/samples/updateSample.test.js
```diff
@@ -33,6 +39,16 @@ describe('updateSample action', () => {
...
   };
 
+  beforeEach(() => {
+    jest.clearAllMocks();
+
+    saveSamples.mockImplementation(() => async () => { });
+
+    enableFetchMocks();
+    fetchMock.resetMocks();
+    fetchMock.doMock();
```
kafkasl, line 49: are these all required? just mentioning because I've seen some tests where many of these are just copy pasted but aren't needed. What does `doMock` do by the way?

src/__test__/redux/actions/samples/updateSample.test.js
```diff
@@ -57,4 +73,48 @@ describe('updateSample action', () => {
...
+    };
+
+    const store = mockStore(mockState);
+    await store.dispatch(updateSample(mockUuid, sampleDiff));
+
+    const actions = store.getActions();
+    expect(_.map(actions, 'type')).toEqual([SAMPLES_SAVING, SAMPLES_SAVED, SAMPLES_UPDATE]);
+    expect(_.map(actions, 'payload')).toMatchSnapshot();
+
+    expect(fetchMock).toHaveBeenCalledWith(
```
kafkasl, line 92: why use `toMatchSnapshot` in the previous line and here build this complex object instead?

src/__test__/redux/actions/samples/updateSample.test.js
```diff
@@ -57,4 +73,48 @@ describe('updateSample action', () => {
...
+    const sampleDiff = {
+      name: 'updated name',
+    };
+
+    const store = mockStore(mockState);
+    await store.dispatch(updateSample(mockUuid, sampleDiff));
+
+    const actions = store.getActions();
+    expect(_.map(actions, 'type')).toEqual([SAMPLES_SAVING, SAMPLES_ERROR]);
+    expect(_.map(actions, 'payload')).toMatchSnapshot();
```
kafkasl, line 118: I like using these maps together with snapshot. I think it will make tests much easier to maintain and update if we do that everywhere.

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
...
+import fetchAPI from 'utils/http/fetchAPI';
+
+import config from 'config';
+import { api } from 'utils/constants';
+
+const sendApiV2Request = async (experimentId, sampleId, diff, dispatch) => {
+  const url = `/v2/experiments/${experimentId}/samples/${sampleId}`;
+  const body = diff;
+
+  dispatch({
```
kafkasl, line 53: why is this dispatch not needed when using API v1? if it's needed move it to the top of `updateSamples`

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
...
         },
-      }));
+      });
+    } else if (config.currentApiVersion === api.V2) {
+      if (_.isNil(diff.name) || diff.metadata) {
+        throw new Error('This action shouldn\'t be used to update metadata with api v2');
+      }
+
+      // in api v2 projectUuid is actually experimentId
+      await sendApiV2Request(sample.projectUuid, sampleUuid, diff, dispatch);
```
kafkasl: `sendApiV2Request` name is too generic imo. if you intend to leave it here I'd call it something more specific

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
...
 
 const updateSample = (
   sampleUuid,
   diff,
 ) => async (dispatch, getState) => {
   const sample = _.cloneDeep(getState().samples[sampleUuid]);
 
-  // eslint-disable-next-line no-param-reassign
-  diff.lastModified = moment().toISOString();
+  try {
```
kafkasl: try/catch best practices is to make them as short as possible. I would wrap only the actions dispatch or something like that.

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
...
+      dispatch({
         type: SAMPLES_UPDATE,
         payload: {
           sampleUuid,
           sample: diff,
         },
-      }));
+      });
+    } else if (config.currentApiVersion === api.V2) {
+      if (_.isNil(diff.name) || diff.metadata) {
```
kafkasl: why not add this inside your `sendApiV2Request`

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
...
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify(body),
+      },
+    );
+  } catch (e) {
+    const errorMessage = handleError(e, endUserMessages.ERROR_SAVING, false);
+
+    dispatch({
```
kafkasl, line 86: why is this done only for v2?

src/redux/actions/samples/updateSample.js
```diff
@@ -3,34 +3,97 @@ import _ from 'lodash';
 
 import endUserMessages from 'utils/endUserMessages';
```
kafkasl, line 5: I feel that this file it's too messy due to having to support both v1 & v2. It's not clear which actions are dispatched and when for each version (as evidenced by Iva's comment) and I can totally see people inadvertently adding duplicated notifications. I have a hard time telling whether the notifications are correct or not. I am not sure what's the best course of action here. The idea is to make it as easy as possible to transition from v1 to v2. However, in this PR it will take a bit more than just removing the extra v1 code. For example, you added an extra function `sendApiV2Request`. But when we remove v1, will be a nuisance because the handleError is duplicated unnecessarily because the try/catch is too large and includes both versions. I would probably do either:
1. add a different function for v1 and v2 and add EVERYTHING inside each (that is, all the try/catch, dispatches, etc... so that when we migrate we can just complete remove v1 without any more changes. Somethine like:
```
if (config.currentApiVersion === api.V1) {
     updateSamplesV1(...)
else {
    updateSamplesV2(...)
}
```

The the other option would be to share all the flow (if that's the desired behavior which I'm not sure anymore) and just change the actual dispatch
```
const updateSample = (
  sampleUuid,
  diff,
) => async (dispatch, getState) => {
  const sample = _.cloneDeep(getState().samples[sampleUuid]);
    if (config.currentApiVersion === api.V1) {

      await dispatch(saveSamples(sample.projectUuid, newSample, true, true, false));
   } else {
      await whatever for v2
   }
      dispatch({
        type: SAMPLES_UPDATE,
        payload: {
          sampleUuid,
          sample: diff,
        },
 try {

  } catch (e) {
    const errorMessage =  handleError(e, endUserMessages.ERROR_SAVING);
    dispatch({
      type: SAMPLES_ERROR,
      payload: {
        error: errorMessage,
      },
    });
  }
```

imo something like first approach is better for the reasons listed above. I think you are trying to mix both approaches (making a function for v2 but no the other, sharing some error handling but not all of it, etc...) which. just makes everything more confusing than needed. I'd prefer code duplication if you don't like any of these options

kafkasl (changes requested):
I think the PR structure needs more work. Having to support v1 and v2 will make the code messier so I think we should take extra care to make it as separate and clear as possible. If you think that's not possible or requires too much effort I think we should seriously consider other migration options because this is very error prone.

src/redux/actions/samples/updateSample.js
cosa65: This action is particularly complex because I needed to add the actions from `saveSamples` in here, I duplicated `handleError` because for api v1 it is duplicated as well, since `saveSample` does it once and then `updateSample` does it too. 

My objective was to make redux work in the same way as it works with v1, in case not doing that introduces any problems or unpredictable behavior.

cosa65: > 1. add a different function for v1 and v2 and add EVERYTHING inside each (that is, all the try/catch, dispatches, etc... > so that when we migrate we can just complete remove v1 without any more changes. Somethine like:

This was my initial implementation when I started working on adding api v2 endpoints but it was suggested in a code review that instead I try to keep the diverging code down.

cosa65: As I said before, in this case it is more messy than normal because I needed to add everything redux-wise that `saveSample` does, the duplicated error handling that I added was not by mistake but because of `saveSamples`, maybe a call to explain this in more detail might help so we can decide how to make this easier to follow.

src/redux/actions/samples/updateSample.js
cosa65: It isn't the thing is that that is done in `saveSample`

src/__test__/redux/actions/samples/updateSample.test.js
cosa65: in this case yes, we are mocking the fetch for api v2 so we can say that the request succeeeded

src/__test__/redux/actions/samples/updateSample.test.js
cosa65: from what I understnad `toMatchSnapshot` is good but I try to avoid abusing it so that I can keep tests readable without having to go to the .snap files all the time

src/redux/actions/samples/updateSample.js
cosa65: Resolved in a call, we decided to keep code for both actions (v1 and v2) in separate functions in this case becuase it makes things too messy otherwise.

src/redux/actions/samples/updateSample.js
cosa65: I wrapped all the things that need to not happen if an error takes place because the catch doesn't rethrow so it continues executing after getting out of it.

src/redux/actions/samples/updateSample.js
cosa65: Explained in the call

src/redux/actions/samples/updateSample.js
cosa65: explained in the call

src/__test__/redux/actions/samples/updateSample.test.js
kafkasl: ok

kafkasl (approved):

PR #688, opened 2022-04-18
[BIOMAGE-1785] - Fix overflowing embedding and heatmap popup boxes

aerlaut:
# Description
#### Background
The popup boxes in the Embedding and Heatmap plots in Data Exploration can not be seen if the cell that is chosen is too close to the top-right or bottom-right part of the panel. This is because the panel is desplayed to the right and bottom of the pointer.

#### Solution
Invert the popup to display to the top-left or bottom-left of the pointer if the location of the cell is close ( < width of the pop box) to the edge.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1785

Staging link: https://ui-agi-pokey-dane.scp-staging.biomage.net/
N/A

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -307,21 +315,25 @@ const Embedding = (props) => {
...
               <div>
                 <CellInfo
+                  width={CELLINFO_WIDTH}
                   componentType={embeddingType}
-                  coordinates={cellCoordintes}
+                  coordinates={cellCoordinates.current}
                   cellInfo={cellInfoTooltip.current}
+                  numTextRows={numCellInfoTextRows}
+                  invertX={cellCoordinates.current.x + CELLINFO_WIDTH > width}
+                  invertY={cellCoordinates.current.y + (numCellInfoTextRows * EM) + CELLINFO_Y_PADDING > height}
```
ivababukova: I wouldn't calculate whether it needs to be inverted or not from the Embedding or from the Heatmap components -- this is something internal to the CellInfo component and if we do it outside of CellInfo, it only creates unnecessary coupling.

ivababukova: you can pass width of the Embedding or the Heatmap as props to CellInfo and do those calculations inside CellInfo

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -35,18 +34,22 @@ const Scatterplot = dynamic(
   { ssr: false },
 );
 
-const initialZoom = 4.00;
+const INITIAL_ZOOM = 4.00;
 const cellRadiusFromZoom = (zoom) => zoom ** 3 / 50;
 
+const EM = 16; // px
+const CELLINFO_WIDTH = 300; // px
+const CELLINFO_Y_PADDING = 6 * EM;
```
ivababukova: some of these, like `EM` and `CELLINFO_Y_PADDING` are duplicates to the constants we have in CellInfo. Look at [this comment](https://github.com/hms-dbmi-cellenics/ui/pull/688/files#r852764097) in this `Embedding.jsx` file, but these constants shouldn't be necessary in here.

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -67,11 +70,16 @@ const Embedding = (props) => {
     hidden: cellSetHidden,
   } = cellSets;
 
+  // These are calculated to correctly position the popup according to the number of text rows in the popup
+  // There are 2 default rows: cellId and louvain clusters
+  const numCustomCellSets = cellSetHierarchy.children?.length || 0;
+  const numCellInfoTextRows = numCustomCellSets + 2;
```
ivababukova: instead of doing this, can't you just get the height of CellInfo from within the CellInfo component? Similarly to invertX and invertY, this `numCellInfoTextRows` looks like something internal to CellInfo that CellInfo should worry about.

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -307,21 +315,25 @@ const Embedding = (props) => {
...
           ) : (
             (cellInfoVisible && cellInfoTooltip.current) ? (
               <div>
                 <CellInfo
+                  width={CELLINFO_WIDTH}
                   componentType={embeddingType}
-                  coordinates={cellCoordintes}
+                  coordinates={cellCoordinates.current}
                   cellInfo={cellInfoTooltip.current}
+                  numTextRows={numCellInfoTextRows}
```
ivababukova: Essentially the same as [this comment](https://github.com/hms-dbmi-cellenics/ui/pull/688/files#r852773243), but try to get rid of this prop and get the height of CellInfo from within CellInfo

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -35,18 +34,22 @@ const Scatterplot = dynamic(
   { ssr: false },
 );
 
-const initialZoom = 4.00;
+const INITIAL_ZOOM = 4.00;
 const cellRadiusFromZoom = (zoom) => zoom ** 3 / 50;
 
+const EM = 16; // px
+const CELLINFO_WIDTH = 300; // px
```
ivababukova: Ideally, we don't hardcode the cell info width -- it will vary, based on how long cell set names or gene names we have.

src/components/data-exploration/heatmap/HeatmapCellInfo.jsx
```diff
@@ -2,22 +2,29 @@ import React from 'react';
 import { Card } from 'antd';
 import PropTypes from 'prop-types';
 
+const EM = 16; // px
+const Y_PADDING = 4 * EM;
+
```
ivababukova: same comments for the Heatmap cell info and the heatmap components, as the ones I wrote for the Embedding

src/components/data-exploration/embedding/Embedding.jsx
ivababukova: Also, 300 is way too much: Screenshot 2022-04-19 at 10 57 46

ivababukova: it hides a large proportion of the plots, I imagine that users won't like it

aerlaut: Hmm... I can try to calculate the width of the text and use that. There's this guideline from Microsoft which I think I can use: https://docs.microsoft.com/en-us/typography/develop/character-design-standards/whitespace#:~:text=For%20an%20average%20width%20font,~1%2F3%20the%20em.

vickymorrison:
There's a lot of unnecessary white space at the right side of the text box. Can it be removed? See example screenshots.
Screenshot 2022-04-19 at 10 50 35
Screenshot 2022-04-19 at 10 51 47

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: Moved into `CellInfo.jsx` as recommended

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: Retrieved from `cellInfo`.

aerlaut:
Right now we have 3 components to show the same UI popup: `CellInfo.jsx`, `HeatmapCellinfo.jsx` and `HeatmapTracksCellinfo.jsx`. The only difference between them is the content that they show. I think these components can be combined and refactored into one component.

I'm also thinking of moving the generation of the popup text into the parent component. This way, the concern of the component would be just to render the content. This would also make it easier to address [the width issue](https://github.com/hms-dbmi-cellenics/ui/pull/688#discussion_r852779895) because the component will be passed the final form of the content, so we can just check which of the text has the longest length.

Do you think this refactor would be a good idea? @ivababukova

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
```diff
@@ -230,7 +230,7 @@ const CellSetsTool = (props) => {
         position: 'relative',
         height: `${height - 40}px`,
         width: `${width - 8}px`,
-        overflow: 'scroll',
```
ivababukova, line 233: what is the difference when this is in auto?

src/components/data-exploration/embedding/Embedding.jsx
ivababukova: yes, that's better

ivababukova:
> Right now we have 3 components to show the same UI popup: `CellInfo.jsx`, `HeatmapCellinfo.jsx` and `HeatmapTracksCellinfo.jsx`. The only difference between them is the content that they show. I think these components can be combined and refactored into one component.
> 
> I'm also thinking of moving the generation of the popup text into the parent component. This way, the concern of the component would be just to render the content. This would also make it easier to address [the width issue](https://github.com/hms-dbmi-cellenics/ui/pull/688#discussion_r852779895) because the component will be passed the final form of the content, so we can just check which of the text has the longest length.
> 
> Do you think this refactor would be a good idea? @ivababukova

I remember that we evaluated this when we were creating those popups. We decided to have them as separate components, because although they are mostly similar, they have some differences which when generalised, make the component much harder to understand and use. The generalisations that we attempted were leading to leaky abstractions. This, combined with the fact that we rarely use and modify these components makes me on the opinion that there is little/no point in refactoring the cellinfo components into one for now. Our time is better spent elsewhere in my opinion

src/components/data-exploration/CellInfo.jsx
```diff
@@ -1,22 +1,37 @@
 import React from 'react';
-import _ from 'lodash';
 import { Card } from 'antd';
 import PropTypes from 'prop-types';
 
+const EM = 16; // px
+const CELLINFO_WIDTH = 200; // px
```
ivababukova: Moving `CELLINFO_WIDTH`to `CellInfo` is definitely better. Can you get rid of this hardcoded width variable entirely and have another way to know when to invert or not? Or another way to detect the width? The width of the cellinfo depends on the size of the text in its content and hardcoding it is not ideal.

src/components/data-exploration/CellInfo.jsx
```diff
@@ -1,22 +1,37 @@
...
+
+const cellInfoStyle = { fontSize: '0.75rem' };
+
 const CellInfo = (props) => {
-  const { coordinates, cellInfo } = props;
+  const {
+    containerWidth, containerHeight, coordinates, cellInfo,
+  } = props;
+
+  const numTextRows = (cellInfo.cellSets?.length || 0) + 2;
```
ivababukova: similarly to the width, is there another way to know the height, or know when to invert without knowing the height?

src/components/data-exploration/cell-sets-tool/CellSetsTool.jsx
aerlaut: `overflow: 'scroll'` will show the scroll bars even though it is not required. '`overflow: 'auto'` only show the scroll bars when it is required.

image

src/components/data-exploration/CellInfo.jsx
aerlaut: The way I thought of is by counting the length of each of the lines that is going to be displayed in the box, and then multiply it by the average width of a character. Add left and right padding to that, and we should get the width of the box.

src/components/data-exploration/CellInfo.jsx
aerlaut: We can't know when to invert if we don't know whether the box will overflow outside the panel or not, and to know that, we need to know the position of the cursor and the height of the box. So there's no way of knowing when to invert if we don't know the height that the box is going to be.

Just like the width, I used an approximation for the height. The height of a line of text is usually 1em. So if we know the number of lines, then we'll know the height, assuming there is no text-overflow to the next line. This is what I used in these popups.

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: moved

src/components/data-exploration/heatmap/HeatmapCellInfo.jsx
aerlaut: moved

src/components/data-exploration/CellInfo.jsx
aerlaut: I tried using `ref` but it turns out there are issues using refs in Antd components: https://github.com/ant-design/ant-design/issues/28582. Hence, we can't get a hold of the rendered component to get it's width and height.

aerlaut: I tried following a modified approach from this [SO answer](https://stackoverflow.com/a/66663992/1940886) using a ref to get the element and then accessing its width and height using `.offsetWidth` and `.offsetHeight`. These return 0 in Chrome because of some optimizations that [Chrome does](https://stackoverflow.com/a/70043254/1940886).

aerlaut: This is solved by using https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node to get the popup element and then getting the width and height for boundary calculation.

ivababukova (approved):
that looks great now, well done!

PR #693, opened 2022-04-25
1859 sample files upload for api v2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1859

https://ui-martinfosco-ui693-api338.scp-staging.biomage.net/
N/A

https://github.com/hms-dbmi-cellenics/api/pull/338
https://github.com/hms-dbmi-cellenics/iac/pull/364

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/redux/actions/samples/createSample.js
```diff
@@ -60,9 +62,13 @@ const createSample = (
...
       sampleTechnology = '10x';
     } else {
-      throw new Error(`Sample type ${type} is not implemented`);
+      throw new Error(`Sample technology ${type} is not recognized`);
     }
 
+    filesToUploadV2Only.forEach((fileName) => {
+      newSample.files[fileName] = { upload: { status: UploadStatus.UPLOADING } };
+    });
+
```
ivababukova, line 71: so this piece of code gets executed for both v1 and v2, even though it is needed only for v2?

src/utils/upload/processUploadV2.js
```diff
@@ -0,0 +1,221 @@
...
+    });
+  } catch (e) {
+    console.log('uploadError');
+    console.log(e);
+
+    console.log('uploadErrorResponse');
+    console.log(e.response);
+
+    console.log('uploadErrorResponseData');
+    console.log(e.response?.data);
```
ivababukova: remove these

src/redux/actions/samples/createSample.js
cosa65: No, it is inside the `if(config.currentApiVersion === api.V2)` block of code

src/redux/actions/samples/createSample.js
```diff
@@ -60,9 +62,13 @@ const createSample = (
     if (type === '10X Chromium') {
       sampleTechnology = '10x';
     } else {
-      throw new Error(`Sample type ${type} is not implemented`);
+      throw new Error(`Sample technology ${type} is not recognized`);
     }
```
aerlaut, line 66: There's currently no check for this for V1?

src/redux/actions/samples/createSample.js
```diff
@@ -38,7 +40,7 @@ const createSample = (
   let body;
 
   const newSample = {
-    ...sampleTemplate,
+    ..._.cloneDeep(sampleTemplate),
```
aerlaut, line 43: Why do you need to `_.cloneDeep` this? Isn't just spreading the object enough?

src/redux/actions/samples/createSample.js
```diff
@@ -83,6 +89,11 @@ const createSample = (
       payload: { sample: newSample, experimentId },
     });
 
+    await dispatch({
+      type: SAMPLES_SAVED,
+      payload: { sample: newSample, experimentId },
+    });
```
aerlaut, line 95: Why not add the effect of `SAMPLES_SAVED` to `SAMPLES_CREATE`?

src/redux/actions/samples/createSample.js
```diff
@@ -60,9 +62,13 @@ const createSample = (
     if (type === '10X Chromium') {
       sampleTechnology = '10x';
     } else {
-      throw new Error(`Sample type ${type} is not implemented`);
+      throw new Error(`Sample technology ${type} is not recognized`);
     }
 
+    filesToUploadV2Only.forEach((fileName) => {
+      newSample.files[fileName] = { upload: { status: UploadStatus.UPLOADING } };
+    });
```
aerlaut, line 70: Why do we need this? I see that it is not passed into the request `body`

src/utils/upload/processUploadV1.js
```diff
@@ -0,0 +1,272 @@
...
+      dispatch(
+        updateSampleFile(
+          sampleUuid,
+          fileName,
+          { upload: { status: UploadStatus.COMPRESSING } },
+        ),
+      )
+    ));
+  } catch (e) {
+    const fileErrorStatus = e === 'aborted' ? UploadStatus.FILE_READ_ABORTED : UploadStatus.FILE_READ_ERROR;
```
aerlaut: Shouldn't this be `e.message === 'aborted'`? `e` is an instance of `Error` right?

src/redux/actions/samples/updateSampleFileUploadV2.js
```diff
@@ -0,0 +1,56 @@
...
+      // This error means that we couldn't update the api on the state of the upload
+      // Should we block the upload from continuing? The upload itself isn't necessarily compromised
+      handleError(e, endUserMessages.ERROR_UPDATE_SERVER_ON_UPLOAD_STATE);
+    }
+  }
+
+  const updatedAt = moment().toISOString();
+
+  // Even if we failed to persist the update in the api it is still
+  // valid information so we should dispatch the action anyways
```
aerlaut: Wouldn't this cause inconsistent state if the user reloads the page?

src/redux/actions/samples/updateSampleFileUploadV2.js
```diff
@@ -0,0 +1,56 @@
...
+  barcodes10x: 'barcodes.tsv.gz',
+  features10x: 'features.tsv.gz',
+};
+
+const updateSampleFileUploadV2 = (
+  experimentId, sampleId, type, uploadStatus, uploadProgress,
+) => async (dispatch) => {
+  // Don't send an api update whenever the progress bar is updated, only for uploadStatus changes
+  // TODO: move progress to not even be a part of redux, manage it in a different way
+  if (!uploadProgress) {
```
aerlaut: Instead of checking for `!uploadProgress`, is it better to check for `uploadStatus !== UploadStatus.UPLOADING` instead?

src/utils/upload/processUploadV2.js
```diff
@@ -0,0 +1,211 @@
...
+  projectId, sampleId, fileType, file, signedUrl, dispatch,
+) => {
+  let loadedFile = null;
+
+  try {
+    loadedFile = await loadAndCompressIfNecessary(file, () => {
+      dispatch(updateSampleFileUploadV2(projectId, sampleId, fileType, UploadStatus.COMPRESSING));
+    });
+  } catch (e) {
+    const fileErrorStatus = e === 'aborted' ? UploadStatus.FILE_READ_ABORTED : UploadStatus.FILE_READ_ERROR;
```
aerlaut: Shouldn't this be `e.message === 'aborted'`? e is an instance of Error right?

src/redux/actions/samples/createSample.js
cosa65: no, because later on we are doing this:

```
filesToUploadV2Only.forEach((fileName) => {
      newSample.files[fileName] = { upload: { status: UploadStatus.UPLOADING } };
    });
```

and it actually writes those files into the `sampleTemplate` until we refresh

src/redux/actions/samples/createSample.js
cosa65: it's so that we know which files to show as `FILE_NOT_FOUND`, without this we would end up showing all files as `FILE_NOT_FOUND` until they are loaded, in this way we are already showing "Uploading" for any file that is already being loaded and `FILE_NOT_FOUND` for any files that are really not there.

There was a different workaround for this in with `processUploadV1` which was way less clear and forced us to create the samples in a weird timing, with this other implementation we are explicitly saying which files exist when we create the sample.

src/redux/actions/samples/updateSampleFileUploadV2.js
cosa65: I didn't like  the alernatives I could come up with. It would also be inconsistent if we don't say anything about the upload continuing even thought it did, right?

Do you have any ideas?

This is related to this other comment I wrote in that same action, what would you suggest?
> // This error means that we couldn't update the api on the state of the upload
> // Should we block the upload from continuing? The upload itself isn't necessarily compromised

src/redux/actions/samples/createSample.js
cosa65: because I didn't want to refactor redux unless it is the most simple way of implementing the v2 process upload (an examples of a case where it was is `SAMPLES_DELETE_API_V2`).

right now changing `SAMPLES_SAVED` means also affecting the api v1 action, and I didn't want to get into that.

It's easier to refactor redux together after we stop using and remove the api v1 code

src/redux/actions/samples/createSample.js
cosa65: I added it for v2 because are doing this if that sets the string to `10x` so it is more likely to impact api v2, it isn't really a problem in api v1 becuase we don't really ever check the type after the initial load

src/redux/actions/samples/createSample.js
aerlaut: Ok. I thought spreading copies the object deeply. It seems that nested objects are passed by reference and not copied.

src/redux/actions/samples/updateSampleFileUploadV2.js
aerlaut:
```
// This error means that we couldn't update the api on the state of the upload
// Should we block the upload from continuing? The upload itself isn't necessarily compromised

```
For this case, we depend on the files being uploaded and the upload status to be set to UPLOADED for users to be able to continue to process the data right? So if there are any erros updating the file, it might be just better to set the upload status to failed and ask the user to reupload, as there are no other checks to check whether the file has been successfully uploaded or not.

aerlaut: I guess that follows for this comment as well

```
  // Even if we failed to persist the update in the api it is still
  // valid information so we should dispatch the action anyways
```

If we let the update continue, the user will be able to process the data on the first run because the all the sample files are uploaded. However, when they login again or come back to `data-managment`, the file upload status will not be UPLOADED, and they can't go to data processing. They'd have to reupload and process data again, right? So I think it's better to mark the upload as failed and let the user reupload rather than having them come back and see to see that they have to reprocess the data. Or, upon failure of saving, is it possible to retry several times?

cosa65: >  it might be just better to set the upload status to failed and ask the user to reupload, as there are no other checks to check whether the file has been successfully uploaded or not.

True

cosa65: good point, I'll just set to upload failed

.github/workflows/ci.yaml
```diff
@@ -77,19 +77,19 @@ jobs:
...
+      #   env:
+      #     AWS_DEFAULT_REGION: eu-west-1
+      #     GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      #   with:
+      #     test-command: 'npm run coverage'
+      #     coverage-comment: false
+
+      # - id: upload-coverage
+      #   name: Upload coverage to Codecov
+      #   uses: codecov/codecov-action@v1
```
ivababukova: don't forget to uncomment this

src/config/index.js
```diff
@@ -1,7 +1,7 @@
 import { api } from 'utils/constants';
 
 const config = {
-  currentApiVersion: api.V1,
+  currentApiVersion: api.V2,
 };
```
ivababukova: don't forget to reverse this after you're finished with testing version v2

src/redux/actions/samples/createSample.js
ivababukova: yup, just saw it, this if statement was part of the unchanged lines

src/utils/upload/processUploadV2.js
```diff
@@ -0,0 +1,211 @@
...
+
+import axios from 'axios';
+
+import { createSample, createSampleFileV2, updateSampleFileUploadV2 } from 'redux/actions/samples';
+
+import UploadStatus from 'utils/upload/UploadStatus';
+import loadAndCompressIfNecessary from 'utils/upload/loadAndCompressIfNecessary';
+import { inspectFile, Verdict } from 'utils/upload/fileInspector';
+
+// const putInS3 = async (loadedFileData, signedUrl, sampleUuid, fileType, onUploadProgress) => (
```
ivababukova: remove this commented out code

src/utils/upload/processUploadV1.js
```diff
@@ -0,0 +1,272 @@
...
+          [fileName]: file,
+        },
+      },
+    };
+  }, {});
+
+  Object.entries(samplesMap).forEach(async ([name, sample]) => {
+    // Create sample if not exists.
+    try {
+      sample.uuid ??= await dispatch(createSample(activeProjectUuid, name, sampleType));
```
ivababukova, line 203: what does `??=` mean?

src/utils/upload/processUploadV1.js
```diff
@@ -0,0 +1,272 @@
...
+          file.name,
+          {
+            ...file,
+            path: `${activeProjectUuid}/${file.name.replace(name, sample.uuid)}`,
+          },
+        ),
+      );
+    });
+  });
+};
```
ivababukova, line 225: why is this piece of code no longer needed for v2?

src/utils/upload/processUploadV2.js
```diff
@@ -0,0 +1,211 @@
...
+
+const getFileType = (fileName) => {
+  let fileType;
+
+  _.forEach(Object.entries(fileTypes), ([name, type]) => {
+    if (fileName.includes(name)) {
+      fileType = type;
+      return false;
+    }
+  });
```
ivababukova, line 73: Not required to fix for this PR: is there a way to write this to have the 10x files format less hardcoded into the data upload function? It is something we will have to do in the near future

cosa65: As part of the UI refactoring to remove api v1 code I would like to make the fileTypes be identified and written correctly as soon as the files are dropped, so this piece of code won't be necessary anymore at that point (we'll store only fileType instead of fileName).

src/utils/upload/processUploadV1.js
cosa65: It's to create the files. It is needed, but now we call one action in particualr for creating the sample files isntead of reusing `updateSampleFile` (we call `createSampleFileV2`)

src/utils/upload/processUploadV1.js
cosa65: it is to assign to uuid if it is not defined already, it was taken over form api v1

aerlaut (approved):

src/utils/upload/processUploadV2.js
ivababukova: ok, good!

src/utils/upload/processUploadV1.js
ivababukova: yes, I noticed that we had it in v1 too, I wasn't sure about the syntax

PR #690, opened 2022-04-22
[BIOMAGE-1849] - Return rendering error

aerlaut:
# Description
#### Background
Errors are repeatedly reported because the current implementation of ErrorBoundary renders the bugged page. We want the error to be rendered and reported only when the UI can not render anymore (showing the "it's not you it's us") page.

#### Solution
Remove the `ErrorBoundary` component and use Next's built-in error handling mechanism. Move the Slack reporting functionality to the built-in mechanism.

### Testing
To test the changes, we'll be using an issue reported in this ticket: https://biomage.atlassian.net/browse/BIOMAGE-1820. You can use a new project or an existing one.

**Using a new project**
1. Create a new project
2. Upload some sample
3. Run data processing. Wait until the data is processed. Take note of the data processing URL.
4. Go back to Data Management.
5. Upload a new sample to the project.
6. DO NOT RUN DATA PROCESSING. Instead, visit the URL directly.

**Using existing project**
1. Visit Data Processing. Take note of the URL.
2. Go back to Data Management.
3. Upload a new sample to the project.
4. DO NOT RUN DATA PROCESSING. Instead, visit the URL directly.

You will see that the "It's not you it's us" error page will be shown. A new error log will be reported in the #error-logs channel. You can also trigger this error in ui-default and compare.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1849

#### Link to staging deployment URL (or set N/A)
https://ui-agi-grumpy-baboon.scp-staging.biomage.net/

#### Links to any PRs or resources related to this PR
N/A

#### Integration test branch
master

# Merge checklist
Your changes will be ready for merging after all of the steps below have been completed.

### Code updates
Have best practices and ongoing refactors being observed in this PR
- [ ] Migrated any selector / reducer used to the new format.

### Manual/unit testing
- [ ] Tested changes using InfraMock locally **or** no tests required for change, e.g. Kubernetes chart updates.
- [ ] Validated that current unit tests for code work as expected and are sufficient for code coverage **or** no unit tests required for change, e.g. documentation update.
- [ ] Unit tests written **or** no unit tests required for change, e.g. documentation update.

### Integration testing
**You must check the box below** to run integration tests on the latest commit on your PR branch.
Integration tests have to pass before the PR can be merged. Without checking the box, your PR
**will not pass** the required status checks for merging.

- [x] Started end-to-end tests on the latest commit.

### Documentation updates
- [ ] Relevant Github READMEs updated **or** no GitHub README updates required.
[...]

PR #695, opened 2022-04-25
[BIOMAGE-1849] - Add environment to error message

aerlaut:
# Description
PR #690 enables error logs to be reported for both the production and staging environment. The PR also includes the environment the error is generated on in the error log. This PR adds that information to the post that is sent to make it easier to distinguish between staging and production errors.

To test, see the **Testing** guide in PR #690.

The changes introduced in this PR vs the old one can be seen here: 

image

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1849

https://ui-agi-stealthy-civet.scp-staging.biomage.net/

PR #690 

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/utils/postErrorToSlack.js
+++ b/src/utils/postErrorToSlack.js
@@ -35,10 +35,8 @@ const trimOutput = (key, item) => {
 
 const buildErrorMessage = (errorObject, reduxState, context) => {
   const {
-    user, timestamp, experimentId, url,
+    user, timestamp, experimentId, url, environment,
   } = context;
-
-  const { networkResources: { environment } } = reduxState;
 
   let message = `
     [${environment.toUpperCase()}] Uncaught UI Error - Exp ${experimentId} - ${timestamp}
@@ -63,11 +61,11 @@ const buildErrorMessage = (errorObject, reduxState, context) => {
 
 const postError = async (errorLog, context) => {
   const {
-    user, timestamp, experimentId, url,
+    user, timestamp, experimentId, url, environment,
   } = context;
 
   const message = `
-  \u26A0  Uncaught UI Error - ExpID ${experimentId} - ${timestamp}
+  \u26A0  [${environment.toUpperCase()}] Uncaught UI Error - ExpID ${experimentId} - ${timestamp}
   URL: ${url}
 
   User: ${user.attributes.name} <${user.attributes.email}> ${user.username}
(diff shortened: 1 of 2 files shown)


src/utils/postErrorToSlack.js
```diff
@@ -110,12 +108,14 @@ const postErrorToSlack = async (errorObject, reduxState) => {
...
   const url = window.location.href;
   const experimentId = extractExperimentId(url);
+  const { networkResources: { environment } } = reduxState;
 
   const context = {
     user,
     timestamp,
     experimentId,
     url,
+    environment,
```
ivababukova, line 116: instead of passing `environment` in here as an argument, why don't you move `const { networkResources: { environment } } = reduxState;` to `buildErrorMessage`?

aerlaut: `const { networkResources: { environment } } = reduxState;` was inside `buildErrorMessage`. This limits `environment` to be only used inside the error log (the file, not the message).

I think it's good to also show the environment the error message (generated which is generated by `postError`, line 122). So I moved `const { networkResources: { environment } } = reduxState;` to the parent context and put `environment` in `context`. This allows the error message to also show the environment.

ivababukova: I see. What's the point of having all this duplicate information? You can see in the post, user name, experiment link and environment are shown twice:
Screenshot 2022-04-25 at 17 09 47
I think it will be neater of we only show the `errorObject.stack` in that collapsible panel in slack, and environment, user and experiment link are kept at the top of the alert, outside of the collapsible panel.

aerlaut: That's not a collapsible panel - it's a preview of the content of the error log file, which is a text attachment. If you'd like to take a deeper look at the error (e.g. check the redux state), then you can download the file. I put the same info in the error log file as in the message so that the error log file contains all information about the error, and you'd be able to debug the error without having to go back to the message.

aerlaut: If the error message contains too much information, maybe we can remove the URL and User part of the message? We can still see those parts inside the preview. That would leave the error log to be like this: 

image

ivababukova: I prefer not to remove them from the message, because they are not clickable inside this text attachment. If we remove them, better to remove them from the text attachment. It is a recommendation, not compulsory change though.

ivababukova (approved):

src/utils/postErrorToSlack.js
aerlaut: I understand your concern. I'll move the `== DETAILS ==` part below the `== ERROR ==` part so that we can see the error details at a glance without having to open the panel.

PR #697, opened 2022-04-28
[BIOMAGE-1858] Implement samples load for api v2

StefanBabukov:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1858

https://ui-grasp6-trippy-yak.scp-staging.biomage.net/
N/A (running in api PR)

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+  const buildApiv1FilesObject = (fileObject) => {
+    const fileNames = [];
+    const apiV1FileObject = {};
+
+    Object.keys(fileObject).forEach((key) => {
+      const fileName = fileObject[key].s3Path.split('/').pop();
+      fileNames.push(fileName);
+
+      apiV1FileObject[fileName] = {
+        path: fileObject[key].s3Path,
```
cosa65: Just curious, what happens if we don't create the file with `path`? 

I think it isn't used anywhere anymore, so if we don't need it, it might be a good idea to assign `undefined` to it or something just to make it easy when we refactor the UI in the end.

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
 import {
   SAMPLES_LOADED,
   SAMPLES_ERROR,
   SAMPLES_LOADING,
 } from 'redux/actionTypes/samples';
 
+const toApiV2 = (samples) => {
+  const apiV2Samples = [{ samples: {} }];
+
+  const buildApiv1FilesObject = (fileObject) => {
```
cosa65: I would just call these `file` to avoid confusion, there's something else called `fileObject` inside the current session uploaded `files` that is used to retry uploads.

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+    const fileNames = [];
+    const apiV1FileObject = {};
+
+    Object.keys(fileObject).forEach((key) => {
+      const fileName = fileObject[key].s3Path.split('/').pop();
```
cosa65: The name shouldn't be taken from the `s3Path`, in fact we don't really need to use s3Path anywhere in the UI.

The fileName can be guessed from the fileType (I do that translation in processUploadV2):
features10x -> 'features.tsv.gz'
barcodes10x -> 'barcodes.tsv.gz'
matrix10x -> 'matrix.mtx.gz'

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+      fileNames.push(fileName);
+
+      apiV1FileObject[fileName] = {
+        path: fileObject[key].s3Path,
+        size: fileObject[key].size,
+        valid: fileObject[key].valid,
```
cosa65: We shouldn't use `valid`. 
Here you can just always write "valid" because if a file was created it means it already passed the validity checks. 
`valid` is a property that wasn't cleaned up when it became redundant.

src/redux/actions/samples/loadSamples.js
cosa65: Also change any other references from `fileObject` to `file`, not only here

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+        },
+      };
+    });
+    return { apiV1FileObject, fileNames };
+  };
+
+  const sampleTechnologyConvert = (technology) => {
+    switch (technology) {
+      case '10x': return '10X Chromium';
+      default: return technology;
```
cosa65: There's something wrong if the technology is not recognized, so we can just throw an error

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+        upload: {
+          status: fileObject[key].uploadStatus,
+        },
+      };
+    });
+    return { apiV1FileObject, fileNames };
+  };
+
+  const sampleTechnologyConvert = (technology) => {
+    switch (technology) {
```
cosa65: For two cases wouldn't it be easier to use if else?

And if we have many cases in the future we could probably just store all possibilities in an object instead of using a switch

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+      case '10x': return '10X Chromium';
+      default: return technology;
+    }
+  };
+
+  samples.data.message.forEach((sample) => {
+    const { apiV1FileObject, fileNames } = buildApiv1FilesObject(sample.files);
+    apiV2Samples[0].samples[sample.id] = {
```
cosa65: Why do we store this in `[0]`?

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,79 @@
...
+  };
+
+  const sampleTechnologyConvert = (technology) => {
+    switch (technology) {
+      case '10x': return '10X Chromium';
+      default: return technology;
+    }
+  };
+
+  samples.data.message.forEach((sample) => {
```
cosa65: Why does the structure have a `.message`?

cosa65 (changes requested):

src/redux/actions/samples/loadSamples.js
StefanBabukov: because that is how the apiv1 object looks like and i wanted to mimick it so its exactly the same format

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,84 @@
...
+    const apiV1Files = {};
+
+    Object.keys(files).forEach((key) => {
+      const fileType = files[key].sampleFileType;
+      let fileName = '';
+      if (fileType.includes('features')) {
+        fileName = 'features.tsv.gz';
+      } else if (fileType.includes('barcodes')) {
+        fileName = 'barcodes.tsv.gz';
+      } else if (fileType.includes('matrix')) {
```
cosa65: Rather than `includes` I would compare directly, we know exactly what types are possible: `features10x`, `barcodes10x` and `matrix10x`

src/redux/actions/samples/loadSamples.js
cosa65: I think this wasn't resolved yet, it's about trhowing an error with ("technology not recognized")

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,84 @@
...
   try {
     dispatch({
       type: SAMPLES_LOADING,
     });
 
-    const data = await fetchAPI(url);
+    let data = await fetchAPI(url);
+
+    if (currentApiVersion === api.V2) {
+      data = toApiV2(data);
```
cosa65: This is `toApiV1` or `fromApiV2`, not `toApiV2`

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,84 @@
...
+      name: sample.name,
+      lastModified: sample.updatedAt,
+      files: apiV1Files,
+      type: sampleTechnologyConvert(sample.sampleTechnology),
+      fileNames,
+      uuid: sample.id,
+    };
+  });
+
+  return apiV2Samples;
```
cosa65: `apiV1Samples`

src/redux/actions/samples/loadSamples.js
```diff
@@ -1,22 +1,85 @@
...
+    Object.keys(files).forEach((key) => {
+      const fileType = files[key].sampleFileType;
+      let fileName = '';
+      if (fileType === 'features10x') {
+        fileName = 'features.tsv.gz';
+      } else if (fileType === 'barcodes10x') {
+        fileName = 'barcodes.tsv.gz';
+      } else if (fileType === 'matrix10x') {
+        fileName = 'matrix.mtx.gz';
+      }
```
cosa65: Hey sorry last question about this, now that I look at it wouldn't it be better to have it as an object? 
{[fileType]: fileName}

and then you don't need any ifs

StefanBabukov: oh yeah I'll create it

cosa65 (approved):

PR #704, opened 2022-05-06
Implement run gem2s for apiv2

cosa65:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1871

N/A (in api)

https://github.com/hms-dbmi-cellenics/api/pull/348
https://github.com/hms-dbmi-cellenics/pipeline/pull/230

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/__test__/redux/actions/pipelines/runGem2s.test.js
+++ b/src/__test__/redux/actions/pipelines/runGem2s.test.js
@@ -22,6 +22,7 @@ import { api } from 'utils/constants';
 const mockStore = configureStore([thunk]);
 enableFetchMocks();
 
+jest.mock('config');
 jest.mock('redux/actions/backendStatus/loadBackendStatus',
   () => jest.fn().mockImplementation(() => async () => { }));
 
--- a/src/__test__/redux/actions/pipelines/runGem2s.test.js
+++ b/src/__test__/redux/actions/pipelines/runGem2s.test.js
@@ -16,6 +16,9 @@ import {
 
 import { runGem2s } from 'redux/actions/pipeline';
 
+import config from 'config';
+import { api } from 'utils/constants';
+
 const mockStore = configureStore([thunk]);
 enableFetchMocks();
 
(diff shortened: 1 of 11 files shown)


src/redux/reducers/backendStatus/backendStatusUpdated.js
```diff
@@ -1,13 +1,14 @@
...
-import initialState from './initialState';
+import initialState from 'redux/reducers/backendStatus/initialState';
+
+import mergeObjectWithArrays from 'utils/mergeObjectWithArrays';
 
 const backendStatusUpdated = produce((draft, action) => {
   const { experimentId, status } = action.payload;
 
-  _.merge(draft[experimentId]?.status, status);
+  mergeObjectWithArrays(draft[experimentId]?.status, status);
```
cosa65: When we receive completedSteps array we want to replace the previous one, not merge the two arrays

src/__test__/redux/actions/pipelines/runGem2s.test.js
```diff
@@ -97,4 +101,21 @@ describe('runGem2s action', () => {
 
     expect(fetchMock).toHaveBeenCalled();
   });
+
+  it('Works well with api v2', async () => {
```
kafkasl, line 105: I am considering that we will need to revisit this once we migrate to v2 to remove the lines `config.currentApiVersion = api.V2;` and potentially renaming the tests. Is there a way to avoid that? maybe instead of adding an extra test case and extra file? like `runGem2s.v2.test.js` so when we migrate we only need to delete the other file without changing any code (we could potentially rename this one to `runGem2s.v1.test.js` to make it easier to find and remove them when we are done)

src/redux/reducers/backendStatus/backendStatusUpdated.js
kafkasl: I don't understand what's happening here. I checked the `mergeObjectWithArrays` but it's not documented and the name is not clear to me. I'd appreciate it if you could elaborate and potentially add a line or two in the original `mergeObjectWithArrays` function describing what it does.

cosa65: I think if I had written the function I would call it `mergeObjectReplacingArrays` instead, what do you think? 
Added comment btw

cosa65: I coudl rename it

src/__test__/redux/actions/pipelines/runGem2s.test.js
cosa65: Many of the tests can be reused once we switch to using v2 and in many cases the changes done in the action creator are not many, that's why I feel like that refactor might not be worth the amount of work it involves (since we will also need work to undo it).

src/redux/reducers/backendStatus/backendStatusUpdated.js
aerlaut: I agree with Pol, I think it's better named something like `mergeObjectReplaceArrays`

src/utils/mergeObjectWithArrays.js
```diff
@@ -1,5 +1,10 @@
 import _ from 'lodash';
 
+/**
+ * When this merge runs into objects, it replaces the original array with the new one,
```
aerlaut: This function merges objects, replacing old arrays in its properties with the new one, unlike `_.merge` which combines old and new arrays.

aerlaut (approved):

PR #708, opened 2022-05-09
[BIOMAGE-1925] - [OLD] Shuffle plot data points

aerlaut:
# Description
#### Problem
Data points in Vega plots are rendered in sequence. This means later cells. This poses a problem because later cells, which maybe in a group, are rendered over the earlier cells, making this later group to look "over-represented" in the plot.

#### Solution
Shuffle plot data points

Before shuffling (currently):
before shuffling

After shuffling:
after shuffling

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1925

N/A

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/components/data-exploration/embedding/Embedding.jsx
+++ b/src/components/data-exploration/embedding/Embedding.jsx
@@ -7,7 +7,6 @@ import {
 import PropTypes from 'prop-types';
 import * as vega from 'vega';
 import 'vitessce/dist/es/production/static/css/index.css';
-import { shuffle } from 'lodash';
 
 import ClusterPopover from 'components/data-exploration/embedding/ClusterPopover';
 import CrossHair from 'components/data-exploration/embedding/CrossHair';
@@ -282,7 +281,7 @@ const Embedding = (props) => {
             uuid={embeddingType}
             viewState={view}
             updateViewInfo={updateCellCoordinates}
-            cells={shuffle(convertCellsData(data, cellSetHidden, cellSetProperties))}
+            cells={convertCellsData(data, cellSetHidden, cellSetProperties)}
             mapping='PCA'
             cellSelection={selectedCellIds}
             cellColors={
--- a/src/components/data-exploration/embedding/Embedding.jsx
+++ b/src/components/data-exploration/embedding/Embedding.jsx
@@ -7,6 +7,7 @@ import {
 import PropTypes from 'prop-types';
 import * as vega from 'vega';
 import 'vitessce/dist/es/production/static/css/index.css';
+import { shuffle } from 'lodash';
 
 import ClusterPopover from 'components/data-exploration/embedding/ClusterPopover';
 import CrossHair from 'components/data-exploration/embedding/CrossHair';
@@ -281,7 +282,7 @@ const Embedding = (props) => {
             uuid={embeddingType}
             viewState={view}
             updateViewInfo={updateCellCoordinates}
-            cells={convertCellsData(data, cellSetHidden, cellSetProperties)}
+            cells={shuffle(convertCellsData(data, cellSetHidden, cellSetProperties))}
             mapping='PCA'
             cellSelection={selectedCellIds}
             cellColors={
--- a/src/components/data-exploration/embedding/Embedding.jsx
+++ b/src/components/data-exploration/embedding/Embedding.jsx
@@ -1,5 +1,7 @@
 // eslint-disable-file import/no-extraneous-dependencies
-import React, { useState, useEffect, useRef } from 'react';
+import React, {
+  useState, useEffect, useRef,
+} from 'react';
 import dynamic from 'next/dynamic';
 import {
   useSelector, useDispatch,
@@ -15,7 +17,7 @@ import PlatformError from 'components/PlatformError';
 import Loader from 'components/Loader';
 
 import { loadEmbedding } from 'redux/actions/embedding';
-import { getCellSetsHierarchyByType, getCellSets } from 'redux/selectors';
+import { getCellSetsHierarchyByType, getCellSetsHierarchyByKeys, getCellSets } from 'redux/selectors';
 import { createCellSet } from 'redux/actions/cellSets';
 import { loadGeneExpression } from 'redux/actions/genes';
 import { updateCellInfo } from 'redux/actions/cellInfo';
@@ -26,6 +28,8 @@ import {
   renderCellSetColors,
   colorByGeneExpression,
   colorInterpolator,
+  generateCellIdMapping,
+  reorderObjectValues,
 } from 'utils/plotUtils';
 import getContainingCellSetsProperties from 'utils/cellSets/getContainingCellSetsProperties';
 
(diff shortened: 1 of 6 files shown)


gerbeldo:
Looking at plots I'd say it works! It's the same embedding but unshuffled/shuffled

No shuffle
Screenshot 2022-05-23 at 16 47 32

Shuffle
Screenshot 2022-05-23 at 16 47 45

alexvpickering:
@aerlaut Did you look into doing this for the `Scatterplot` as well?

gerbeldo (approved):

alexvpickering (changes requested):
Looks great Agi - Thanks!

This also needs to be implemented in the Scatterplot for consistency

aerlaut:
> Looks great Agi - Thanks!
> 
> This also needs to be implemented in the Scatterplot for consistency

You mean for the Embedding in Data Exploration? The rendering in the embedding is already shuffled, as can be seen in this screenshot:
Screenshot 2022-05-23 at 12 14 39
 
As you can see, some of the blue circles are below the red circles, while other blue circles are above the red. So rendering of the embedding is not order-dependent,

alexvpickering:
Are you plotting by sample? Here is what I see in production -- unless the changes you made affect the Scatterplot as well?

find me a blue point above a yellow one 🙃:

image


Another region:

image


Different samples:

image

aerlaut:
Ah I see. I didn't show the embedding by samples. I've implemented the shuffling for the embedding. Can you please have a look again @alexvpickering?

src/utils/plotUtils.js
```diff
@@ -105,4 +133,6 @@ export {
   colorInterpolator,
   hexToRgb,
   convertRange,
+  generateCellIdMapping,
+  reorderObjectValues,
```
ivababukova, line 137: I see that these two functions are only used in Embedding.jsx. In this case, it is better to have them in there, rather than in a different file.

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -76,6 +81,9 @@ const Embedding = (props) => {
   const [selectedIds, setSelectedIds] = useState(new Set());
   const [cellColors, setCellColors] = useState({});
   const [cellInfoVisible, setCellInfoVisible] = useState(true);
+  const [cellIndexMapping, setCellIndexMapping] = useState(null);
+  const [getCellSetCellId, setGetCellSetCellId] = useState(null);
+  const [getEmbeddingCellIdx, setGetEmbeddingCellIdx] = useState(null);
```
ivababukova, line 86: don't use "get" in the names, because it is confusing -- when I saw it, I thought that these are not variables, but rather getter methods. Also, what are these? What is `getCellSetCellId`, `cellIndexMapping` and `getEmbeddingCellIdx` and why do we need to have a state variable for each of these?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -1,5 +1,7 @@
 // eslint-disable-file import/no-extraneous-dependencies
-import React, { useState, useEffect, useRef } from 'react';
+import React, {
```
ivababukova, line 2: the changes in this `Embedding.jsx` file seem to be much more than just shuffling points. Can you explain what is happening here?

src/components/data-exploration/embedding/Embedding.jsx
```diff
@@ -84,6 +92,20 @@ const Embedding = (props) => {
...
+    const allCellIds = sampleKeys
+      .reduce((acc, sampleKey) => [...acc, ...cellSetProperties[sampleKey].cellIds], [])
+      .filter((cellId) => cellId !== undefined);
+
+    const { forwardMap, reverseMap } = generateCellIdMapping(allCellIds);
+
+    setGetCellSetCellId(() => (embeddingCellIdx) => reverseMap[embeddingCellIdx]);
+    setGetEmbeddingCellIdx(() => (cellSetCellId) => forwardMap[cellSetCellId]);
+    setCellIndexMapping(forwardMap);
+  }, [sampleKeys?.length]);
```
ivababukova, line 107: why is this useEffect necessary?

ivababukova (changes requested):
Can you explain why all these changes in Embedding.jsx are necessary and what are you aiming with them? Looking at this code, it is not clear to me. After that, let's see if we can make the code clearer.

alexvpickering:
> Ah I see. I didn't show the embedding by samples. I've implemented the shuffling for the embedding. Can you please have a look again @alexvpickering?

The samples in the embedding are shuffled now, but it has introduced a bug where the expression isn't associated with the correct cells. IGKC in staging:

image

IGKC in production (same samples):

image

Can you also add a test for this if there isn't one already? Thanks!

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: This `useEffect` block wraps the code which generates the mapping between cellIds that we use in the platform vs the cellIds that we is used for the embedding. The block depends on the number of samples - which shouldn't change. This ensures that the mapping is only calculated once per page render.

aerlaut:
> Can you explain why all these changes in Embedding.jsx are necessary and what are you aiming with them? Looking at this code, it is not clear to me. After that, let's see if we can make the code clearer.

These changes are made so that the cells are "shuffled" when we view the embedding, so that no one group of cells are over-represented in the plot. This is because later cells are rendered on top earler cells. This is very evident when looking at the embedding by sample.

The embedding takes in an object with cellIds as the key and the cell coordinate as the value. The coordinates are used to display the cell. This data is derived from the embedding.

    {
        0: { mapping: {PCA : [1, 2] }},
        1 : { mapping: {PCA:  [3, 4] }}
        ....
    }

When we hover over the Embedding, the cellId (e.g. key) is returned. We use the cellId to access values in other redux stores to display information (e.g. cluster or gene expression). Therefore, I can not just shuffle the cells, because shuffling the cells will mess up the association between the coordinates and the cell id. For example, if I shuffle the 2 cells above into

    {
        0: { mapping: {PCA : [3, 4] }},
        1 : { mapping: {PCA:  [1, 2] }}
        ....
    }

when I hover over the spot (3,4), I'll get 0 as the cell id. If I use this id to get data associated with the cell (e.g. expression), I'll get the wrong value, because the cell on (3,4) should be cell 1.

Therefore, to be able to "shuffle" the cells and still correctly retrieve cellId, I created a mapping to map the cellIds to a different set of ids that'll be used by the Scatterplot. I also created a reverse mapping to make reverse lookup easier. This mapping and the reverse is used to translate the ids back and forth between the cell id and the id used in the Embedding. When referencing a cell in the embedding using the cell id, we have to translate it to the id used in the embedding using the mapping. When receiving an id from the embedding (e.g. when hovering over a cell), we have to translate it back to the correct cell id using the reverse.

src/components/data-exploration/embedding/Embedding.jsx
aerlaut: `cellIndexMapping` is used to store the calculated mapping between cellId and the id that is used in the embedding. `getCellSetCellId` is a function that is used to map a cellId to the embedding id. `getEmbeddingCellIdx` is a function that is used to map back embedding id to the corresponding cellId.

I put all these values in a state so that they can all be calculated once in the `useEffect` body and then used for any interactions without calculating again.

aerlaut:
There is an alternative implementation in UI PR #714 which fixes things more elegantly. Instead of handling shuffling in the UI, shuffling of the cellids is implemented in [worker PR 236](https://github.com/hms-dbmi-cellenics/pipeline/pull/236). The changes in UI PR #714 introduces changes in the UI to allows the shuffling to come through in the UI. For better explanation, see the PR.

PR #714, opened 2022-05-26
[BIOMAGE-1925] - Shuffle plot data points

aerlaut:
# Description
**Problem**
Data points in Vega plots are rendered in sequence. This means later cells. This poses a problem because later cells, which maybe in a group, are rendered over the earlier cells, making this later group to look "over-represented" in the plot.

**Solution**
Right now, the categorical and continuous plots generates data by iterating through the groups (e.g. samples/cluster) and appending the cell ids. This causes "ordering" of cell ids by sample/cluster . In code: 
   
    // sample1 = { cellIds: [1, 10]}
    // sample2 = { cellIds: [0, 5 ]}

    const cellIds = []
    samples.forEach((sample) => { cellIds.push(...sample.cellIds) }

    // Result:
    cellIds = [
    1, // from sample 1
    10, // from sample 1
    0, // from sample 2
    5, // from sample 2
    .....
    ]

The changes in this PR uses the order of cells in the embedding data returned from the pipeline. These cells are shuffled. In code:

    // Contains the x,y coordinate of each cell. Null means data is filtered in QC.
    // embeddingData = [[1, 2], null, [5, 6]]
    const data = []
    embeddingData.forEach((coordinates, idx) => { 
        if (!coordinates) return

        data.push({
            cellId: idx,
            coordinates
        }) 
    }

    // Result
    data = [
        {cellId: 0}, // e.g. from sample 2
        {cellId: 2}, // eg.  from sample 1
        ...
    ]

Therefore, the order of rendering follows the order of the shuffled cells generated in the pipeline.

The Embedding in data-exploration is not affected by this issue because it uses the cell ordering as returned in the embeddingData. Therefore, just shuffing the data as done in [pipeline PR #236](https://github.com/hms-dbmi-cellenics/pipeline/pull/236) is enough without further changes in the UI.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1925

https://ui-agi-squeaky-termite.scp-staging.biomage.net/

https://github.com/hms-dbmi-cellenics/pipeline/pull/236

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
+++ b/src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
@@ -232,8 +232,8 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
 
     const inCellSet = colorToCellIdsMap.find((map) => map.cellIds.has(cell.cellId));
 
-    // If cell is not in the cell set, then return null
-    if (!inCellSet) return null;
+    // If cell is not in the cell set, then return
+    if (!inCellSet) return acc;
 
     const { key, name, color } = inCellSet;
 
--- a/src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
+++ b/src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
@@ -227,7 +227,9 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
   let cellSetLegendsData = [];
   const addedCellSetKeys = new Set();
 
-  filteredCells = filteredCells.map((cell) => {
+  filteredCells = filteredCells.reduce((acc, cell) => {
+    if (!cell) return acc;
+
     const inCellSet = colorToCellIdsMap.find((map) => map.cellIds.has(cell.cellId));
 
     // If cell is not in the cell set, then return null
@@ -240,15 +242,15 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
       cellSetLegendsData.push({ key, name, color });
     }
 
-    return {
+    acc[cell.cellId] = {
       ...cell,
       cellSetKey: key,
       cellSetName: name,
       color,
     };
-  });
 
-  filteredCells = filteredCells.filter((cell) => cell !== null);
+    return acc;
+  }, {});
 
   // Sort legends to show them in the order that cellSetKeys are stored
   cellSetLegendsData = _.sortBy(
@@ -263,16 +265,16 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
 const generateData = (cellSets, sampleKey, groupBy, embeddingData) => {
   const { filteredCells, cellSetLegendsData } = filterCells(cellSets, sampleKey, groupBy);
 
-  const plotData = filteredCells
-    .filter((d) => d.cellId < embeddingData.length)
-    .filter((data) => embeddingData[data.cellId]) // filter out cells removed in data processing
+  const plotData = embeddingData
+    .map((coordinates, cellId) => ({ cellId, coordinates }))
+    .filter(({ coordinates }) => coordinates !== undefined)
     .map((data) => {
-      const { cellId, ...supportingData } = data;
+      const { cellId, coordinates } = data;
 
       return {
-        ...supportingData,
-        x: embeddingData[cellId][0],
-        y: embeddingData[cellId][1],
+        ...filteredCells[cellId],
+        x: coordinates[0],
+        y: coordinates[1],
       };
     });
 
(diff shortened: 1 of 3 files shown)


src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
```diff
@@ -227,7 +227,9 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
   let cellSetLegendsData = [];
   const addedCellSetKeys = new Set();
 
-  filteredCells = filteredCells.map((cell) => {
+  filteredCells = filteredCells.reduce((acc, cell) => {
+    if (!cell) return acc;
```
cosa65, line 231: What is this guard for? In which cases would we run into a null/undefined cell?

cosa65: Oh I see it's to do everything in one iteration instead of a filter after.

cosa65: Could you write a small comment with something like "Ignore cells that were filtered"?

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
```diff
@@ -227,7 +227,9 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
   let cellSetLegendsData = [];
   const addedCellSetKeys = new Set();
 
-  filteredCells = filteredCells.map((cell) => {
+  filteredCells = filteredCells.reduce((acc, cell) => {
```
cosa65, line 230: From what I'm getting this wasn't part of the changes, it was a performance improvement, right?

I was thinking, using a `reduce` vs a `map` + `filter` is less declarative. I feel like perhaops we shudl consider using map and filter when this doesn't improve the performance much.

I was reading [this](https://www.davideasaf.com/mapfilter-vs-reduce/) a little bit which I think fits this case.

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
```diff
@@ -263,16 +265,16 @@ const filterCells = (cellSets, sampleKey, groupBy) => {
...
-      const { cellId, ...supportingData } = data;
+      const { cellId, coordinates } = data;
 
       return {
-        ...supportingData,
-        x: embeddingData[cellId][0],
-        y: embeddingData[cellId][1],
+        ...filteredCells[cellId],
+        x: coordinates[0],
+        y: coordinates[1],
```
cosa65, line 277: Nice! way more readable and fixes the issue.
Do you know what was the purpose of `.filter((d) => d.cellId < embeddingData.length)` by the way?

aerlaut: I think it was just a short cut to truncate data that's longer than the embedding data returned from the worker.

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
aerlaut: A `map` + `filter` produces arrays, whereas in this case I'm making a dictionary object with the `cellId` as the property, which would make it much easier and faster to access the properties of a cell later on down the line.

This reduce can be replaced with a `forEach` which adds to a resulting object at every iteration. What do you think?

aerlaut:
@alexvpickering the staging link is working, can you please check it out?

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
cosa65: Ah ok, never mind then.

About using a `forEach` or a `reduce` I think both options are fine, if you want to keep it like this I understand now and it makes sense

src/utils/plotSpecs/generateEmbeddingCategoricalSpec.js
cosa65: yes, although it is still weird, why would we need to perform the truncation

cosa65 (approved):
lgtm

alexvpickering:
> @alexvpickering the staging link is working, can you please check it out?

Shuffling looks good. Only comment is that a random shuffling (different from what comes from worker) is equally acceptable if performance is an issue for large datasets or it is easier to maintain. I'll leave that to you to choose though :)

aerlaut:
> Shuffling looks good. Only comment is that a random shuffling (different from what comes from worker) is equally acceptable if performance is an issue for large datasets or it is easier to maintain. I'll leave that to you to choose though :)

I think it's better implemented this way as shuffling is done only once in the pipeline and we don't have to do it again in the UI.

PR #737, opened 2022-06-14
Disable dimensionality reduction method except for Seurat v4

Kristian-A:
# Description
Disable dimensionality reduction method options for all data integration methods except Seurat v4, found in the "Data Integration" step of the pipeline. Display _N/A_ if the dimensionality reduction options are not applicable to (disabled for) the method. This is done by adding an additional parameter to each integration method `enableDimReductionMethod: true` if the dimensionality reduction method is relevant to the integration method (can be omitted if the dropdown should be disabled as undefined/null is evaluated to false).

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1959

https://ui-kristian-ui737.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/data-processing/DataIntegration/CalculationConfig.jsx
```diff
@@ -195,49 +242,9 @@ const CalculationConfig = (props) => {
...
-            >
-              <Select
-                value={dimensionalityReduction.method}
-                onChange={(val) => updateSettings({ dimensionalityReduction: { method: val } })}
-                disabled={disabled}
-              >
-                <Option key='rpca' value='rpca'>Reciprocal PCA (RPCA)</Option>
-                <Option key='cca' value='cca'>Canonical Correlation Analysis (CCA)</Option>
-              </Select>
+            {selectedMethod.enableDimReductionMethod ? renderDimReductionMethod() : (<></>)}
```
StefanBabukov: You can just have here `dataIntegration.method === 'seuratv4' ? renderDimReductionMethod() : (<></>)` instead of adding this enableDimReductionMethod variable

Kristian-A: I didn't want to hardcode the name but rather make an attribute that you can use to turn the settings on or off. It is a good solution if some other method would need dimensionality reduction options. In this case it might be an overkill especially if `seuratv4` is the only method that needs it. 

Edit: Actually, even if `seuratv4` is the only method using dimensionality reduction, I think it is better to have it as a setting rather than a hardcoded `if` statement which acts as a side effect to one particular data integration method.

StefanBabukov: yeah thats what i thought that only seuratv4 uses it and it makes sense to add a variable if there were more methods that would have the dimreduction option enabled and not just 1. But I dont think it matters that much, just whether it makes the code more understandable

Kristian-A: Yes, in terms of readability hardcoding it would improve the code. I will change it now

StefanBabukov (approved):
lgtm

PR #742, opened 2022-06-15
Reorder genes plots

jszpila314:
# Description
Added a separate tab under gene selection in the marker heatmap plot to allow reordering of genes.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1918
https://ui-jakub-ui742.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional
- [x] Staging environment is unstaged before merging.


src/components/plots/gene-reorder-tool/GeneReorderTool.jsx
```diff
@@ -0,0 +1,67 @@
+import React, {
+  useEffect, useState,
+} from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+import PropTypes from 'prop-types';
+
+import { arrayMoveImmutable } from 'utils/array-move';
+import { updatePlotConfig } from 'redux/actions/componentConfig';
+import HierarchicalTreeGenes from '../hierarchical-tree-genes/HierarchicalTreeGenes';
```
cosa65: This can be imported with an absolute path instead of relative

src/components/plots/gene-reorder-tool/GeneReorderTool.jsx
```diff
@@ -0,0 +1,67 @@
+import React, {
```
cosa65: I would take this out of the folder and remove `gene-reorder-tool` if there isn't anything else you need to use inside there.

cosa65 (commented):
I took a look overall since you mentioned it is a work in progress. It looks good! 

One suggestion I would make to get rid of the lag between dropping a gene and the list updating would be having `onGeneReorder` call 

```
setGeneTreeData(composeGeneTree(newOrder));
debouncedUpdatePlotConfig(newOrder)
``` 


with `debouncedUpdatePlotConfig` being something like this:
```
const debouncedUpdatePlotConfig = useCallback(
    _.debounce((newOrder) => dispatch(updatePlotConfig(plotUuid, { selectedGenes: newOrder })), time), [],
  );
```
(`time` here can be pretty small probably, it's just about it updating a little bit later).

With these 2 changes it should no longer need to wait for redux to finish updating before setting the gene in its new place in the list, so it might look faster (with the plot updating a little bit later).


If you search for `_.debounce` you should find some places where similar things were done in case you want to check other examples. I think it was done a lot in "Data Processing".

src/components/plots/gene-reorder-tool/GeneReorderTool.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const onGeneReorder = (geneKey, newPosition) => {
+    const oldOrder = geneTreeData.map((treeNode) => treeNode.title);
+
+    const newOrder = arrayMoveImmutable(Object.values(oldOrder), geneKey, newPosition);
+
+    dispatch(updatePlotConfig(plotUuid, { selectedGenes: newOrder }));
+  };
+
+  return (
+    <>
```
ivababukova: there is no need for these empty `<>`, it should work without them

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -1,7 +1,8 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { within } from '@testing-library/dom';
 import userEvent from '@testing-library/user-event';
 import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
-import _ from 'lodash';
+import _, { drop } from 'lodash';
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: When using lodash usually we import only the `_` and then do `_.drop` or whatever you want to use

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -24,6 +25,10 @@ import mockAPI, {
 } from '__test__/test-utils/mockAPI';
 import createTestComponentFactory from '__test__/test-utils/testComponentFactory';
 
+import { arrayMoveImmutable, arrayMoveMutable } from 'utils/array-move';
+import { array } from 'vega';
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: Is this being used? If not, remove this import

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -24,6 +25,10 @@ import mockAPI, {
...
+import { string } from 'prop-types';
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: Is this being used? If not, remove this import

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -24,6 +25,10 @@ import mockAPI, {
 } from '__test__/test-utils/mockAPI';
 import createTestComponentFactory from '__test__/test-utils/testComponentFactory';
 
+import { arrayMoveImmutable, arrayMoveMutable } from 'utils/array-move';
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: I can't see where this one is used. If it isn't, then remove the import

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -1,7 +1,8 @@
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
+import { within } from '@testing-library/dom';
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx:3]: I can't see where this one is used. If it isn't, then remove the import

Comment (cosa65) [reply src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx:3]: Do you have eslint in vsocde? it should help you by showing red lines on unused imports

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -275,4 +286,44 @@ describe('Marker heatmap plot', () => {
...
+    const geneToRemove = screen.getByText(genesListBeforeRemoval[1]);
+    
+    const geneRemoveButton = geneToRemove.nextSibling.firstChild;
+
+    userEvent.click(geneRemoveButton);
+
+    const genesListAfterRemoval = getTreeGenes(geneTree);
+    
+    // The gene should be deleted from the list
+    expect(_.isEqual(genesListAfterRemoval.length, genesListBeforeRemoval.length)).toEqual(false);
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: An idea, you could also remove manually the gene from `genesListBeforeRemoval` and compare it against `genesListAfterRemoval` expecting them to be equal

src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx
```diff
@@ -275,4 +286,44 @@ describe('Marker heatmap plot', () => {
...
+    // Check that initially there are 5 marker genes - the default
+    markerGenesData5.order.forEach((geneName) => {
+      expect(screen.getByText(geneName)).toBeInTheDocument();
+    });
+
+    const geneTree = screen.getByRole('tree');
+
+    // Introduce testing for drag and drop if possible
+    // In Jest (jsdom) DragEvent and properties to do with rendering (widths, coordinates)
+    // don't exist -> not possible to test?
```
Comment (cosa65) [inline src/__test__/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.test.jsx]: If you want to test it one option could be to get the `onDrop` prop your `HierarchicalTreeGenes` component passed to `Tree` and then trigger it yourself. 

```
   const tree = component.find('HierarchicalTreeGenes Tree');
    tree.getElement().props.onDrop(dropInfo);
    component.update();
```

It would involve using enzyme (which kind of breaks a little bit the idea/philosophy of using rtl, but since rtl doesn't support drag and drop I think it makes sense in this case).

This is just and idea because I read this comment, but it's not necessarily something you need to do.

src/components/plots/GeneReorderTool.jsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const dispatch = useDispatch();
+
+  const config = useSelector((state) => state.componentConfig[plotUuid]?.config);
+
+  const experimentId = useSelector((state) => state.componentConfig[plotUuid]?.experimentId);
+
+  const loadedMarkerGenes = useSelector(
+    (state) => state.genes.expression.views[plotUuid]?.data,
+  ) || [];
```
cosa65: Two things:
- In the api we can't but in the ui we have access to the null-coalescing operator (`??`) which is better than `||` because it only works with `null` or `undefined`, so it's more specific.

- In `composeGeneTree` you are handling the case where `treeGenes` is `undefined`/`null`, so I think you don't really need to handle it in the `useSelector` as well, just one of these two places should be ok (since `composeGeneTree` has to deal with more cases maybe it's better to keep the handling there).

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -7,7 +7,7 @@ import {
   Select,
   Radio,
 } from 'antd';
-import _ from 'lodash';
+import _, { update } from 'lodash';
```
Comment (cosa65) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: When using lodash usually we import only the _ and then do _.update or whatever you want to use

src/components/plots/hierarchical-tree-genes/HierarchicalTreeGenes.jsx
```diff
@@ -0,0 +1,87 @@
...
+  const renderXButton = (geneKey) => (
+    <Button
+      type='text'
+      onClick={() => {
+        onNodeDelete(geneKey);
+      }}
+    >
+      <CloseOutlined />
+    </Button>
+  );
```
cosa65: I think I would just move this into `renderTitles` rather than have 2 functions

src/components/plots/hierarchical-tree-genes/HierarchicalTreeGenes.jsx
```diff
@@ -0,0 +1,87 @@
...
+  const renderTitles = (data) => {
+    const toRender = data.map((e) => {
+      // modified needs to be a copy of a given gene
+      const modified = { ...e };
+      modified.title = (
+        <Space>
+          {modified.title}
+          {renderXButton(modified.key)}
+        </Space>
+      );
```
cosa65: Why is this called `modified`? And why are you setting `modified.title =` something that depends on `modified.title`?

ivababukova:
@jszpila314, before implementing a new interface for the genes reorder functionality, did you try to make the existing `Select` component do the reorder?

I was looking at the Select component examples in Antd: https://ant.design/components/select/, this specific type:
Screenshot 2022-06-21 at 19 07 55
I also saw this Stackoverflow post that shows how to make each of the boxes draggable: https://stackoverflow.com/questions/49498684/cannot-get-antd-select-tags-to-be-draggable Have you tried this? Before implementing a new design, it is better to check if we can adjust the existing one to do what we want.

cosa65 (approved):

ivababukova:
My suggestion in my previous comment should work. I did a couple of small trials here: https://codesandbox.io/s/eager-archimedes-6uj4fk?file=/index.js What you need to do is to make each entry in `config.selectedGenes` in the `MarkerGeneSelection` component have a label with `draggable: true` in it. Basically, this thing for each gene entry in `selectedGenes`:

```
label: (
            <div
              onMouseDown={e => {
                e.stopPropagation();
              }}
            >
              <div
                draggable="true"
                onDragStart={() => {
                  message.info("Drag should work");
                }}
              >
                  <p>MY_GENE</p>
              </div>
            </div>
          )
```
You can create a new state variable `transformedSelectedGenes` that has this structure:
```
  const [transformedSelectedGenes, setTransformedSelectedGenes] = useState([]);

```
and have a function that populates the entries in this new variable and call it:
```
const populateEntries = () => {
  const newGenes = []
  
  config.selectedGenes.forEach((geneName) => {
    newGenes.append(
      {
              key: geneName,
              label: (
                <div
                  onMouseDown={e => {
                    e.stopPropagation();
                  }}
                >
                  <div
                    draggable="true"
                    onDragStart={() => {
                      message.info("Drag should work");
                    }}
                  >
                      <p>geneName</p>
                  </div>
                </div>
              )
     })
  })
}

setTransformedSelectedGenes(newGenes);
}
```

Then you pass `transformedSelectedGenes` to be the value of `Space` in `MarkerGeneSelection`:

```
value={transformedSelectedGenes}
```

And you should get the same UI, but draggable and reorderable. Afterwards, you will need to take a look and see if the backend takes this order into an account and if not, fix it.

This is just a pseudo-code, it won't compile if you paste it, but I hope that it helps with getting the idea of what I mean. I think that it is worth a try, it should work. There is no point implementing a brand new separate interface just for ordering genes if we can make the existing interface reorder them.

ivababukova (approved):

PR #746, opened 2022-06-20
Lazyload samples

Kristian-A:
# Description
Loads the samples of an experiment only when a user opens the experiment rather than loading all samples at once. This should substantially improve the performance of the ui especially when logging in as an admin, which has access to all experiments.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1985

https://ui-kristian-ui746.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/pages/data-management/index.jsx
+++ b/src/pages/data-management/index.jsx
@@ -16,6 +16,7 @@ import ProjectsListContainer from 'components/data-management/ProjectsListContai
 import ProjectDetails from 'components/data-management/ProjectDetails';
 import { loadProcessingSettings } from 'redux/actions/experimentSettings';
 import loadBackendStatus from 'redux/actions/backendStatus/loadBackendStatus';
+import { loadSamples } from 'redux/actions/samples';
 
 const DataManagementPage = () => {
   const dispatch = useDispatch();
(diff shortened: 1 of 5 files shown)


src/pages/data-management/index.jsx
```diff
@@ -58,6 +63,9 @@ const DataManagementPage = () => {
 
     dispatch(loadProcessingSettings(activeExperimentId));
 
+    console.log(areSamplesLoaded());
+    if (!areSamplesLoaded()) dispatch(loadSamples(null, activeProjectUuid));
+
```
kafkasl, line 67: after migrating to SQL experimentId = projectUuid, so the `loadSamples` function can be simplified and use only a single parameter iirc

Kristian-A: Yes, it worked. I removed the parameter only from the function calls because inside the function there was some api v1 code that uses it. I figured it will get removed anyways when the v1 code gets deleted.

src/pages/data-management/index.jsx
```diff
@@ -58,6 +63,9 @@ const DataManagementPage = () => {
 
     dispatch(loadProcessingSettings(activeExperimentId));
 
+    console.log(areSamplesLoaded());
```
aerlaut: Don't forget to remove this

aerlaut:
Awesome fix :) I tested in staging and the lazy loading works well. Btw, I noticed that while the sample is loading, it is hovering above an empty samples table (see picture).

image

Can you change it so that it only shows the loader when loading, and then displayes the samples table once the samples are loaded? I think it'll look better.

aerlaut (approved):

PR #748, opened 2022-06-23
[BIOMAGE-1934] Remove account specific hardcodings in the UI code

StefanBabukov:
# Description

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1934

https://ui-grasp6-muzzy-emu.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/utils/awsConfig.js
```diff
@@ -0,0 +1,16 @@
...
+  }
+  return accountID;
+};
+
+const getAWSRegion = () => {
+  const region = process.env.AWS_REGION || process.env.DEFAULT_REGION || 'eu-west-1';
+  return region;
+};
+
+export { getAccountId, getAWSRegion };
```
ivababukova: this won't work because of server/client side rendering -- we tried with Agi a few weeks ago. To get the environment, you will need to use an existing selector in redux for this.

ivababukova: it works locally, but as soon as you deploy it in staging, you will see that the environment is set wrongly to development

src/utils/apiEndpoint.js
```diff
@@ -10,7 +10,7 @@ const getApiEndpoint = (location) => {
       return 'http://localhost:3000';
     }
 
-    return 'https://api.scp.biomage.net';
+    return process.env.API_ENDPOINT_PRODUCTION || 'https://api.scp.biomage.net';
```
ivababukova: don't set `'https://api.scp.biomage.net'` as default -- if the endpoint is not defined, raise an error. We don't want some non-Biomage deployments to start misbehaving weirdly, because they are pointing to Biomage.

ivababukova: where is this `API_ENDPOINT_PRODUCTION` defined? I can see that the ui pod has the `DOMAIN_NAME`  environment variable. Use DOMAIN_NAME instead, adding in `api` at the front of it.

src/utils/awsConfig.js
StefanBabukov: yeah thats what i do, I pass the environment from redux into this function?

src/utils/awsConfig.js
```diff
@@ -0,0 +1,20 @@
...
+
+const getAccountId = () => {
+  let accountID;
+  const environment = useSelector((state) => state.networkResources.environment);
+
+  if (environment === 'development') {
+    accountID = '000000000000';
+  } else {
+    accountID = '242905224710';
+  }
```
ivababukova: don't do that -- we can't be hardcoding account id like this. You can use aws cli to get the account id, see this function that we use in the api:

Screenshot 2022-06-27 at 10 54 09

ivababukova: Alternatively, if that doesn't work, add an account id environment variable to the ui pod

src/utils/awsConfig.js
StefanBabukov: Yeah I'm just passing the variable to the function (which will be from redux in the places that call this function) so it should work?

src/utils/getAWSRegion.js
```diff
@@ -0,0 +1,6 @@
+const getAWSRegion = () => {
+  const region = process.env.AWS_REGION || process.env.DEFAULT_REGION || 'eu-west-1';
```
ivababukova: can you remove the default region hardcoding to 'eu-west-1' in here, make sure region is defined when in development environment, and raise an error in here if region is not defined?

src/utils/getAWSRegion.js
```diff
@@ -0,0 +1,6 @@
+const getAWSRegion = () => {
```
ivababukova, line 1: Does that function even work in staging? I think you should rather use the `networkResources` part of the redux store and add region there?

src/utils/apiEndpoint.js
```diff
@@ -10,7 +10,7 @@ const getApiEndpoint = (location) => {
       return 'http://localhost:3000';
     }
 
-    return 'https://api.scp.biomage.net';
+    return `https://api.${process.env.DOMAIN_NAME}`;
```
ivababukova: Does that work in staging? Have you checked that `process.env.DOMAIN_NAME` is defined correctly?

src/utils/getAWSRegion.js
StefanBabukov: Okay so maybe i can make this happen in the ssr along with the environment

src/utils/apiEndpoint.js
StefanBabukov: yes, I checked in lens and the pod config i can see that DOMAIN_NAME in staging is - scp-staging.biomage.net and production - scp.biomage.net

ivababukova (approved):
If it works in staging, LGTM.

PR #759, opened 2022-07-05
[BIOMAGE-2001] Replace gene selection

jszpila314:
# Description
Replace selection box in gene selection in Plots and Tables with search box and gene re-ordering tree. Limit the height of the tree and scroll the tree automatically when element is hovered near the bottom or the top.
These changes affect the Marker Heatmap and the Dot Plot.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-2001

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -252,6 +249,12 @@ const MarkerHeatmap = ({ experimentId }) => {
     dispatch(loadPaginatedGeneProperties(experimentId, ['dispersions'], searchBarUuid, state));
   }, []);
 
+  const treeScrollable = document.querySelector('div#ScrollWrapper');
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx]: Because you are querying an id, which should be unique in the document, you can use `document.getElementById()` which is faster (`https://gomakethings.com/javascript-selector-performance/`)

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -252,6 +249,12 @@ const MarkerHeatmap = ({ experimentId }) => {
...
+
+  useEffect(() => {
+    ScrollOnDrag();
+  }, [treeScrollable]);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:256]: Why are you putting it here? Will does the variable change?

src/components/plots/ScrollOnDrag.jsx
```diff
@@ -0,0 +1,44 @@
...
+  // until obj.offsetParent is undefined, needed to compare with event.clientY for drag event
+  do {
+    posTop += elem.offsetTop;
+  } while (elem = elem.offsetParent);
+
+  return posTop;
+};
+
+const ScrollOnDrag = () => {
+  const treeScrollable = document.querySelector('div#ScrollWrapper');
```
aerlaut: Same as the comment above, I think it's better to use `document.getElementById()` if you are querying for an id.

aerlaut: This is the same element that is queried inside the `useEffect` clause right? Instead of fetching it again here, maybe you can pass it as a parameter to the function?

src/components/plots/ScrollOnDrag.jsx
```diff
@@ -0,0 +1,44 @@
...
+  if (!treeScrollable) return;
```
aerlaut: If you pass `treeScrollable` as a parameter to this function, this check can be done inside the `useEffect` before `ScrollOnDrag` is called. That saves caling this function if the value of `treeScrollable` is not set.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -166,9 +167,31 @@ const DotPlotPage = (props) => {
     // If using marker genes, check that the selected number is more than 0
     if (config.useMarkerGenes && config.nMarkerGenes === 0) return false;
 
+    // Skip fetch for initial reorder
+    // if (initialReorder) return false;
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: What is this for? Why is it commented? Is it still relevant? If not I think this can be deleted.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
```diff
@@ -323,42 +327,27 @@ const MarkerHeatmap = ({ experimentId }) => {
...
+        <MarkerGeneSelection
+          config={config}
+          plotUuid={plotUuid}
+          searchBarUuid={searchBarUuid}
+          experimentId={experimentId}
+          onUpdate={updatePlotWithChanges}
+          onReset={onReset}
+          onDataChange={onDataChange}
+        />
+        <div style={{ paddingTop: '10px' }}>
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:332]: Instead of using this `<div>`, does replacing it with `<Space>` give the same spacing effect?

src/components/plots/styling/MarkerGeneSelection.jsx
```diff
@@ -4,13 +4,15 @@ import {
...
   Button,
 } from 'antd';
 
+import GeneReorderTool from 'components/plots/GeneReorderTool';
+import GeneSearchBar from 'components/plots/GeneSearchBar';
+
 const MarkerGeneSelection = (props) => {
   const {
-    config, onUpdate, onReset, onGeneEnter,
+    config, plotUuid, searchBarUuid, onUpdate, onReset, onDataChange,
```
aerlaut: `data` is a general. The data that's contained here is genes right? Would naming it to `onGeneChange` make the variable more concise?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -89,27 +91,26 @@ const DotPlotPage = (props) => {
...
   const {
     loading: cellSetsLoading,
     error: cellSetsError,
-    hierarchy: cellSetHierarcy,
+    hierarchy: cellSetHierarchy,
     properties: cellSetProperties,
   } = cellSets;
 
   const [moreThanTwoGroups, setMoreThanTwoGroups] = useState(false);
+  const [reorderFetched, setReorderFetched] = useState(false);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: what is `reorderFetched`? Is it whether new genes have been fetched after reordering? Can you rename this variable to make it easier to understand?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -183,33 +206,84 @@ const DotPlotPage = (props) => {
...
+
       previousComparedConfig.current = currentComparedConfig;
-      dispatch(fetchPlotDataWork(experimentId, plotUuid, plotType));
+
+      // dispatch is different depending on whether selected genes don't change or are added, re-ordered or deleted
+      // to prevent unnecessary rerenders
+      if (_.isEqual(currentSelected, previousSelected)) {
+        dispatch(fetchPlotDataWork(experimentId, plotUuid, plotType));
+        return;
+      }
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: What is the difference between this and the one on line 228-231?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -183,33 +206,84 @@ const DotPlotPage = (props) => {
...
+
+      if (currentSelected.length === previousSelected.length) {
+        reorderData(currentSelected);
+        return;
+      }
+      
+      // if (currentSelected.length < previousSelected.length) {
+      const removedGene = previousSelected.filter((gene) => !currentSelected.includes(gene))[0];
+      deleteData(removedGene);
+      // }
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I imagine the if clause here is commented because this is the default case?

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: If it is, is this case that is the easiest to deduce if all previous if statement fails?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -183,33 +206,84 @@ const DotPlotPage = (props) => {
...
+      return;
     }
 
-    if (config?.selectedGenes.length === 0 && highestDispersionGenes?.length > 0) {
-      updatePlotWithChanges({ selectedGenes: highestDispersionGenes });
-    }
-  }, [highestDispersionGenes, config, genesFetching]);
+    loadHighestDispersionGenes();
+  }, [geneData]);
+
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: You use `config` here, but you don't have it in the dependency array on line 276. This means if config changes, this won't be recalculated. Is that what's expected?

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (jszpila314) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I renamed it to `reorderAfterFetch`. It's used to enable a `useEffect` where adding a gene to the plot (via search box) will trigger the data to be reordered according to the order of selected genes - it's needed because `fetchWork` "forgets" the previous order of genes. This is related to the issue of inconsistencies between results returned by the worker with `fetchWork`.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (jszpila314) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: It is not the default case, but the if would always run if it doesn't exit through one of the previous ifs. The ifs are checked when there are config changes and it checks the possible changes to selected genes (add, re-order, remove). If the selected genes are different, a gene wasn't added and they weren't reordered then a gene must have been deleted.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (jszpila314) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: This is meant to load the initial data to the plot (genes with the highest dispersion) after `geneData` is loaded (the list of all genes). The `!config` check is needed to correctly load a skeleton if the config is not loaded as well as throw correct errors in tests, because `loadHighestDispersionGenes` uses the `config`.

src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx
Comment (jszpila314) [reply src/pages/experiments/[experimentId]/plots-and-tables/marker-heatmap/index.jsx:332]: A div with padding is needed to create some distance between the reset button and the `Gene labels:` text. Replacing it with a space would instead space the text and the buttons below it.

src/components/plots/styling/MarkerGeneSelection.jsx
jszpila314: I renamed it to `onGenesChange` since it tracks selected genes and defines the behaviour when genes are added or removed.

src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Hmm ok. I think it would be better to have the if cases ordered the following way: 
- `if (currentSelected.length > previousSelected.length)`
- `if (currentSelected.length < previousSelected.length)`
- `if (currentSelected.length === previousSelected.length)`
- `if (_.isEqual(currentSelected, previousSelected))`

It would be good if you can put a comment and explain why are you treating `===` and `_.isEqual()` differently.

Comment (jszpila314) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: `===` and `_.isEqual()` check different things: the first will pass if the genes are reordered (the length of selected genes doesn't change) and the second when there are changes elsewhere in the `config` i.e. the selected genes remain the same.
That's why the cases are ordered like this:
- `if (_isEqual(currentSelected , previousSelected)` -> if the changes are elsewhere in the config, need to fetchWork
- `if (currentSelected.length > previousSelected.length)` -> if the number of selected genes increased, need to fetchWork
- `if (currentSelected.length === previousSelected.length)` -> if the number of selected genes didn't change, reorder
- `if (currentSelected.length < previousSelected.length)` -> if the number of selected genes decreased, remove it from the plot data

The `isEqual()` case has to be before the `===` case because in both cases the length is the same.

Comment (aerlaut) [reply src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: Ah I see, I understand. In that case, will you delete the comment `// if (currentSelected.length < previousSelected.length) {`?

aerlaut (approved):

aerlaut (approved):

PR #769, opened 2022-07-11
Rerendering cluster names

Kristian-A:
# Description
Added cluster names to the object used for generating the ETag so renaming clusters would trigger a new work request. The ETag cached in the browser would no longer match the ETag that would be generated when a cluster name changes so a new work request would be sent and the cache would be updated. This  fixes the problem with the dot plot not showing the new cluster names due to the one in cache being used. Staging deployment: https://ui-kristian-ui769.scp-staging.biomage.net/ (bypassing the integration tests).

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-2013

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx
```diff
@@ -95,13 +95,19 @@ const DotPlotPage = (props) => {
...
+
   const {
     loading: cellSetsLoading,
     error: cellSetsError,
     hierarchy: cellSetHierarcy,
     properties: cellSetProperties,
   } = cellSets;
 
+
+  const clusterNames = Object.values(cellSetProperties).map((el) => el.name);
```
Comment (aerlaut) [inline src/pages/experiments/[experimentId]/plots-and-tables/dot-plot/index.jsx]: I don't think you should get the values here `clusterNames` here. Implementing it here would add another parameter to be passed to `fetchPlotDataWork`. I think it would be better to attach `clusterNames` inside the body request body that's generated for this plot.

aerlaut (changes requested):
I think it's better to insert `clusterNames` into the work request body of this plot. The work request body is hashed to get the ETag, changes to the clusterNames would change the work request body and the Etag. I suggest you make the changes inside `generatePlotWorkBody`

src/utils/work/fetchWork.js
```diff
@@ -106,7 +106,13 @@ const fetchWork = async (
   getState,
   optionals = {},
 ) => {
-  const { extras = undefined, timeout = 180, broadcast = false } = optionals;
+  const {
+    extras = undefined,
+    timeout = 180,
+    broadcast = false,
+    clusterNames = undefined,
```
ivababukova: why not using these `optionals` and `extras` and adding the cluster names in them?

ivababukova: FetchWork function is used for fetching every work, clusterNames are something specific to just some of the fetchWork calls, then why would we want to introduce a new parameter for fetchWork that is relevant only to some work types?

Kristian-A: This is exactly Agi's point which I agree with. I put the cluster names in the dot plot's work body where they are relevant. I also didn't like how I did it initially but was not sure how to pass them otherwise. I think now it should be good

src/redux/actions/componentConfig/fetchPlotDataWork.js
```diff
@@ -7,12 +7,24 @@ import endUserMessages from 'utils/endUserMessages';
 import generatePlotWorkBody from 'utils/work/generatePlotWorkBody';
 import { fetchWork } from 'utils/work/fetchWork';
 
-const fetchPlotDataWork = (experimentId, plotUuid, plotType) => async (dispatch, getState) => {
+const getClusterNames = (state) => {
+  const clusterIds = state.cellSets.hierarchy[0]?.children.map((entry) => entry.key);
```
aerlaut: hierarchy[0] is `louvain` right? This means the code only gets names of the louvain clusters? What if there are changes to names in the custom cell sets?

src/utils/work/generatePlotWorkBody.js
```diff
@@ -9,17 +9,18 @@ const composeDotPlotWorkBody = (config) => {
     numberOfMarkers: config.nMarkerGenes,
     customGenesList: config.selectedGenes,
     groupBy: config.selectedCellSet,
+    clusterNames,
```
aerlaut: I think it would be good to put `clusterNames` as the last entry in the body and put a comment, saying that this property is added to make the worker reload if there are changes to the cluster name. The comment would be useful because the property is not used for the calculation,

src/redux/actions/componentConfig/fetchPlotDataWork.js
```diff
@@ -7,12 +7,24 @@ import endUserMessages from 'utils/endUserMessages';
...
+  return clusterNames;
+};
+
+const fetchPlotDataWork = (
+  experimentId,
+  plotUuid,
+  plotType,
+) => async (dispatch, getState) => {
   const config = getState().componentConfig[plotUuid]?.config ?? initialPlotConfigStates[plotType];
+  const clusterNames = getClusterNames(getState());
```
aerlaut, line 26: Really sorry about this. On second thoughts, I think it's better to put `clusterNames` into `config`. This means you can do `config.clusterNames = clusterNames;` after this line, and remove `clusterNames` from the signature of `generatePlotWorkBody`

src/redux/actions/componentConfig/fetchPlotDataWork.js
```diff
@@ -9,8 +9,10 @@ import { fetchWork } from 'utils/work/fetchWork';
 
 const getClusterNames = (state) => {
   const clusterIds = state.cellSets.hierarchy[0]?.children.map((entry) => entry.key);
-  const clusterNames = clusterIds?.map((clusterId) => state.cellSets.properties[clusterId].name);
+  const customClusterIds = state.cellSets.hierarchy[1]?.children.map((entry) => entry.key);
```
aerlaut: You're hardcoding the position of the cell sets in in the hierarchy (e.g. `0`, `1`) - I don't think you should do this. We want to detect changes to any cell sets in the platform right? So instead of hardcoding the values, I think you can iterate over all the cellsets in hierarchy and then get the name from there? E.g. `state.cellSets.hierarchy.reduce((cellSet) => cellSet...)`

aerlaut (approved):

PR #773, opened 2022-07-15
[BIOMAGE-1774] - Check sample file sizes

aerlaut:
# Description
Add a check for sample files. Check if samples have the correct barcodes, features and matrix files.

To test, you'll need to modify the some of your samples to trigger the errors: 
- Invalid barcodes: barcodes has more/less lines than record in matrix. Edit a barcodes.tsv and increase/delete lines.
- Invalid features: features has more/less lines than recorded in matrix. Edit a features.tsv and increase/delete lines.
- Transposed matrix: happens if the number of expected features and barcode files are swapped. Edit a matrix.mtx file and swap the 1st and 2nd entry of the 3rd line (e.g. 100 200 3000 -> 200 100 3000).

If you need help in testing, ask Agi.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-1774

https://ui-agi-lumpy-flounder.scp-staging.biomage.net/

N/A

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/utils/upload/sampleInspector.js
```diff
@@ -0,0 +1,136 @@
...
+  INVALID_TRANSPOSED_MATRIX: 'INVALID_TRANSPOSED_MATRIX',
+};
+
+const verdictText = {
+  [Verdict.INVALID_BARCODES_FILE]: 'Barcodes file is invalid',
+  [Verdict.INVALID_FEATURES_FILE]: 'Features file is invalid',
+  [Verdict.INVALID_TRANSPOSED_MATRIX]: 'Matrix file is transposed',
+};
+
+const CHUNK_SIZE = 2 ** 18; // 250 kb
```
cosa65: What is this chunk size?

src/utils/upload/processUpload.js
```diff
@@ -129,12 +131,25 @@ const processUpload = async (filesList, sampleType, samples, experimentId, dispa
...
+    const { valid, verdict } = await inspectSample(sample);
+    const validationMessage = verdict.map((item) => `${verdictText[item]}`).join('\n');
+
     const filesToUploadForSample = Object.keys(sample.files);
 
+    if (!valid) {
+      pushNotificationMessage('error', `Error uploading sample ${name}: ${validationMessage}`);
+      return;
+    }
+
```
cosa65: I was thinking instead of showing a validation error we could show these errors in `FileUploadModal`. We already perform some checks in `onDrop`, so maybe just add that there?

That way they can see more easily exactly what file is wrong and how instead of getting a lot of push notifications.

src/utils/upload/sampleInspector.js
aerlaut: chunk size is the size of file that we want to process at a time. To process the file, we don't open the file at once, but take it part by part (as large as chunk sizes). This allows us to process large files without having to load all of them into memory

src/utils/upload/processUpload.js
aerlaut: We can't, because the files aren't yet uploaded in the file upload modal, and we need to open the file to check its validity

cosa65: After speaking in a call, we can do this (we don't need the files to be uploaded to perform these checks) but it involves some ux design steps first so it was decided to move it to another ticket so I am resolving this comment

src/utils/upload/sampleInspector.js
cosa65: that's great! could you add a comment explaining that?

cosa65:
with the (valid) WT1 sample I am getting that the features file is invalid.

I also tried with some bigger experiments and I nothing seems to happen (I don't get any invalid error but the samples don't show up either)

src/utils/upload/sampleInspector.js
```diff
@@ -0,0 +1,136 @@
...
+  const numFeaturesLines = await getNumLines(features);
+
+  const verdict = [];
+  const valid = numBarcodeLines === barcodeSize && numFeaturesLines === featuresSize;
+
+  const isSampleTransposed = numBarcodeLines === featuresSize && numFeaturesLines === barcodeSize;
+  const isBarcodesInvalid = numBarcodeLines !== barcodeSize;
+  const isFeaturesInvalid = numFeaturesLines !== featuresSize;
+
+  if (isSampleTransposed) { verdict.push(Verdict.INVALID_TRANSPOSED_MATRIX); } else {
```
kafkasl: I think that considering we have the information about how many numbers of lines we are expecting and which ones are not matching we can do a better job on error reporting than "Invalid matrix file". We can provide more info like "Invalid matrix file. Barcodes contains 25k genes but matrix has 30k genes" or stuff like that maybe even simpler like "Number of features not matching"

src/__test__/data/mock_files/barcodes.tsv
```diff
@@ -0,0 +1,2 @@
+AAACCCATCAAACCTG-1
```
kafkasl, line 1: is it intentional to add both the `barcodes.tsv` and `barcodes.tsv.gz`?

src/utils/upload/sampleInspector.js
aerlaut: Turns out we can't arbitrarily take chunks out of files and read them separately. Part of the GZIP compression works by changing repetitive characters into "pointers" to a previous part located in the earlier part of the file. If we take a middle slice of the file arbitrarily, some pointers will lose reference to the earlier part because they are not included in the slice.

The file is divided into blocks which, I think, should be able to be decoded independently. However, this will require more reading time and understanding, because it seems that block sizes might vary. More info on GZIP compression: https://commandlinefanatic.com/cgi-bin/showarticle.cgi?article=art001

Looking at available samples, the bulk of data is located in the matrix file. The size of features and barcodes files are not much (in the order of 10 MB). For these files, I think we can decompress them into memory without slicing them.

src/__test__/data/mock_files/barcodes.tsv
aerlaut: yup, so that we can just use the gzipped file and not have to gzip the unzipped file everytime we test.

src/utils/upload/sampleInspector.js
aerlaut: Error message changed

src/utils/upload/sampleInspector.js
```diff
@@ -0,0 +1,136 @@
...
+  const fileArrBuffer = await fileObject.slice(0, 500).arrayBuffer();
+
+  matrixHeader = compressed
+    ? await decode(await decompress(fileArrBuffer))
+    : await decode(fileArrBuffer);
+
+  // The matrix header is the first line in the file that splits into 3
+  header = matrixHeader.split('\n').find((line) => line.split(' ').length === 3);
+
+  const [featuresSize, barcodeSize] = header.split(' ');
```
kafkasl: You are not checking that the number of lines in the matrix matches the header (the 3rd parameter in the header). I have tried to just remove lines from the bottom and it passes the validation.

src/utils/upload/sampleValidator.js
```diff
@@ -0,0 +1,121 @@
...
+  const features = sample.files['features.tsv.gz'] || sample.files['features.tsv'] || sample.files['genes.tsv.gz'] || sample.files['genes.tsv'];
+  const matrix = sample.files['matrix.mtx.gz'] || sample.files['matrix.mtx'];
+
+  const [
+    expectedNumFeatures,
+    expectedNumBarcodes,
+  ] = await extractSampleSizes(matrix);
+
+  const numBarcodesFound = await getNumLines(barcodes);
+  const numFeaturesFound = await getNumLines(features);
```
aerlaut, line 91: There are error messages for invalid matrix file size, but I don't see it getting checked here. Are we checking the matrix file size? @kafkasl

kafkasl: no, I removed it

kafkasl (approved):

PR #789, opened 2022-08-01
Fix cell sets racing condition and improve cell sets redux status checking

cosa65:


src/components/plots/FrequencyPlot.jsx
```diff
@@ -23,13 +23,13 @@ const FrequencyPlot = (props) => {
   const [plotSpec, setPlotSpec] = useState({});
 
   useEffect(() => {
-    if (cellSets.loading && !cellSets.error) {
```
cosa65, line 26: `cellSets.loading` here was unnecessary, it is being checked inside `loadCellSets` already

src/components/plots/styling/SelectData.jsx
```diff
@@ -72,12 +67,12 @@ const SelectData = (props) => {
...
-  if (!config || cellSetsLoading) {
-    return <Skeleton.Input style={{ width: 200 }} active />;
+  if (cellSets.error) {
+    return <InlineError message='Error loading cell set' />;
   }
 
-  if (cellSetsError) {
-    return <InlineError message='Error loading cell set' />;
+  if (!config || !cellSets.accessible) {
+    return <Skeleton.Input style={{ width: 200 }} active />;
```
cosa65, line 74: Reordered so that error is checked first (accessible includes error)

src/redux/selectors/cellSets/getCellSets.js
```diff
@@ -2,6 +2,20 @@ import createMemoizedSelector from 'redux/selectors/createMemoizedSelector';
...
-const getCellSets = () => (state) => (Object.keys(state).length ? state : initialState);
+const getCellSets = () => (state) => {
+  const stateToReturn = (Object.keys(state).length ? state : initialState);
+
+  const accessible = !(
+    stateToReturn.loading
+    || stateToReturn.initialLoadPending
+    || stateToReturn.updatingClustering
+    || stateToReturn.error
+  );
```
aerlaut: I think the `!(...)` is very easy to miss. This is equal to `!stateToReturn.loading && ...` right? I think it's easier written that way.

src/redux/selectors/cellSets/getCellSets.js
```diff
@@ -2,6 +2,20 @@ import createMemoizedSelector from 'redux/selectors/createMemoizedSelector';
...
+
+  return {
+    ...stateToReturn,
+    accessible,
+  };
```
aerlaut, line 16: Why not just store `accessible` as a property? I think the idea of having `accessible` is good. However, returning it as an additional export from `getCellSets()` forces us to always use `getCellSets()` if we want to check for accesibility.

src/components/plots/ContinuousEmbeddingPlot.jsx
```diff
@@ -36,7 +36,7 @@ const ContinuousEmbeddingPlot = (props) => {
   const [plotSpec, setPlotSpec] = useState({});
 
   useEffect(() => {
-    if (cellSets.loading && !cellSets.error) {
+    if (!cellSets.error) {
```
kafkasl: some questions:
1. when we load a new embeddings method the old cellsets are discarded so we need to reload them again, is that right?
2. this useEffect() depending on `experimentId` makes sense? how can the experimentId change while staying on the same plot? afaik the only way for the experimentId to change is to navigate to another experiment (which will be a new page) and hence the cellsets can be loaded as part of the initial load. Is that true?

src/redux/actions/cellSets/loadCellSets.js
```diff
@@ -7,19 +7,22 @@ import endUserMessages from 'utils/endUserMessages';
...
 const loadCellSets = (experimentId, forceReload = false) => async (dispatch, getState) => {
   const {
-    loading, error, updatingClustering,
+    loading, error, updatingClustering, initialLoadPending,
   } = getState().cellSets;
 
-  if (!forceReload && ((!loading && !error) || updatingClustering)) {
-    return null;
-  }
+  const loadingAlreadyReserved = loading || updatingClustering;
```
kafkasl: what does `loadingAlreadyReserverd` mean? the name is not very clear to me, does it mean that the load request has already been sent?

src/redux/selectors/cellSets/getCellSets.js
kafkasl: I agree, it's more clear also that none of the conditions should be true (using &&)
```
! stateToReturn.loading
&& !stateToReturn.initialLoadPending
&& !stateToReturn.updatingClustering
&& !stateToReturn.error
```

src/redux/selectors/cellSets/getCellSets.js
cosa65: I think it's better to store no more data than the absolute necessary and then derive properties from that data stored. (it makes it easy to keep the data consistent and avoid weird behaviors because we forget to set some property that depends on other props, there's less stuff to keep track of).

However, I do see your point on the fact that having it in `getCellSets()` makes us dependent on it. To deal with that I could add another selector that returns whether cell sets is `accessible` or not to make this accessible property more directly usable. (I saw some examples of this usage in the redux page)

src/redux/actions/cellSets/loadCellSets.js
cosa65: It means that either:
> load request has already been sent
or the clustering is being updated, so there's no point in fetching outdated cell sets until the new cell sets clustering is done. 

I'll look into changing the name

src/redux/selectors/cellSets/getCellSets.js
cosa65: image

> We specifically recommend that Redux apps should [keep the Redux state minimal, and derive additional values from that state whenever possible](https://redux.js.org/style-guide/#keep-state-minimal-and-derive-additional-values).

https://redux.js.org/usage/deriving-data-selectors

aerlaut: Hmm... then instead of returning an object with 2 items, maybe return null if the cellSets is not ready? I don't think it's nice to have another selector just to check accessibility

cosa65: the new selecotr is memoized so there isn't any performance impact there)

cosa65: > maybe return null if the cellSets is not ready

if we return `null` then we can't read whether it was an error or what happened exactly

cosa65: Resolved over slack, we decided to keep the original version with `accessible` in `getCellSets`

src/redux/actions/cellSets/loadCellSets.js
cosa65: Renamed to `loadingBlocked`

src/components/plots/ContinuousEmbeddingPlot.jsx
cosa65: > 1. when we load a new embeddings method the old cellsets are discarded so we need to reload them again, is that right?

No, the cell sets are always the same (the ones stored in the cell-sets-* bucket). They just need to be downloaded if they haven't been yet.

What's happening here is it's calling `loadCellSets` but after the cellSets are loaded the first time nothing will happen anymore. Actually, having `loadCellSets` in this `useEffect` is probably a bit confusing, it would be more clear for it to be in a `useEffect` of its own without any dependency.

> 2. this useEffect() depending on experimentId makes sense? 

No, it doesn't, I'll remove that.

kafkasl (approved):

PR #793, opened 2022-08-05
[BIOMAGE-2047] Fix adding multiple genes

jszpila314:
# Description
After the gene selection in Marker Heatmap and Dot Plot was updated it is impossible to add multiple genes to these plots at the same time. This is intended to allow pasting a gene list to add as well as searching for multiple genes one by one.

# Details
#### URL to issue
https://biomage.atlassian.net/browse/BIOMAGE-2047

https://ui-jakub-ui793.scp-staging.biomage.net/

#### Integration test branch
master

# Merge checklist

### Code updates

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


src/components/plots/styling/MarkerGeneSelection.jsx
```diff
@@ -43,7 +46,15 @@ const MarkerGeneSelection = (props) => {
...
       <Space direction='vertical' size='small'>
-        <p style={{ margin: 0 }}>Type in a gene name and select it to add it to the plot.</p>
+        <p style={{ margin: 0 }}>
+          Type in a gene name and select it to add it to the plot.
+          <Tooltip
+            title='To add multiple genes, separate them with a space or comma.'
+          >
+            &nbsp;
+            <QuestionCircleOutlined />
+          </Tooltip>
```
aerlaut: I think this information should be placed together with the description, not in a tooltip.

aerlaut: Any reason to allow **both** space and comma to delimit gene names? Why not just use comma? I think it'll be easier for us to handle and less ambiguous for users.

src/components/plots/GeneSearchBar.jsx
```diff
@@ -25,31 +26,56 @@ const GeneSearchBar = (props) => {
...
+    setValue(genes.join(', '));
+    setOptions([]);
   };
 
-  const onSearch = (searchText) => {
-    setValue(searchText);
+  const onSearch = (input) => {
+    setValue(input);
+
+    const genes = input.split(/(?<!-)[,\s]+(?!-)/);
```
aerlaut: This regex pattern is used 3 times in this file. It's not good practice to have duplicated code. Can you refactor it into a variable to avoid duplication?

src/components/plots/GeneSearchBar.jsx
```diff
@@ -25,31 +26,56 @@ const GeneSearchBar = (props) => {
...
     setOptions(!searchText ? [] : filterGenes(searchText, geneList, config?.selectedGenes));
   };
 
+  const onClick = () => {
+    if (value === '') return;
+
+    const newGenes = value.split(/(?<!-)[,\s]+(?!-)/).map((gene) => gene.trim()).filter((gene) => geneList.includes(gene));
+    const genes = _.uniq([...config?.selectedGenes, ...newGenes]);
+
+    if (_.isEqual(genes, config?.selectedGenes)) return;
```
aerlaut: This piece of code is to handle the case where the users selects a gene that's already selected right? Instead of doing this, would it be simpler to remove genes that are already selected from the list of genes that is displayed?

src/components/plots/GeneSearchBar.jsx
```diff
@@ -25,31 +26,56 @@ const GeneSearchBar = (props) => {
...
+  const onSearch = (input) => {
+    setValue(input);
+
+    const genes = input.split(/(?<!-)[,\s]+(?!-)/);
+    const searchText = genes[genes.length - 1];
+
     setOptions(!searchText ? [] : filterGenes(searchText, geneList, config?.selectedGenes));
   };
 
+  const onClick = () => {
```
aerlaut: Let's rename this to something else - maybe `addGene`?. The naming pattern `on...` is usually used for callbacks that are passed into the component via props.

src/components/plots/styling/MarkerGeneSelection.jsx
jszpila314: This is to allow two use cases:
- searching for genes individually (where separating with a comma seems more natural)
- pasting a list of genes from somewhere (separated by whitespaces)

It might make more sense to have them only separated by spaces, but users might run into an edge case with genes with multiple IDs which have format `name - ENSG#` while searching. That's also what the regex is for.

jszpila314: I didn't want the description to be too cluttered so I put it in a tooltip, but I will try to rewrite it to be consice and also include the tooltip info.

src/components/plots/GeneSearchBar.jsx
jszpila314: The genes are already disabled in the gene search if they have already been added. This doesn't prevent attepting to add them with the add button, which could happen if:
- the entire name is written manually instead of selecting an option and then added
- a pasted list of genes includes genes that are already loaded

src/components/plots/GeneSearchBar.jsx
jszpila314: I moved the regex outside the functions to avoid duplications, where it becomes `const genes = value.split(/(?<!-)[,\s]+(?!-)/);` but I'm running into a problem with `onSearch` where i both update `value` and need to use the updated `value` to define `searchText` for filtering options. I think i need to duplicate the regex there, because it needs to use `input` instead of `value`.

aerlaut: Oh no, I mean take out the Regex. Can you do something like

    const GENES_REGEX = /(?<!-)[,\s]+(?!-)/
    ....
    const genes = values.split(GENES_REGEX)
    ....
    const inputGenes = input.split(GENES_REGEX)

src/components/plots/GeneSearchBar.jsx
```diff
@@ -25,31 +26,59 @@ const GeneSearchBar = (props) => {
...
+        onSearch={onSearch}
+        placeholder='Search for genes...'
+      />
+      <Button
+        type='primary'
+        onClick={addGenes}
+        style={{ width: '20%' }}
+      >
+        Add
+      </Button>
```
aerlaut, line 80: Is it possible to add some space between the `AutoComplete` and `Button` tag? if not or it's too much of a hassle, then it's fine.

aerlaut (approved):

src/components/plots/GeneSearchBar.jsx
```diff
@@ -25,31 +26,60 @@ const GeneSearchBar = (props) => {
...
+      <AutoComplete
+        allowClear
+        style={{ width: '80%' }}
+        value={value}
+        options={options}
+        onSelect={onOptionSelect}
+        onSearch={onSearch}
+        placeholder='Search for genes...'
+      />
+
```
aerlaut: please remove this space

PR #848, opened 2023-07-19
Merge

cosa65:
# Description

# Details
https://github.com/hms-dbmi-cellenics/pipeline/pull/312
https://github.com/hms-dbmi-cellenics/worker/pull/307
https://github.com/hms-dbmi-cellenics/api/pull/442
https://github.com/hms-dbmi-cellenics/ui/pull/848

#### Integration test branch
master

# Merge checklist

### Code updates
- [ ] All new dependency licenses have been checked for compatibility.

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional


.github/workflows/pr_validate.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       - id: check-issue-url
         name: Check issue URL
         run: |-
-          REGEX="#### URL to issue\s?\r\n(https:\/\/github\.com\/hms-dbmi-cellenics\/issues\/issues\/[0-9]+|N\/A)"
+          REGEX="#### URL to issue\s?\r\n(https:\/\/github\.com\/hms-dbmi-cellenics\/issues\/issues\/[0-9]+|https:\/\/biomage\.atlassian\.net\/browse\/BIOMAGE-[0-9]+|https:\/\/github\.com\/biomage-org\/issues\/issues\/[0-9]+|N\/A)"
```
alexvpickering: same question as other PRs

src/__test__/components/header/HelpButton.test.jsx
```diff
@@ -23,7 +23,25 @@ describe('HelpButton', () => {
...
+    expect(guideLink).toHaveAttribute('target', '_blank');
+
+    const websiteLink = screen.getByText(/tutorial videos/i).closest('a');
+    expect(websiteLink).toHaveAttribute('href', 'https://www.youtube.com/@biomageltd4616/featured');
+    expect(websiteLink).toHaveAttribute('target', '_blank');
 
-    expect(screen.getByText(/For 1-2-1 support with your analysis/i)).toBeDefined();
+    const forumLink = screen.getByText(/Cellenics community forum/i).closest('a');
+    expect(forumLink).toHaveAttribute('href', 'https://community.biomage.net/');
+    expect(forumLink).toHaveAttribute('target', '_blank');
```
alexvpickering: should these tests be deployment-agnostic (similarly to the components themselves)?

alexvpickering (commented):
initial comments

src/__test__/utils/upload/fileInspector.test.js
alexvpickering, line 1: just checking - these duplications intentional?

src/components/ContentWrapper.jsx
```diff
@@ -340,12 +341,13 @@ const ContentWrapper = (props) => {
         return <PipelineRedirectToDataProcessing experimentId={routeExperimentId} pipelineStatus='running' />;
       }
 
-      if (process.env.NODE_ENV === 'development') {
-        return children;
+      if (qcStatusKey === pipelineStatusValues.NOT_CREATED
+        && currentModule !== modules.DATA_PROCESSING && !isSeurat) {
+        return <PipelineRedirectToDataProcessing experimentId={routeExperimentId} pipelineStatus='toBeRun' />;
```
alexvpickering, line 346: why is this added and the below one deleted? They are similar but different

src/components/ContentWrapper.jsx
```diff
@@ -367,12 +369,21 @@ const ContentWrapper = (props) => {
...
     );
 
+    const {
+      DATA_EXPLORATION, DATA_MANAGEMENT, DATA_PROCESSING, PLOTS_AND_TABLES,
+    } = modules;
+
+    // disable links if user is not in one of the experiment analysis modules
+    const nonExperimentModule = ![DATA_EXPLORATION,
+      DATA_MANAGEMENT, DATA_PROCESSING, PLOTS_AND_TABLES]
+      .includes(currentModule) && disableIfNoExperiment;
```
alexvpickering, line 379: what is, and is not, an experiment analysis module?

src/components/Header.jsx
```diff
@@ -36,6 +35,7 @@ const Header = (props) => {
         extra={(
           <Space size='large'>
             <Space>
+              {renderDomainSpecificContent('Courses')}
```
alexvpickering: maybe this should be written so that it is not so explicitly a location to put an external link to courses in case another party wants to use it for a different purpose. Something like `'OptionalFirstHeaderButton'` which happens to be for courses in the case of Biomage?

src/components/TagManager.jsx
```diff
@@ -2,17 +2,22 @@ import React from 'react';
...
 
 const TagManager = ({ environment }) => {
   const { enabled, containerId } = getTrackingDetails(environment);
-
+  const accountId = nextConfig()?.publicRuntimeConfig?.accountId;
   // if tracking is not enabled don't add tag manager to the head
   if (!enabled) return (null);
 
+  // whether to use biomage or HMS matomo account
+  const matomoName = accountId === AccountId.BIOMAGE ? 'biomage' : 'cellenics';
```
alexvpickering: I prefer not to have organization names (HMS, Biomage, or other) appearing in the code base more than absolutely necessary and also to make it so that it would be easy for another party to join. We could add this to a config and here it could just be something like:

 `const matomoName = matomoNameByAccountId[accountId]`

src/config/defaultConfig.js
```diff
@@ -1,32 +1,32 @@
...
 };
 
-const feedbackWebhookEndpointsByAccountId = {
-  [AccountId.HMS]: 'aHR0cHM6Ly9ob29rcy5zbGFjay5jb20vc2VydmljZXMvVDAzUTI3MjRRUksvQjAzUThCR0JLSEIvUk41elZtU1RudXUyMWxURUh1a3I3ajdJ', // pragma: allowlist secret
-};
-
-const referralWebhookEndpointsByAccountId = {
-  [AccountId.HMS]: 'aHR0cHM6Ly9ob29rcy5zbGFjay5jb20vc2VydmljZXMvVDAzUTI3MjRRUksvQjAzVDdDSjFVODIvWFJ3cGtaa2huT2ZCbEVObkNMamtad0pn', // pragma: allowlist secret
-};
+const getAccountInfo = (account) => accountInfo[account] || accountInfo.default;
```
alexvpickering: Having Biomage or HMS as the default for account configuration should be avoided where possible. Is this necessary for local development only where `nextConfig()` won't work? If so, can we just add an environmental variable `CELLENICS_AWS_ACCOUNT_ID` or similar that is used as the default instead and then index `accountInfo` by organization?

The current setup makes it so that I see Biomage specific HelpButton locally.

src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx
```diff
@@ -56,21 +54,13 @@ const BatchDiffExpression = (props) => {
...
-  useEffect(() => {
-    if (hierarchy && hierarchy.length === 0) return;
-
-    const samples = hierarchy?.find(
-      (rootNode) => (rootNode.key === 'sample'),
-    )?.children;
-
-    setNumSamples(samples.length);
-  }, [Object.keys(properties).length]);
-
```
Comment (alexvpickering) [inline src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx:73]: This alternative implementation makes batch differential expression work for Seurat object uploads. The issue is that `state.experimentSettings.info.sampleIds.length` will always be 1 for Seurat object uploads (a single .rds file is uploaded with no way to know that there are multiple samples until after being processed).

The alternative uses the sample cellsets to determine the number of samples, which will be correct for Seurat object uploads. It is also consistent with the current implementation in Data Exploration where there is a similar check to see if between-group comparisons are allowed.

src/utils/domainSpecificContent.jsx
```diff
@@ -0,0 +1,79 @@
...
+        {reusedContent.HelpButton.BiomageUserGuide}
+        <br />
+        {reusedContent.HelpButton.OneToOneSupport}
+      </>
+    ),
+  },
+};
+
+export default function renderDomainSpecificContent(component) {
+  const accountId = nextConfig()?.publicRuntimeConfig?.accountId;
```
alexvpickering: if you  add a `CELLENICS_AWS_ACCCOUNT_ID` for local dev also update this and any other locations

src/utils/tracking.js
alexvpickering, line 1: I think this file needs a bit of editing to make the account specific stuff a bit better. Ternary operator (`?`) works for HMS and Biomage only but does not make it simple for another party to join. I would prefer to have a config by accountId that is directly indexed like the following:

```javascript
const matomoConfig = {

  [AccountId.HMS] : {
    matomoName: 'cellenics',
    productionContainerId: 'zdMhc9ey',
    ...
    
  }
  [AccountId.BIOMAGE] : {
    matomoName: 'biomage',
    productionContainerId: 'lkIodjnO',
    ...
  }
      
}

...
containerId: matomoConfig[accountId].productionContainerId,
...

```

src/__test__/utils/upload/fileInspector.test.js
cosa65: `.mockReturnValueOnce(` mocks the result for one call, if more than one call is expected to be performed, then `mockReturnValueOnce` needs to be called many times (in particular if the returns are supposed to differ across each call)

alexvpickering (changes requested):
most comments are just small things related to approaching multi-organization usage. Code looks really good - thanks for improving on my implementations!

src/components/ContentWrapper.jsx
cosa65: This `PipelineRedirectToDataProcessing` is supposed to be returned when qc is not created, but it was being returned when gem2s was not created, so I replaced it with the check on `qcStatusKey === pipelineStatusValues.NOT_CREATED` which didn't seem wordy enough to warrant a new variable in my opinion.

I changed `pipelineNotCreated` for `gem2sNotCreated` because I think the ambiguous naming of `pipeline` to refer to gem2s/seurat only and not qc makes us prone to problems like this.

src/components/ContentWrapper.jsx
cosa65: Based on the odd ones out of `modules`, I would say they are `REPOSITORY` and `SETTINGS`, but @StefanBabukov might know more

src/__test__/components/header/HelpButton.test.jsx
cosa65: I would say the components aren't deployment-agnostic, there’s code that depends on what deployment is showing up so I would argue that testing even for both paths is useful.

Also they serve the purpose of checking that, for example, we don’t break any hms/biomage deployment specific functionality and viceversa.

src/components/ContentWrapper.jsx
alexvpickering: ahh okay gotcha makes sense -- this is disabling the links if in the datasets repository or settings page

src/__test__/components/header/HelpButton.test.jsx
cosa65: Just pushed a version with tests for both deployments, unresolving this comment so that you can tkae a look (feel free to resolve again)

src/components/Header.jsx
cosa65: right, `'Courses'` only makes sense in the context of the biomage deployment

cosa65: I renamed it to `'HeaderExtraButton'`

src/components/TagManager.jsx
cosa65: Moved into `getDomainSpecificContent`, let me know if it's ok now

src/config/defaultConfig.js
cosa65: The `AWS_ACCOUNT_ID`  is picked up originally from `next.config.js` and stored in `publicRuntimeConfig.accountId`.

Following your idea, I added a `DEV_ACCOUNT` that should be added when doing `npm run` like this:  `DEV_ACCOUNT=HMS npm run` and then sets the corresponding account id according to what dev account aws selected. Let me know if you think this would help. Also, if `DEV_ACCOUNT` is not set it will fail (so basically we won't allow having a default env)

cosa65: With [this](https://github.com/hms-dbmi-cellenics/ui/pull/848#discussion_r1270674147) you shouldn't see biomage anymore locally. 

About the `default` you mentioned should be avoided where possible, the problem here is that we want to display `supportEmail` and all of these for more than `AccountId.BIOMAGE`, I do think that this could be moved over to `getDomainSpecificContent` instead though, let me know if you would prefer that. However, keep in mind that in `getDomainSpecificContent` the `default` is set to `BIOMAGE_PRIVATE` (so the behavior would be the same).

src/utils/tracking.js
cosa65: Moved the domain specific stuff into `getDomainSpecificContent`

src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx
Comment (cosa65) [reply src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx:73]: I see, we still don't need a `useEffect` I think, switched to one of the custom selectors which takes care of most of it. Also added a `useMemo` for the part that performs the `length === 1` check to reduce the amount of re-renders

src/components/TagManager.jsx
alexvpickering: Thanks! looks good

src/config/defaultConfig.js
alexvpickering: I like the changes that you have proposed in https://github.com/biomage-org/ui/pull/250. While it still uses non-specified AWS account IDs (the `default` case) for Biomage private deployments, it is much better in that it contains it within a single file (`getDomainSpecificContent.jsx`), which can be readily changed in the future. Your proposed changes also remove the account id checks and organization name drops across multiple files which is greatly appreciated.

Would it make sense to move the domain-specific content in the api to a similar centralized file like in the ui with `getDomainSpecificContent`? There are fewer items currently (only in `getAdminSub.js`, `src/default-config.js`, and a few `agreedToTerms` related items that I can see).

src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx
Comment (alexvpickering) [reply src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx:73]: sounds good -- I'll leave the comment unresolved just as a reminder to myself to check that batch DE works for Seurat object uploads

src/utils/getDomainSpecificContent.jsx
```diff
@@ -0,0 +1,95 @@
...
+      </>
+    ),
+  },
+};
+
+const biomageContainerIds = {
+  [Environment.PRODUCTION]: 'lkIodjnO',
+  [Environment.STAGING]: 'FX7UBNS6',
+  [Environment.DEVELOPMENT]: 'lS8ZRMXJ',
+};
```
alexvpickering, line 37: consider moving up to `reusedContent`

src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx
Comment (alexvpickering) [reply src/pages/experiments/[experimentId]/plots-and-tables/batch-differential-expression/index.jsx:73]: works nicely :)

alexvpickering (approved):
LGTM :rocket:

PR #977, opened 2024-01-31
warn annotation loss with resolution change

alexvpickering:
# Description
Makes it so that resolution changes also require confirmation before they are run. Also adds a warning that re-running QC will cause loss of annotations.

staging:

https://ui-alex-ui977.staging.single-cell-platform.net/

# Details
#### URL to issue
https://github.com/hms-dbmi-cellenics/issues/issues/7

N/A

#### Integration test branch
master

# Merge checklist

### Code updates
- [x] All new dependency licenses have been checked for compatibility.

### Manual/unit testing

### Integration testing

### Documentation updates

### Optional

### DIFF
--- a/src/components/data-processing/ConfigureEmbedding/CalculationConfig.jsx
+++ b/src/components/data-processing/ConfigureEmbedding/CalculationConfig.jsx
@@ -1,5 +1,5 @@
 import React, {
-  useState, useEffect, useCallback,
+  useState, useEffect,
 } from 'react';
 import _ from 'lodash';
 import { useDispatch, useSelector } from 'react-redux';
@@ -9,9 +9,8 @@ import {
 import PropTypes from 'prop-types';
 import { QuestionCircleOutlined } from '@ant-design/icons';
 
-import { updateFilterSettings, saveProcessingSettings } from 'redux/actions/experimentSettings';
+import { updateFilterSettings } from 'redux/actions/experimentSettings';
 
-import { runCellSetsClustering } from 'redux/actions/cellSets';
 import PreloadContent from '../../PreloadContent';
 
 import SliderWithInput from '../../SliderWithInput';
@@ -32,7 +31,7 @@ const EMBEDD_METHOD_TEXT = 'Reducing the dimensionality does lose some informati
   + 't-SNE and UMAP are stochastic and very much dependent on choice of parameters (t-SNE even more than UMAP) and can yield very different results in different runs. ';
 
 const CalculationConfig = (props) => {
-  const { experimentId, onConfigChange, disabled } = props;
+  const { onConfigChange, disabled } = props;
   const FILTER_UUID = 'configureEmbedding';
   const dispatch = useDispatch();
 
@@ -46,21 +45,7 @@ const CalculationConfig = (props) => {
   const { umap: umapSettings, tsne: tsneSettings } = data?.embeddingSettings.methodSettings || {};
   const { louvain: louvainSettings } = data?.clusteringSettings.methodSettings || {};
 
-  const debouncedClustering = useCallback(
-    _.debounce((resolution) => {
-      dispatch(runCellSetsClustering(experimentId, resolution));
-    }, 1500),
-    [],
-  );
-
-  const [resolution, setResolution] = useState(null);
   const [minDistance, setMinDistance] = useState(null);
-
-  useEffect(() => {
-    if (!resolution && louvainSettings) {
-      setResolution(louvainSettings.resolution);
-    }
-  }, [louvainSettings]);
 
   useEffect(() => {
     if (!minDistance && umapSettings) {
@@ -68,19 +53,8 @@ const CalculationConfig = (props) => {
     }
   }, [umapSettings]);
 
-  const dispatchDebounce = useCallback(_.debounce((f) => {
-    dispatch(f);
-  }, 1500), []);
-
   const updateSettings = (diff) => {
-    if (diff.embeddingSettings) {
-      // If this is an embedding change, indicate to user that their changes are not
-      // applied until they hit Run.
-      onConfigChange();
-    } else {
-      // If it's a clustering change, debounce the save process at 1.5s.
-      dispatchDebounce(saveProcessingSettings(experimentId, FILTER_UUID));
-    }
+    onConfigChange(diff);
 
     dispatch(updateFilterSettings(
       FILTER_UUID,
@@ -394,21 +368,14 @@ const CalculationConfig = (props) => {
               max={10}
               step={0.1}
               disabled={disabled}
-              value={resolution}
-              onUpdate={(value) => {
-                if (value === resolution) { return; }
-
-                setResolution(value);
-                updateSettings({
-                  clusteringSettings: {
-                    methodSettings: {
-                      louvain: { resolution: value },
-                    },
+              value={louvainSettings.resolution}
+              onUpdate={(value) => updateSettings({
+                clusteringSettings: {
+                  methodSettings: {
+                    louvain: { resolution: value },
                   },
-                });
-
-                debouncedClustering(value);
-              }}
+                },
+              })}
             />
           </Form.Item>
         </Form>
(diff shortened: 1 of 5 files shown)


src/redux/actions/pipeline/runQC.js
```diff
@@ -42,12 +56,27 @@ const runQC = (experimentId) => async (dispatch, getState) => {
...
-      processing.configureEmbedding.embeddingSettings.method,
-      dispatch,
-    );
+  const embeddingChanged = changedQCFilters.has('embeddingSettings');
+  const clusteringChanged = changedQCFilters.has('clusteringSettings');
+  const otherChanged = [...changedQCFilters].some((value) => value !== 'embeddingSettings' && value !== 'clusteringSettings');
+
+  // if only embedding or clustering changed
+  if (!otherChanged) {
+    await dispatch(saveProcessingSettings(experimentId, 'configureEmbedding'));
```
StefanBabukov:
```suggestion
const embeddingChanged = changedQCFilters.has('embeddingSettings');
const clusteringChanged = changedQCFilters.has('clusteringSettings');

if (embeddingChanged || clusteringChanged) {
  await dispatch(saveProcessingSettings(experimentId, 'configureEmbedding'));

```

StefanBabukov: So now that means that if there's more than one changed filter, running it with changed embedding or clustering in the last step will discard them all and run only embedding / clustering?

StefanBabukov: Also if you move back the `await dispatch(saveProcessingSettings(experimentId, 'configureEmbedding'));` line to the `runOnlyClustering` and `runOnlyConfigureEmbedding` functions, this whole block can be only:

```
  const embeddingChanged = changedQCFilters.has('embeddingSettings');
  const clusteringChanged = changedQCFilters.has('clusteringSettings');

    if (embeddingChanged) {
      runOnlyConfigureEmbedding(
        experimentId,
        processing.configureEmbedding.embeddingSettings.method,
        dispatch,
      );
    }
    if (clusteringChanged) {
      runOnlyClustering(
        experimentId,
        processing.configureEmbedding.clusteringSettings.methodSettings.louvain.resolution,
        dispatch,
      );
    }
```

src/components/data-processing/ConfigureEmbedding/CalculationConfig.jsx
```diff
@@ -46,41 +45,19 @@ const CalculationConfig = (props) => {
...
-      // If this is an embedding change, indicate to user that their changes are not
-      // applied until they hit Run.
-      onConfigChange();
-    } else {
-      // If it's a clustering change, debounce the save process at 1.5s.
-      dispatchDebounce(saveProcessingSettings(experimentId, FILTER_UUID));
-    }
+    // updates to configure embedding run on worker if they are the only changes
+    // need to know if change was to embedding or clustering settings
+    const settingType = Object.keys(diff)[0];
```
StefanBabukov, line 59: Why only the first changed from the diff and not the whole `Object.keys(diff)`?

src/redux/actions/pipeline/runQC.js
```diff
@@ -42,12 +56,27 @@ const runQC = (experimentId) => async (dispatch, getState) => {
   const { processing } = getState().experimentSettings;
   const { changedQCFilters } = processing.meta;
 
-  if (changedQCFilters.size === 1 && changedQCFilters.has('configureEmbedding')) {
```
StefanBabukov, line 45: With changedQCFilters.size === 1 removed, now that means that if there's more than one changed filter, running it with changed embedding or clustering in the last step will discard them all and run only embedding / clustering?

src/components/data-processing/ConfigureEmbedding/CalculationConfig.jsx
alexvpickering: Each of the diffs is a single object where the top level key is either going to be `embeddingSettings` or `clusteringSettings`. This just extracts that key.

src/redux/actions/pipeline/runQC.js
alexvpickering: > So now that means that if there's more than one changed filter, running it with changed embedding or clustering in the last step will discard them all and run only embedding / clustering?

The problem with this is that it will run the conditional block if there are other changes to Data Processing in addition to an embedding or clustering change. I want the condition to run if there is **only** an embedding or clustering change (or both but nothing else). Embedding and clustering will be run after the pipeline finishes if there are any other changes so I don't want to call embedding/clustering in those cases.

alexvpickering: > Also if you move back the `await dispatch(saveProcessingSettings(experimentId, 'configureEmbedding'));` line to the `runOnlyClustering` and `runOnlyConfigureEmbedding` functions, this whole block can be only:
> 
> ```
>   const embeddingChanged = changedQCFilters.has('embeddingSettings');
>   const clusteringChanged = changedQCFilters.has('clusteringSettings');
> 
>     if (embeddingChanged) {
>       runOnlyConfigureEmbedding(
>         experimentId,
>         processing.configureEmbedding.embeddingSettings.method,
>         dispatch,
>       );
>     }
>     if (clusteringChanged) {
>       runOnlyClustering(
>         experimentId,
>         processing.configureEmbedding.clusteringSettings.methodSettings.louvain.resolution,
>         dispatch,
>       );
>     }
> ```

happy to move the line back if you have a preference. I moved it out so that it wasn't duplicated between the two functions

src/redux/actions/pipeline/runQC.js
alexvpickering: `changedQCFilters.size === 1` is removed because we will now be looking for either exactly `embeddingSettings` , `clusteringSettings`, or both. If both are changed but nothing else we also need to run this conditional block (so size is no longer a good way to distinguish).

StefanBabukov: Sorry I cant test in staging because i dont have an hms account but what i see is that now this will be run even if there are other `changedQCFilters` from previous steps. There is a `return` on line 81, so that means qc wont be triggered. 
What will happen if i change for example the mitochondrial content filter and then go to configure embedding and change clustering / embedding there? I think mitochondrial filter changes will be discarded and clustering / embedding will be ran right?

src/redux/actions/pipeline/runQC.js
StefanBabukov: I was only thinking in case we want to use the `runOnlyClustering` and `runOnlyConfigureEmbedding` functions  somewhere else in the code, someone might expect them to do a save and forget to `dispatch(save` before running them.

src/redux/actions/pipeline/runQC.js
alexvpickering: If you send me an email I'll create an account for you to test. Here is why it doesn't run if there are other steps changed:

```javascript
const otherChanged = [...changedQCFilters].some((value) => value !== 'embeddingSettings' && value !== 'clusteringSettings');
```
This checks to see if there are other settings changed. If there **are not**, then the below conditional block runs. See added comments to hopefully clarify:

```javascript
if (!otherChanged) {

// this runs because exactly embedding or clustering or both has changed (but nothing else)
await dispatch(saveProcessingSettings(experimentId, 'configureEmbedding'));

if (embeddingChanged) {
      runOnlyConfigureEmbedding(
        experimentId,
        processing.configureEmbedding.embeddingSettings.method,
        dispatch,
      );
    }
    if (clusteringChanged) {
      runOnlyClustering(
        experimentId,
        processing.configureEmbedding.clusteringSettings.methodSettings.louvain.resolution,
        dispatch,
      );
    }
    // we return because we are in the !otherChanged block so nothing else changed and we don't re-run QC
    return;
}
 ```

StefanBabukov: oh alright sorry

StefanBabukov (approved):

src/redux/actions/pipeline/runQC.js
alexvpickering: Not at all -- thank you -- always good to have a careful review :)

